PR #5, opened 2018-03-08
Will's WIP branch

hellais:
I am opening a PR to make it easier to answer comments.

### DIFF
--- a/ooni.go
+++ b/ooni.go
@@ -154,6 +154,8 @@ func ReadConfig(path string) (*Config, error) {
 		if err = c.Validate(); err != nil {
 			return nil, errors.Wrap(err, "validating")
 		}
+
+		return c, nil
 	}
 
 	if err != nil {
@@ -186,5 +188,7 @@ func ReadDefaultConfigPaths() (*Config, error) {
 			return c, nil
 		}
 	}
-	return nil, errors.New("failed to find a config")
+
+	// Run from the default config
+	return ReadConfig(paths[0])
 }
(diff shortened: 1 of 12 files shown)


hellais:
> Moved into the Default() method around instantiation of a default config:
> 90edffb
> Does that seem reasonable, or do you have a better suggestion?

My understanding of golang code style is that the `Default` method is usually meant to not have any consequences to calling it (it's actually used mostly as a kind of "property method").

I would suggest putting all the directory creation logic into some separate function (not a method) that is called at the beginning only if we detect that the system has not been initialised (or if the initialisation is partial).

willscott:
refactored to a function as suggested in 62b515e2e0b6f9ed96284fed28300327ef604168

ooni.go
```diff
@@ -140,11 +147,29 @@ func ParseConfig(b []byte) (*Config, error) {
...
+}
+
 // ReadConfig reads the configuration from the path
 func ReadConfig(path string) (*Config, error) {
 	b, err := ioutil.ReadFile(path)
 
 	if os.IsNotExist(err) {
+		if err = EnsureDefaultConfigDir(); err != nil {
+			return nil, errors.Wrap(err, "Creating new config")
+		}
```
hellais: I would not put this in here as this function is used to read the config file from a specific path. It doesn't make any assumption about what is on the disk. We should rather put this as something that is done when we construct the path to the config file inside of the ConfigDir

hellais: See: https://github.com/OpenObservatory/gooni/commit/efb7e87d1b66a7a5dad48f460a8f8cb553a3761c

PR #9, opened 2018-03-23
Feature/web connectivity

hellais:


nettests/websites/web_connectivity.go
```diff
@@ -1,10 +1,56 @@
...
+func lookupURLs(ctl *nettests.Controller) ([]string, error) {
+	var (
+		parsed = new(URLResponse)
+		urls   []string
+	)
+	reqURL := fmt.Sprintf("%s/api/v1/urls?probe_cc=%s",
+		orchestrateBaseURL,
+		ctl.Ctx.Location.CountryCode)
+
+	resp, err := http.Get(reqURL)
```
willscott: should we be able to request this through tor / domain fronting? should we have a central request manager interface to deal with that?

nettests/websites/web_connectivity.go
```diff
@@ -1,10 +1,56 @@
...
+	if err != nil {
+		return urls, errors.Wrap(err, "failed to perform request")
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return urls, errors.Wrap(err, "failed to read response body")
```
willscott: do we want to have a way to retry built in on failures to interact with the orchestrator, which may be transient?

nettests/websites/web_connectivity.go
```diff
@@ -1,10 +1,56 @@
...
+}
+
+const orchestrateBaseURL = "https://events.proteus.test.ooni.io"
+
+func lookupURLs(ctl *nettests.Controller) ([]string, error) {
+	var (
+		parsed = new(URLResponse)
+		urls   []string
+	)
+	reqURL := fmt.Sprintf("%s/api/v1/urls?probe_cc=%s",
```
willscott: add a Todo to include other config constraints here, like which categories the user is okay with,
I guess also - in the orchestrator url request, does the client include any form of ID of itself to allow linking of the different requests its making to proteus?

nettests/websites/web_connectivity.go
hellais: Yes for sure. This is just quick and dirty prototype code.

nettests/websites/web_connectivity.go
hellais: Yeah for sure. The nettest class is designed to have knowledge of the config options via `ctl.Config.Foo`, but currently config support is not complete. As soon as we have the chosen categories as part of the config, it will be possible to pass those to the orchestrator.

willscott:
My comments were mostly to suggest adding some signpost TODO's in the areas that are going to need additional work so that in the future it's clear where to hook in the additional complexity that's missing from this first pass. This looks good though 👍

nettests/websites/web_connectivity.go
hellais: I should also add, that this functionality should actually be part of measurement-kit (that of fetching URL lists from the orchestration backend), however I implemented it inside of gooni for the time being to get the whole prototype working faster.

utils/geoip.go
```diff
@@ -31,31 +31,82 @@ var geoipFiles = map[string]string{
...
-// DownloadGeoIPDatabaseFiles into the target directory
-func DownloadGeoIPDatabaseFiles(dir string) error {
-	for filename, url := range geoipFiles {
+var legacyGeoipFiles = map[string]string{
+	"GeoIPASNum.dat": "http://download.maxmind.com/download/geoip/database/asnum/GeoIPASNum.dat.gz",
+	"GeoIP.dat":      "http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz",
+}
+
+// GeoIPDir returns the geoip data dir for the given OONI Home
+func GeoIPDir(home string) string {
```
willscott: Somethings weird about having these path things split up so that geoid is in util, the db folder is in `database/database.go`,
and the creation is in `/ooni.go`.

Consider a refactor to make this a bit easier to track down, potentially?

utils/geoip.go
```diff
@@ -31,31 +31,82 @@ var geoipFiles = map[string]string{
...
+		if _, err := io.Copy(outFile, gzf); err != nil {
+			return errors.Wrap(err, "error reading file from gzip")
+		}
+		outFile.Close()
+	}
+	return nil
+}
 
-		f, err := os.Open(out.Name())
+// DownloadGeoIPDatabaseFiles into the target directory
```
willscott: This may be an issue on devices with limited memory. The in-process tgz will do worse at being able to do this - in that it isn't streaming the file in the way that a native tar process would. that said, this is nice from a compatibility perspective.

utils/geoip.go
hellais: Yeah maybe all the path stuff should go into a path package. It can't live in `ooni` or `database`, because of circular dependencies.

utils/geoip.go
hellais: These files are less than 10MB, so I don't think it's going to be an issue for this particular use case (I think OONI Probe already uses more than that memory in general).

Issue #14, opened 2018-08-01
Stabilise the SQLite data format written by the CLI


hellais:
The model we follow in the probe-cli (and desktop app) is that the tests are run by `probe-cli` and the results are written to two SQLite database tables called `results` and `measurements`.

We make the assumption that the database will only ever be written by `probe-cli`, while `probe-desktop` is only a reader of the sqlite db.

Since this does lead to some pretty tight coupling between the two apps, it's important that we minimise the amount of changes we make to the database schema, since this is kind of a shared model between the two apps.

@lorenzoPrimi and @xanscale brought up a very important issue with the current database model that I think should be addressed sooner than later.

Below is a simplified schema of the current DB:
![screen shot 2018-08-01 at 11 56 46](https://user-images.githubusercontent.com/424620/43515124-0c48c3f4-9582-11e8-978b-182cdae5fb0f.png)

Basically each Results row will have associated to it several Measurement rows. The main issue that @lorenzoPrimi and @xanscal highlighted is the misuse of the column `summary`.

`summary` is an opaque JSON column, which contains in the case of the `Results` table top level information that is to be used for creating the `Results` view (which are relevant to several measurements, ex. the number of tested websites and the number of blocked sites) and we have similar information in each of the `Measurement` summary columns.

The problem is that the `summary` in the `Results` view actually contains values that are computed based on the values of the `Measurements` table. As such it's tricky to keep these two columns in sync and we are actually duplicating some information that could otherwise be computed with a query.

The reason to follow this approach initially, was that, in particular for the sums, we didn't want to have to scan through all the rows and parse JSON from `summary` for every row (SQLite JSON support is a bit iffy: https://www.sqlite.org/json1.html). However by doing some refactoring of this table and for the counts putting everything in a dedicated `anomaly` column, we can maybe get around this.

It would be great if we were to come up with a schema that we think can work for both mobile and desktop apps and share an approach for doing this so we don't have to re-invent the wheel.

I believe this is a requirement to having a beta release as changes in this area will impact several bits of the desktop app.

xanscale:
i think we need to follow this criterias:
1) no data duplication to avoid inconsistance problems
2) head data can be generalized for all test must be outside of summary column

using this criterias i think that summary column in result can be removed and summary column in measurements can contains only test_keys

hellais:
So I believe I have come up with a reasonable model for this.

[...]

PR #16, opened 2018-08-07
Update the schema for the measurements and results tables

hellais:
Based on discussions with @xanscale and @lorenzoPrimi in:
https://github.com/ooni/probe-cli/issues/14


data/migrations/1_create_msmt_results.sql
```diff
@@ -11,36 +13,155 @@ DROP TABLE `measurements`;
...
+    -- Used to indicate if the user has seen this result
+    `is_viewed` TINYINT(1) NOT NULL,
+
+    -- This is a flag used to indicate if the result is done or is currently running.
+    `is_done` TINYINT(1) NOT NULL,
+    `data_usage_up` INTEGER NOT NULL,
+    `data_usage_down` INTEGER NOT NULL,
+    -- It's probably reasonable to set the maximum length to 260 as this is the
+    -- maximum length of file paths on windows.
+    `log_file_path` VARCHAR(260) NOT NULL
```
lorenzoPrimi: This is unused, can be removed.

data/migrations/1_create_msmt_results.sql
```diff
@@ -11,36 +13,155 @@ DROP TABLE `measurements`;
...
+    `result_id` INTEGER NOT NULL,
+    FOREIGN KEY (`result_id`) REFERENCES `results`(`id`)
+      ON DELETE CASCADE ON UPDATE CASCADE, -- If we delete a result we also want
+                                           -- all the measurements to be deleted as well.
+
+    -- This is a variable used internally to track the path to the on-disk
+    -- measurements.json. It may make sense to write one file per entry by
+    -- hooking MK and preventing it from writing to a file on disk which may
+    -- have many measurements per file.
+    `report_file_path` VARCHAR(260) NOT NULL,
```
lorenzoPrimi: Also this can be removed as in the app we use the format `performance-Measurement_id.json`

data/migrations/1_create_msmt_results.sql
```diff
@@ -11,36 +13,155 @@ DROP TABLE `measurements`;
...
+-- table.
+-- 2. (most important) We want to avoid duplicating a bunch of information that
+-- is going to be common to several networks the user is on.
+-- Example:
+-- We may wish to add to this table the location from of the probe from the GPS
+-- or add support for allowing the user to "correct" a misclassified measurement
+-- or distinguishing between wifi and mobile.
+CREATE TABLE `networks` (
+  `id` INTEGER PRIMARY KEY AUTOINCREMENT,
+  `network_name` VARCHAR(255), -- String name representing the network_name which by default is populated based
```
lorenzoPrimi: network_type is missing here

data/migrations/1_create_msmt_results.sql
hellais: On desktop it's actually more convenient to have this explicit in the DB as the paths can vary depending on the platform. I am going to keep it.

data/migrations/1_create_msmt_results.sql
hellais: Same as for the above.

data/migrations/1_create_msmt_results.sql
lorenzoPrimi: so are you going to add the report file path as well?

hellais:
I think this has reached a satisfactory level of completion. I have done some pretty extensive testing of it on macOS and some light testing on windows.

*IMPORTANT* if you were using probe-cli (or prob-desktop) before this version you *MUST* do a hard reset by running `ooni reset --force` and will loose all measurements gathered until that point.

This is due to the fact that the schema of the DB has changed and I did not want to work on a possibly complicated migration script due to this being an unreleased version.

PR #29, opened 2019-05-10
Use MK v0.10.x, zap legacy GeoIP, build also on Linux

### LINKED ISSUE #27: Upgrade to measurement-kit 0.9.1 and drop legacy maxmind db support
According to "[GeoLite Legacy databases are now discontinued](https://support.maxmind.com/geolite-legacy-discontinuation-notice/)", the URLs we use to download the GeoIP databases no longer work. The link also talks about possible migration paths.

Edit: Thanks to @chinmayisk for pointing out the problem.

bassosimone:
The only bummer is that building on macOS with Travis is gonna take forever, because we need to cross compile MK. I propose that we move forward this PR and we create issues for having binary brew packages for macOS (and Windows, also via brew). This way, we should be able to build on Travis for all the supported architectures in a reasonably reasonable time.

Closes #27.


.gitignore
```diff
@@ -1,4 +1,3 @@
-/vendor
```
hellais: Why are you dropping vendor from `.gitignore`?

.travis.yml
```diff
@@ -1,13 +1,7 @@
...
-- make download-mk-libs-macos
+os: linux
+dist: xenial
+language: minimal
+services:
+- docker
 script:
-- make test-internal
-- make build
+- ./build.sh _travis-linux
```
hellais: Is there a way we can manage to retain the ability to run, at least the unit tests, outside of the docker container?

For example by building just for the platform we are currently on and not supporting cross compilation?

Readme.md
```diff
@@ -4,61 +4,14 @@ The next generation OONI Probe Command Line Interface.
 
 ## Development setup
 
-Be sure you have golang >= 1.8.
+Be sure you have golang >= 1.12. We use Go modules. Make sure you
+don't have GOPATH set.
```
hellais: Ah I understand now that you have upgraded to use go modules. I am not too familiar with this workflow.

.gitignore
bassosimone: Because I've upgraded to go modules. (The main perk is that it's super flexible to build from anywhere, you don't need to worry about `go path` or installing `dep`.)

.travis.yml
bassosimone: > Is there a way we can manage to retain the ability to run, at least the unit tests, outside of the docker container?

What is the use case driving this question? I see quick unit tests as mainly useful when working locally, and this is enabled when you are on macOS, where you can run all tests for macOS and Windows with a very short setup time. With Linux, it's true that you need Docker, but since the script makes you build a local container, the setup time is still very low.

Regardless of your motivations, the answer to this question is yes. We need to use build tags such that everything that is not dependent on MK can run regardless of whether you have a version of MK installed. Since MK depends on `cgo`, since we are discussing a Go engine, and since there may be people that want OONI using only Go code in the future, the solution to this problem seems to me the following:

1. write a package abstracting the MK functionality;

2. (in perspective)route the calls to either the Go engine or the MK engine;

3. when `cgo` is enabled, the calls will really reach the MK engine;

4. when `cgo` is not enabled, they will cause a "not implemented"-like error.

With this scenario, you can run unit tests with:

```
CGO_ENABLED=0 go test -v ./...
```

wherever you are, without wondering about MK being installed.

There are also other possibilities, like maybe attempting to copy some MK binaries over, but since Travis is using Ubuntu and the right way(TM) to have static binaries that work on all Linux binaries is to use Alpine (which is based on musl libc), I believe this is a bloodbath that I do really like to avoid.

Anyway, before adding the tags functionality, which may still be useful in the future, I'd like you to further explain the reasoning beyond your request. My guess is that you are worried that it would take time for you to run tests when developing, but I'd rather ask than guess.

> For example by building just for the platform we are currently on and not supporting cross compilation?

Doing this will be problematic in terms of providing binaries. The current situation is a delicate Pareto efficient equilibrium between the toil of building MK and the simplicity of developing OONI in Go, where any macOS OONI developer does not see extra toil. Therefore, your above statement ("building on the platform we are currently on") is already correct for macOS.

I am afraid that, by making it more simple for Linux or Windows users to build OONI on their platform, I am going to make my life much more miserable, and I am going to have to support too many configurations and builds. On the contrary, the current scenario is optimal for me and you, because we're both on macOS, and since I'm (more) free from the building tasks, I have more time to focus on working on improving the implementation.

In going forward (but not necessarily in this PR), as I mention above, I'd really welcome a situation where we can use build tags to control the amount of C++ MK used in OONI and to increasingly use more Go code in OONI. This should be conducive to a situation in which more and more unit tests can be run in the platform in which you are, and, ideally, at a certain point anyone with Go on a system that Go supports can compile OONI.

How does that trajectory sound like?

Readme.md
bassosimone: Yes, indeed. Well, the upgrade path is very smooth. Unset `GOPATH`. Use `go build`.

hellais: Yeah, the only thing I need to figure out is how to make this play well with the several other projects I have which may not work with such recent versions of golang (I am still on 1.11).

bassosimone: I can help in doing the update! :-)

.travis.yml
hellais: > What is the use case driving this question? 

Now that I am looking at this more carefully I actually missed the fact that `build.sh` does not use a docker container to build the macOS binary, so I believe everything should work fine for me.

The main motivation is that of having the ability to run the unittests as part of my IDE every time I save a file.

I agree with you that having tags for tests is the right way to approach separating which tests can be run very frequently (i.e. unit tests) and which ones should be run less frequently (i.e. integration tests or stuff dependant on cgo maybe).

> Doing this will be problematic in terms of providing binaries

Ok I understand the points you bring up here. Will it be possible for linux developers to make a linux build of probe-cli by using docker?

I think that if that is the case then it's probably not too bad.

Another aspect to consider is the issue of building binaries for architectures other than amd64 and x86.

bassosimone: > Will it be possible for linux developers to make a linux build of probe-cli by using docker?

Yes, indeed. The process will be as fast as it us for us on macOS. Compared to working on macOS, the main annoyance is that you need a container to develop. I am not going as far as to claim that it's the same thing(TM), however, it has occurred to me in some circumstances to use a scratch container for doing development and it is doable.

Also, after writing the previous sentence, I did some research and [it seems to be possible to integrate a Docker container into your C/C++ development workflow with VSCode](https://devblogs.microsoft.com/cppblog/c-development-with-docker-containers-in-visual-studio-code/). This leads me to think that a Linux developer of OONI could setup a VSCode environment where they can run unit tests inside Docker, which may be more pleasant.

> Another aspect to consider is the issue of building binaries for architectures other than amd64 and x86.

With respect to Linux on Arm, we need an Alpine container. Since I can find [pretty recent Alpine images for armv7](https://hub.docker.com/r/balenalib/kitra520-alpine-golang/tags), I would say that there is hope for cross compiling over there. The Dockerfile I am using in script-build-linux should be good enough. Also, I know that @critzo has played around quite a lot with cross compiling libndt in Docker, so here he maybe can shed some more light.

With respect to Windows, my understanding is that [less than 1% of gamers use a 32 bit version of Windows](https://store.steampowered.com/hwsurvey/), but I'm a bit in the dark with respect to relating this notion to our potential user base.

In all cases, I'd rather be happy knowing that something is doable, invest energies to make it possible for us _not_ to do that with too much effort (i.e. use Go rather than cross compile), and do that in the moment in which there is need for this support.

hellais:
I created an issue for the Makefile build.sh script merge: https://github.com/ooni/probe-cli/issues/31 and I was not able to reproduce the issue with linux.

PR #42, opened 2019-05-22
Use ooni/probe-engine: episode 1

bassosimone:
In this first episoed of the ooni/probe-engine integration series I start by adding a `Session` (the "root" object of the probe-engine) to the `Context` (the "root" object of the probe-cli).

The addition of a `Session` allows us to perform a variety of operations (e.g. downloading assets, getting the path of assets, performing the probe lookup) using the `Session`.

In turn, this allows to cleanup some code in here.

Of course, it's not that code disappeared. Rather the code that was here has been moved and refactored into the github.com/ooni/probe-engine repository.


nettests/nettests.go
```diff
@@ -90,26 +93,22 @@ func (c *Controller) Init(nt *mk.Nettest) error {
...
+		ProbeCC:  c.Ctx.Session.ProbeCC(),
+		ProbeASN: c.Ctx.Session.ProbeASNString(),
+		ProbeIP:  c.Ctx.Session.ProbeIP(),
 
 		DisableReportFile: false,
 		DisableCollector:  !c.Ctx.Config.Sharing.UploadResults,
 		RandomizeInput:    false, // It's important to disable input randomization to ensure the URLs are written in sync to the DB
 		SoftwareName:      "ooniprobe-desktop",
-		SoftwareVersion:   ooni.Version,
+		SoftwareVersion:   version.Version,
```
hellais: IRC there was something that wrote the version number using `ldflags` at build time (or at least there was an idea of doing that).

See: https://goreleaser.com/environment/#using-the-main-version

go.mod
```diff
@@ -16,13 +16,12 @@ require (
 	github.com/mattn/go-colorable v0.0.9
 	github.com/mattn/go-isatty v0.0.7 // indirect
 	github.com/mattn/go-sqlite3 v1.10.0 // indirect
-	github.com/measurement-kit/go-measurement-kit v0.0.0-20190509222408-0d0cd3233336
-	github.com/oschwald/geoip2-golang v1.2.1
-	github.com/oschwald/maxminddb-golang v1.3.0 // indirect
+	github.com/measurement-kit/engine v0.0.0-20190516143312-a38282337fe6 // indirect
+	github.com/measurement-kit/go-measurement-kit v0.0.0-20190521082856-635e836bbb9d
+	github.com/ooni/probe-engine v0.0.0-20190522105151-0c60545f09ea
```
hellais: What is the difference between `measurement-kit/engine` & `ooni/probe-engine`?

ooni.go
```diff
@@ -123,12 +91,18 @@ func (c *Context) Init() error {
 	return nil
 }
 
-// NewContext instance.
+// NewContext creates a new context instance.
```
hellais: I see what you mean with the context name being a bit confusing.

Perhaps this can all end up inside of Session?

internal/cli/geoip/geoip.go
```diff
@@ -1,47 +1,33 @@
...
-	shouldUpdate := cmd.Flag("update", "Update the geoip database").Bool()
-
 	cmd.Action(func(_ *kingpin.ParseContext) error {
 		output.SectionTitle("GeoIP lookup")
 		ctx, err := root.Init()
 		if err != nil {
 			return err
 		}
 
-		if err = ctx.MaybeDownloadDataFiles(); err != nil {
```
hellais: How is the download of the data files now being handled?

Is it done as part of `MaybeLocationLookup`?

hellais:
I left some comments in the PR. The CI appears to be failing. We should fix that before merging.

nettests/nettests.go
bassosimone: Yeah, doing that was also my intention. Do you want me to open an issue?

go.mod
bassosimone: I needed to re-run `go mod tidy`, sorry for the confusion :-)

ooni.go
bassosimone: I think not all of this can, as there are probe specific bits. I think we should continue this integration of probe-engine and, in the end, rename `Context` maybe `Probe`.

Importing `ooni` and using `ooni.Probe`would not surprise me for a tool called `ooniprobe`.

Actually, I do not see fields in `Context` that it makes sense moving to engine, but:

```Go
// Probe is a ooniprobe instance
type Probe struct {
	Config   *config.Config          // probe specific and `config.json` specific
	DB       sqlbuilder.Database   // definitely probe specific
	IsBatch  bool                           // probe specific: the engine doesn't have a real logger
	Session  *session.Session

	Home    string  // probe specific: but the engine uses a subdir
	TempDir string  // not sure what's the purpose of this :-)

	dbPath     string // also probe specific
	configPath string // ditto: I see the engine configured by calling code
}
```

the final name seems actually nice, no? (I've added comments explaining why I think there is no need to move more variables to the engine currently)

internal/cli/geoip/geoip.go
bassosimone: Yes. You give the engine an existing directory. The engine makes sure there are assets in this directory matching the sha256 it knows. If they're missing or old, a new download starts.

In the common case, after checking the sha256, there's nothing else to do.

bassosimone:
@hellais I've fixed the CI. Or, well, I may have fixed it, but we're still seeing issues like:

```
go: finding github.com/alecthomas/kingpin v2.2.6+incompatible
go: github.com/Netflix/go-expect@v0.0.0-20180615182759-c93bf25de8e8: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /Users/travis/go/pkg/mod/cache/vcs/09039755419264a66e5df43bb18c8ae637e56854873f932871794d0de9ad1855: exit status 128:
	fatal: unable to access 'https://github.com/Netflix/go-expect/': Could not resolve host: github.com
Fetching https://golang.org/x/sync?go-get=1
https fetch failed: Get https://golang.org/x/sync?go-get=1: dial tcp: lookup golang.org: no such host
go: golang.org/x/sync@v0.0.0-20190412183630-56d357773e84: unrecognized import path "golang.org/x/sync" (https fetch: Get https://golang.org/x/sync?go-get=1: dial tcp: lookup golang.org: no such host)
```

The "no such host" makes me think that Travis made a boo boo. Restarting the builds.

nettests/nettests.go
hellais: Yes let's open an issue for that.

ooni.go
hellais: Ah yes probe could also be a good name!

hellais:
I restarted the CI. For me it's good to be merged as soon as the CI is all green and we have filed the issue for the version number.

nettests/nettests.go
bassosimone: See https://github.com/ooni/probe-cli/issues/43

ooni.go
bassosimone: See https://github.com/ooni/probe-cli/issues/45

Issue #44, opened 2019-05-23
Use github.com/ooni/probe-engine for running tests


bassosimone:
Started working on this in #42. It will likely require 2/3 steps to complete.

bassosimone:
This is done in #42. Review pending.

bassosimone:
Now working on #46 to move forward this story.

Issue #45, opened 2019-05-23
refactor: rename Context => Probe


bassosimone:
To be done after #44

hellais:
I think that as part of this work, we should also look into simplifying the `Controller` related logic, probably ditching it entirely in favour of something simpler and also looking into using a cancellable `context.Context` for handling running tests, as mentioned by @bassosimone in his review of: https://github.com/ooni/probe-cli/pull/84.

bassosimone:
> I think that as part of this work, we should also look into simplifying the Controller related logic, probably ditching it entirely in favour of something simpler and also looking into using a cancellable context.Context for handling running tests

Yes, good idea. I guess some simplifications will be easier for tests written in Go, but this needs to be verified by taking the time to look into the code and reasoning about that more.

PR #46, opened 2019-05-24
Integrate further with ooni/probe-engine: episode two

bassosimone:
1. use ooni/probe-engine for contacting the bouncer, using the collector

2. use ooni/probe-engine for running the IM tests

3. DO NOT use it for other tests for now (to keep the diff small and ease comparison between tests using the ooni/probe-engine and tests using measurement-kit/go-measurement-kit)

I'd like @hellais to review. If this is fine, we can either continue the work in this branch and making all tests use ooni/probe-engine or merge and open a new PR for other tests.

Related issue #44


internal/cli/run/run.go
```diff
@@ -95,6 +96,17 @@ func init() {
 			log.WithError(err).Error("Failed to create the network row")
 			return err
 		}
+		if ctx.Config.Advanced.BouncerURL != "" {
+			ctx.Session.SetAvailableHTTPSBouncer(ctx.Config.Advanced.BouncerURL)
```
hellais: We may want in the future to support non-https bouncers (ex. onion service bouncers) and we may have to adjust this code here. Perhaps the function call on `Session` can be `SetAvailableBouncer` or will we do the switch in here based on the target bouncer format?

nettests/im/whatsapp.go
```diff
@@ -25,7 +28,12 @@ type WhatsAppTestKeys struct {
 }
 
 // GetTestKeys generates a summary for a test run
-func (h WhatsApp) GetTestKeys(tk map[string]interface{}) (interface{}, error) {
+func (h WhatsApp) GetTestKeys(otk interface{}) (interface{}, error) {
+	tk, ok := otk.(map[string]interface{})
+	if !ok {
+		return nil, errors.New("Unexpected test keys format")
```
hellais: Maybe we want to do this check (which is done for every test) when `GetTestKeys` is called to avoid duplicating it for every test?

nettests/nettests.go
```diff
@@ -273,6 +279,117 @@ func (c *Controller) Init(nt *mk.Nettest) error {
...
+	ctx := context.Background()
+
+	// This will configure the controller as handler for the callbacks
+	// called by ooni/probe-engine/experiment.Experiment.
+	exp.Callbacks = handler.Callbacks(c)
+
+	c.msmts = make(map[int64]*database.Measurement)
+
+	// These values are shared by every measurement
+	reportID := sql.NullString{String: "", Valid: false}
```
hellais: `<bikeshed>`I generally use the pattern `var reportID sql.NullString` for similar situations. It's more clear that I am initialising the variable and I am not actually setting the values to anything meaningful`</bikeshed>`

nettests/nettests.go
```diff
@@ -273,6 +279,117 @@ func (c *Controller) Init(nt *mk.Nettest) error {
...
+
+	log.Debugf("status.end")
+	for idx, msmt := range c.msmts {
+		log.Debugf("adding msmt#%d to result", idx)
+		if err := msmt.AddToResult(c.Ctx.DB, c.res); err != nil {
+			return errors.Wrap(err, "failed to add to result")
+		}
+	}
+	return nil
+}
```
hellais: I believe that some of the logic inside of this `Run` method is also present inside of `Init` which I see is no longer getting called. I suggest dropping the `Init` method as part of this PR as well.

internal/cli/run/run.go
bassosimone: My suggestion: let's handle this problem in the future? It's not super clear to me what benefits an onion service would add to our users compared to the current situation where the engine already knows how to proxy any non measurement traffic (including `mlab-ns`) through any proxy that speaks socks5 (hence also including tor) and uses such proxy service to somehow reach our services exiting on the plain old internet and using https.

nettests/im/whatsapp.go
bassosimone: With the current code, yes. More specifically, with the current code this is a bit of a bummer because all test keys are actually `map[string]interface{}`. Yet, we have this issue:

1. the `map[string]interface{}` is how Go parses a JSON when you don't tell it anything about its structure, hence the typecast here is always gonna succeed;

2. however, for tests that use Go (`ndt7` and `psiphon`, which are already implemented into this engine but not added to this pull request), it's not true that we can cast to `map[string]interface{}` because they're actually Go data structures.

So, my instinct was that we either (1) serialised to JSON also the test keys of other tests and then got `map[string]interface{}` or (2) that the interface of each test needed to be changed and the type cast pushed down to the specific function.

At the same time, yes, this code duplication is annoying. Do you want me to create another method to the API that takes a `map[string]interface{}`, move the cast in generic code, and choose such method or the current one taking `interface{}` depending on the result of the typecast? This would probably tidy up the code a bit, even though it would add one more function to the set of functions in the `Nettest` interface.

nettests/nettests.go
bassosimone: Yes, I also use the same pattern as it's more idiomatic Go. This code has been copied from existing code in the `Init` function and didn't change it because I wanted to keep it easier for one to see how the code has changed in this refactoring process.

nettests/nettests.go
bassosimone: I cannot drop it until all tests have been converted to use `Run` rather than `Init`. Currently, I just converted the IM tests. If you're okay with this approach and feel like I succeeded in preserving the meaning of the original code in `Run`, then I'll be happy to follow up and make sure all the tests use `Run` rather than `Init`.

internal/cli/run/run.go
hellais: Sounds reasonable. Just though I would bring it up.

bassosimone: Yeah, this seems like something we want to have a ticket for. I'll create one.

nettests/nettests.go
hellais: I see. Yes I think this approach is valid and I would suggest we drop all the `Init` code and update the tests to use `Run` instead.

nettests/im/whatsapp.go
hellais: What if you make the return value of `ndt` and `psiphon` also return `map[string]interface{}` instead of a golang data struct?

nettests/nettests.go
```diff
@@ -273,6 +279,117 @@ func (c *Controller) Init(nt *mk.Nettest) error {
...
+	reportID := sql.NullString{String: "", Valid: false}
+	resultID := c.res.ID
+
+	log.Debug(color.RedString("status.queued"))
+	log.Debug(color.RedString("status.started"))
+	log.Debugf("OutputPath: %s", c.msmtPath)
+
+	// TODO(bassosimone): double check that we're passing the right options
+	// to MK? We may be passing less option than needed currently, even though
+	// the code works for the common configuration.
```
bassosimone: Note to self: this should either be addressed in this PR or formalised as an issue.

nettests/im/whatsapp.go
bassosimone: My immediate instinct is that I don't like this strategy because it would make the implementation of the engine much less obvious to accomodate for an implementation detail of the cli.

I mean, tests written in Go would need to serialise the data structure they return as JSON, then parse such JSON again into a `var v interface{}`, only to store that `v` into `TestKeys` such that `TestKeys` is in the format expected by the cli.

Especially in a scenario where we'll have more and more tests written in Go, this seems to me the sort of complexity that we'll deeply regret adding in the future.

It may be that there is a better way. But I need to spend more time thinking. I believe it may be worth asking the questions whether:

1. we should have in the engine an exported data structure for each experiment written in C++ and parse the results returned by MK into such data structure, then setting `TestKeys` to such data structure, so that later the CLI just needs to perform a type assertion to the test specific structure to get an already organised view of the results (this would also solve some other `XXX` that I have seen in the probe-cli code)

2. find a way for the experiment to return directly the specific data structure rather than making it opaque by setting the `TestKeys` (but this seems to me more of a bummer because we'll lose the generality of running all experiments using generic code)

Do you have any bad feelings towards option 1?

bassosimone: (Oh, and @hellais, I think we've hit here the place where the design of experiments written in Go  collides the most with experiments written in C++, so I'm quite glad we've seen this issue now rather than some months in the future, and super good we're discussing it.)

bassosimone: So, after thinking a bit more, I guess the right approach is the following: the engine will return a serialised engine as the result of a measurement. The cli will parse such JSON overlying on top of it the keys to which it's interested. This preserves the previous behaviour (simplifies this PR among other things) and is also nice because it reduces the data coupling between the engine and the cli. We don't actually need to have a full understanding of the JSON structure in the cli, rather we just need to understand some fields. So with this design it seems to me we're hitting a sweet spot where everything ends up being simpler.

hellais: Yes that approach seems reasonable

bassosimone: So, returning a serialised JSON is not feasible. Will make the engine API bad and force to do dirty unclean stuff. It's clearly not the right design.

At the same time it does not feel he right design to discard by default the result of a test as a well defined Go data structure. We're currently suffering from bad data schema for more or less all our measurements and we have complexity everywhere because of that. So, we should instead strive to have more well defined data structures. When more tests will be written in Go, we'll have more precise data structures, because Go kinda forces that.

The right solution seems most likely to add glue to the cli that converts the test keys to the format that the CLI wants, forced by the way in which MK works, and, when we'll have all tests written in Go, get rid of such glue and just cast `interface{}` to the proper data structure of each test. An issue should also be created as a reminder.

I'm playing with code doing that.

internal/cli/run/run.go
bassosimone: Here: https://github.com/ooni/probe-cli/issues/47

nettests/nettests.go
bassosimone: Done!

bassosimone:
(We should cancel the Travis webhook to have two builds rather than four per commit.)

internal/enginex/enginex.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Warn emits a warning message.
+func (la LoggerAdapter) Warn(msg string) {
+	la.Logger.Warn(color.RedString("engine") + ": " + msg)
+}
+
+// Warnf formats and emits a warning message.
+func (la LoggerAdapter) Warnf(format string, v ...interface{}) {
+	la.Logger.Warnf(color.RedString("engine")+": "+format, v...)
+}
```
hellais: Is there a reason to do the formatting of the log messages inside of here instead of inside of a apex/log handler: https://github.com/ooni/probe-cli/blob/master/internal/log/handlers/cli/cli.go

nettests/nettests.go
```diff
@@ -67,252 +74,148 @@ func (c *Controller) SetNettestIndex(i, n int) {
...
-	if err != nil {
-		return err
-	}
+// Run runs the selected nettest using the related experiment
+// with the specified inputs.
+//
+// This function will continue to run in most cases but will
+// immediately halt if something's wrong with the file system.
+func (c *Controller) Run(exp *experiment.Experiment, inputs []string) error {
+	ctx := context.Background()
```
hellais: Maybe we can call this `background` or something similar to avoid confusing it with `c.Ctx` (we ought to eventually rename `c.Ctx` to something better)?

nettests/nettests.go
```diff
@@ -50,6 +51,12 @@ type Controller struct {
 	msmts       map[int64]*database.Measurement
 	msmtPath    string          // XXX maybe we can drop this and just use a temporary file
 	inputIdxMap map[int64]int64 // Used to map mk idx to database id
+
+	// numInputs is the total number of inputs
+	numInputs int
+
+	// curInputIdx is the current input index
+	curInputIdx int
```
hellais: Why don't you define these two as int64 so that the int size is not platform dependent and you can avoid casting it to `int64` when passing it to `inputIdxMap`?

internal/enginex/enginex.go
bassosimone: The probe-engine does not depend on apex/log, rather it expects you to pass it [a structure that implements an interface that apex.log.Log happens to match](https://github.com/ooni/probe-engine/blob/79338ce6ff54e6497c9c4dfcb79d6f726269bcfc/log/log.go).

The pre-existing code was [filtering the logs emitted by MK and prefixing them with a coloured string](https://github.com/ooni/probe-cli/pull/46/files#diff-eacd810feae41332a4946c2af72b0b5aL118).

This change is my attempt at preserving the previously existing behaviour by making sure that messages emitted by probe-engine are still prefix.

I'd rather avoid depending on apex/log (and God forbids colouring messages) in the probe engine, but I think the following two questions are relevant to your discussion:

1. do we need to colour messages emitted by the engine during tests?

2. if the answer to the first question is affirmative, is there an apex/log way to construct an apex/log structure that (a) matches the required interface and (b) implements the desired behaviour that is more manageable than my implementation?

nettests/nettests.go
bassosimone: I'd rather keep it named `ctx` and perform the `c.Ctx` renaming as a follow up. We discussed calling it `Probe`, and hence `probe` when instantiated. I wanted, at the same time, to avoid starting a refactoring such as that one, which is mainly mechanic, with this PR pending.

internal/enginex/enginex.go
hellais: tbh 1. is mostly an aestetic question. I think it would be nice, but I would not say we **need** it.

To do 2 inside of apex/log it's actually pretty simple (and as a matter of fact we already do it). When you call apex.log with a certain log level you get a `e.Level` which is the log level and you can color the message based on that, see: https://github.com/ooni/probe-cli/blob/master/internal/log/handlers/cli/cli.go#L134

nettests/nettests.go
bassosimone: That's a good question. The direct answer is that Go has a notion of integer that depends on the native size of registers on the platform. This is reasonable because `int64` operations are expensive when registers are actually `32` bit wide.

Because of this performance improving design choice, Go uses `int` as index. Therefore, I find it mostly natural that the code uses `int` as index.

Unless there is a way to force to to return an `int64` index when `range`-ing over a slice, I do not think it is wise to switch these values to `int64` because we would anyway need to case the loop index to an int64.

I believe the reason why we're using `int64` here is mostly caused by the fact that MK was using a `int64` to represent indices in JSON messages. But that's a different story because JSON defines the integer as `int53`, so there was no choice but define that field as being 64 bit wide.

However, now that we're integrating more and more Go code with Go code, keeping the 64 bit indexing structure seems to me a legacy implementation detail.

If Go on a specific platform cannot handle offsets larger than 32 bit, there is no point for our Go code interfacing with Go code to cast to 64 bit.

There may however be specific database reasoning of which I'm not aware that may suggest to use 64 bits integer for representing offsets in the database.

If that is the case, then I'd recommend to refactor the code such that we perform the cast to `int64` when we're as close as possible to the database related code, and just use `int` in all the rest of the code to avoid unnecessary casts.

Otherwise, and especially because we've not made any public release yet, I would say that we would be better off by just always using `int`, including in the database.

This seems to me the best way to address this issue. (Incidentally, this is consistent with the Google code style that advocates for using the naturally sized integers on any platform, unless there are specific needs to user a specific integer size, i.e. for binary protocols - since Go originated at Google, I can see how this recommendation transpired into the language and I'd argue that following along is the idiomatic way to go in this case.)

nettests/nettests.go
bassosimone: (Just double checked that we've a tracking issue: https://github.com/ooni/probe-cli/issues/45)

nettests/nettests.go
hellais: Then maybe we should change to int64 in the index of `inputIdxMap`?

internal/enginex/enginex.go
bassosimone: 1. Right. I actually like this aesthetic feature.

2. Does this mean that there is a specific logger that I should pass probe engine here or that I can obtain the same result by constructing a new logger?

nettests/nettests.go
bassosimone: Do you mean changing to `int`? If so, yes, I was considering on doing that. But I did not perform this action as part of this PR, as the diff was quite messy already.

Is it fine if I create an issue for that?

internal/enginex/enginex.go
hellais: > Does this mean that there is a specific logger that I should pass probe engine here or that I can obtain the same result by constructing a new logger?

 You just create any new instance of apex/log and then use it with it's log levels to obtain the desired result.

Maybe you want to give it an extra key to indicate it's coming from probe-engine. See as an example: https://github.com/ooni/probe-cli/blob/master/internal/output/output.go#L129

So you could do:

```
	log.WithFields(log.Fields{
		"type":            "engine_log",
	}).Info("Something")
```

nettests/nettests.go
hellais: Sure let's make an issue for it.

internal/enginex/enginex.go
bassosimone: OK, so, as we just confirmed on Slack, I need to pass:

```Go
log.WithFields(log.Fields{
  "type": "engine",
})
```

to the probe-engine. This will give me a logger with fields. And then we can build on that to continue to have the coloured first entry as we used to have.

nettests/nettests.go
bassosimone: See #48

bassosimone:
@hellais I should have addressed all your comments. Please, take another look.

internal/log/handlers/cli/cli.go
```diff
@@ -104,6 +104,9 @@ func logTable(w io.Writer, f log.Fields) error {
 // TypedLog is used for handling special "typed" logs to the CLI
 func (h *Handler) TypedLog(t string, e *log.Entry) error {
 	switch t {
+	case "engine":
+		fmt.Fprintf(h.Writer, "[engine] %s\n", e.Message)
+		return nil
```
hellais: Yes this is good and how I was thinking it could be done.

The benefit is that we can also filter through these in the JSON based output for usage inside of the desktop app.

nettests/nettests.go
```diff
@@ -67,252 +74,148 @@ func (c *Controller) SetNettestIndex(i, n int) {
...
-		log.Debugf(color.RedString(e.Key))
-
-		if err := c.msmts[e.Value.Idx].Done(c.Ctx.DB); err != nil {
-			log.WithError(err).Error("failed to mark msmt as done")
+		// Make sure we share what the user wants us to share.
+		if c.Ctx.Config.Sharing.IncludeIP == false {
+			measurement.ProbeIP = model.DefaultProbeIP
+		}
+		if c.Ctx.Config.Sharing.IncludeASN == false {
+			measurement.ProbeASN = fmt.Sprintf("AS%d", model.DefaultProbeASN)
```
hellais: Why is the `DefaultProbeASN` not a string with the `AS` prefix directly?

hellais:
I took another look at the code and it looks good to me.

I haven't tested it yet, but I will do that now.

nettests/nettests.go
bassosimone: Because the code model used in `ooni/probe-cli` used an `int` for that. I believe we can however define a `model.DefaultProbeASNString` constant in `probe-engine`, if you prefer that we have a canonical constant for that. Do you want me to do that?

internal/log/handlers/cli/cli.go
bassosimone: Right, then I guess I can resolve this bit of the conversation.

nettests/websites/web_connectivity.go
```diff
@@ -72,17 +41,16 @@ type WebConnectivity struct {
...
-
 	urls, urlIDMap, err := lookupURLs(ctl)
 	if err != nil {
 		return err
 	}
 	ctl.SetInputIdxMap(urlIDMap)
-	nt.Options.Inputs = urls
-
-	return nt.Run()
+	experiment := web_connectivity.NewExperiment(
```
hellais: I believe the logic of how this test now runs has significantly changed. It now does a GeoIP lookup for every URL and opens a report for each measurement.
This significantly increases the overhead of running a test and the progress in it is also not as clear.

We should see if we can make this change in such a way that it implements the same flow as before.

bassosimone: What leads you to say so? I'm using MK v0.10.4 and I cannot see this behaviour. The geolookup and report management functionality are now delegated to the Go engine and MK is explicitly configured to basically just run the measurement. However, it may be that with an older version of MK some functionality is duplicated because I've added support for disabling more functionality only in v0.10.4.

Please, see below an excerpt of the verbose log where I dissect what happens. This seems to me a long but also data-centred way of understanding if we're doing in MK some bits of processing that we should not be doing now. (Spoiler alert: I found a couple of issues that need fixing but AFAICT we're not doing unnecessary geolookups or collector calls.) While there, I have also added comments on how we could further improve the performance.

First, ooni/probe-cli starts up and prints some information:

```
   • ooni version 3.0.0-beta.3
   • Reading default config file
   • Connecting to database sqlite3:///Users/sbasso/.ooni/db/main.sqlite3
   • running migrations       
   • performed 0 migrations   
   • Using collector:         
   • Using bouncer:           
```

Then, ooni/probe-engine performs the GeoIP lookup:

```
[engine] iplookup: using akamai
[engine] dns: resolving a248.e.akamai.net
[engine] dns: got 2 entries
[engine] - 92.123.101.210
[engine] - 84.53.177.17
[engine] connect: using tcp, 92.123.101.210:443
[engine] connect: connected to tcp, 92.123.101.210:443
[engine] tls: starting handshake
[engine] tls: handshake OK
[engine] - negotiated protocol: http/1.1
[engine] - version: TLSv1.2
[engine] http: connection to 92.123.101.210:443 ready; sending request
[engine] > GET / HTTP/1.1
[engine] > User-Agent: ooniprobe-desktop/3.0.0-beta.3
[engine] > Accept-Encoding: gzip
[engine] > Host: whatismyip.akamai.com
[engine] >
[engine] http: request sent; waiting for response
[engine] http: start receiving response
[engine] < HTTP/1.1 200 OK
[engine] < Expires: Sun, 16 Jun 2019 07:54:45 GMT
[engine] < Cache-Control: max-age=0, no-cache, no-store
[engine] < Pragma: no-cache
[engine] < Date: Sun, 16 Jun 2019 07:54:45 GMT
[engine] < Connection: keep-alive
[engine] < Content-Type: text/html
[engine] < Content-Length: 14
[engine] <
[engine] iplookup: IP: 109.116.253.87
[engine] mmdblookup: ASN: &{AutonomousSystemNumber:30722 AutonomousSystemOrganization:Vodafone Italia S.p.A.}
[engine] mmdblookup: Country: &{Country:{IsoCode:IT ...}
```

Then, ooni/probe-engine queries the bouncer for backends (if we want this to be more efficient, we should probably also have a single API call returning all the info):

```
[engine] jsonapi: method: GET
[engine] jsonapi: URL: https://bouncer.ooni.io/api/v1/collectors
[engine] dns: resolving bouncer.ooni.io
[engine] dns: got 1 entries
[engine] - 103.104.244.87
[engine] connect: using tcp, 103.104.244.87:443
[engine] connect: connected to tcp, 103.104.244.87:443
[engine] tls: starting handshake
[engine] tls: handshake OK
[engine] - negotiated protocol: http/1.1
[engine] - version: TLSv1.2
[engine] http: connection to 103.104.244.87:443 ready; sending request
[engine] > GET /api/v1/collectors HTTP/1.1
[engine] > Accept-Encoding: gzip
[engine] > User-Agent: ooniprobe-desktop/3.0.0-beta.3
[engine] > Host: bouncer.ooni.io
[engine] >
[engine] http: request sent; waiting for response
[engine] http: start receiving response
[engine] < HTTP/1.1 200 OK
[engine] < Date: Sun, 16 Jun 2019 07:54:46 GMT
[engine] < Content-Type: application/json
[engine] < Content-Length: 466
[engine] < Connection: keep-alive
[engine] < Keep-Alive: timeout=120
[engine] < Etag: "f78333f849c8eaa385ba2fc03fcb69c758585c6b"
[engine] < Server: nginx
[engine] <
[engine] jsonapi: response body: [
    {"type": "https", "address": "https://c.collector.ooni.io:443"}
, ...]
[engine] jsonapi: method: GET
[engine] jsonapi: URL: https://bouncer.ooni.io/api/v1/test-helpers
[engine] http: connection to 103.104.244.87:443 ready; sending request
```

However, note how using a global HTTP client is nice because here we're clearly reusing the connection used for calling the previous API (so we already save some round-trips).

```
[engine] > GET /api/v1/test-helpers HTTP/1.1
[engine] > Host: bouncer.ooni.io
[engine] > Accept-Encoding: gzip
[engine] > User-Agent: ooniprobe-desktop/3.0.0-beta.3
[engine] >
[engine] http: request sent; waiting for response
[engine] http: start receiving response
[engine] < HTTP/1.1 200 OK
[engine] < Server: nginx
[engine] < Date: Sun, 16 Jun 2019 07:54:47 GMT
[engine] < Content-Type: application/json
[engine] < Content-Length: 1126
[engine] < Connection: keep-alive
[engine] < Keep-Alive: timeout=120
[engine] < Etag: "953c84ce7845161ff0c99fcb881e4efdebfa6b5e"
[engine] <
[engine] jsonapi: response body: {
    "traceroute": [{"type": "legacy", "address": "37.218.247.110"},  ...
]}
```

Then, ooni/probe-cli prepares for running `websites`:

```
   • Running websites         
   • Running test group Websites
   • Creating result {
    0 websites 2019-06-16 07:54:47.128591 +0000 UTC
    16 0 false false 0 0
    $HOME/.ooni/msmts/websites-2019-06-16T075447.128591Z}
   • Running test websites.WebConnectivity
```

As part of this, ooni/probe-engine queries the test-lists API (if this API was served by the same domain as above, this would mean continuing to reuse the same connection):

```
[engine] jsonapi: method: GET
[engine] jsonapi: URL: https://events.proteus.ooni.io/api/v1/urls?probe_cc=IT
```

Incidentally, this should be `events.ooni.io` (I probably need to `go get -v`):

```
[engine] dns: resolving events.proteus.ooni.io
[engine] dns: got 1 entries
[engine] - 37.218.242.63
[engine] connect: using tcp, 37.218.242.63:443
[engine] connect: connected to tcp, 37.218.242.63:443
[engine] tls: starting handshake
[engine] tls: handshake OK
[engine] - negotiated protocol: http/1.1
[engine] - version: TLSv1.2
[engine] http: connection to 37.218.242.63:443 ready; sending request
[engine] > GET /api/v1/urls?probe_cc=IT HTTP/1.1
[engine] > User-Agent: ooniprobe-desktop/3.0.0-beta.3
[engine] > Host: events.proteus.ooni.io
[engine] > Accept-Encoding: gzip
[engine] >
[engine] http: request sent; waiting for response
[engine] http: start receiving response
[engine] < HTTP/1.1 200 OK
[engine] < Connection: keep-alive
[engine] < Server: nginx/1.6.2
[engine] < Date: Sun, 16 Jun 2019 07:54:47 GMT
[engine] < Content-Type: application/json; charset=utf-8
[engine] <
[engine] jsonapi: response body: {"metadata":{
    "count":-1,"current_page":-1,"limit":100,"next_url":"","pages":-1
},"results":[{
    "category_code":"COMT",
    "url":"https://www.iconnecthere.com/","country_code":"XX"},
...}]}
```

Then ooni/probe-cli processes all the URLs:

```
   • Going over URL 0         
   • returning url 198        
   • Mapped URL https://www.iconnecthere.com/ to idx 0 and urlID 198
   • Going over URL 1         
   • returning url 199        
[...]
   • Going over URL 99        
   • returning url 283        
   • Mapped URL http://emchurch.org/ to idx 99 and urlID 283
```

Then, the nettest proper starts (note that `status.queue` and `status.start` are not MK events but I've kept the output consistent with what we used to emit):

```
   • status.queued            
   • status.started           
   • OutputPath: /.../ooni283065313/msmt-websites.WebConnectivity-2019-06-16T075447.134412Z.jsonl
```

Then, the ooni/probe-engine is again invoked to open the report (same consideration as above with respect to possibly reusing connections if reusing domain names):

```
[engine] experiment: unsupported collector type: onion
[engine] jsonapi: request body: {
  "probe_asn":"AS30722",
  "probe_cc":"IT",
  "software_name":"ooniprobe-desktop",
  "software_version":"3.0.0-beta.3",
  "test_name":"web_connectivity",
  "test_version":"0.0.1"
}
[engine] jsonapi: method: POST
[engine] jsonapi: URL: https://c.collector.ooni.io:443/report
[engine] dns: resolving c.collector.ooni.io
[engine] dns: got 1 entries
[engine] - 37.218.242.210
[engine] connect: using tcp, 37.218.242.210:443
[engine] connect: connected to tcp, 37.218.242.210:443
[engine] tls: starting handshake
[engine] tls: handshake OK
[engine] - negotiated protocol: http/1.1
[engine] - version: TLSv1.2
[engine] http: connection to 37.218.242.210:443 ready; sending request
[engine] > POST /report HTTP/1.1
[engine] > Content-Length: 163
[engine] > Content-Type: application/json
[engine] > Host: c.collector.ooni.io:443
[engine] > Accept-Encoding: gzip
[engine] > User-Agent: ooniprobe-desktop/3.0.0-beta.3
[engine] >
[engine] http: request sent; waiting for response
[engine] http: start receiving response
[engine] < HTTP/1.1 200 OK
[engine] < Server: nginx/1.6.2
[engine] < Date: Sun, 16 Jun 2019 07:54:47 GMT
[engine] < Content-Type: application/json
[engine] < Content-Length: 159
[engine] < Connection: keep-alive
[engine] < Keep-Alive: timeout=120
[engine] <
[engine] jsonapi: response body: {
  "backend_version": "1.3.5",
  "supported_formats": ["yaml", "json"],
  "report_id": "..."
}
```

At this point (and excuse me for this long message, but I wanted to be sure what I thought and the code behaviour were consistent), we finally invoke MK. I will dissect the settings that we're passing to it, which is functional to explaining what is happening:

```
   • status.report_create     
   • status.measurement_start 
[engine] measurementkit: settings: {
```

As you can see, we're invoking MK's Web Connectivity for a single input. (The log level is DEBUG because I wanted to show you more logs here.)

```
  "inputs":["https://www.iconnecthere.com/"],
  "log_level":"DEBUG",
  "name":"WebConnectivity",
```

Note how we're passing to MK the backend that we discovered using ooni/probe-engine and how we're also passing the paths to the assets downloaded by the engine:

```
  "options":{
    "backend":"https://c.web-connectivity.th.ooni.io:443",
    "net/ca_bundle_path":"$HOME/.ooni/assets/ca-bundle.pem",
    "geoip_asn_path":"$HOME/.ooni/assets/asn.mmdb",
```

Note how we disable bouncer, collector, saving report to disk, and GeoIP lookup:

```
    "no_bouncer":true,
    "no_collector":true,
    "no_file_report":true,
    "no_geoip":true,
```

Note how we're passing downstream all the information that we have:

```
    "probe_asn":"AS30722",
    "probe_cc":"IT",
    "probe_ip":"109.116.253.87",
    "probe_network_name":"Vodafone Italia S.p.A.",
```

That we're passing `save_real_probe_ip` equal to `true` looks like a bug to me, in the sense that this disables the IP scrubbing performed by MK.

```
    "save_real_probe_ip":true,
    "software_name":"ooniprobe-desktop",
    "software_version":"3.0.0-beta.3"
}}
```

Then MK will perform the resolver lookup. This is actually performed for each measurement and I believe this is a bug and we should do it only once in the engine:

```
[engine] getaddrinfo('whoami.akamai.net') => error: code=0, reason=''
[engine] ip address of resolver is 173.194.169.102
```

Then there is the measurement proper. MK runs three measurement in parallel, but since we're passing it just one entry, notice how two threads out of three immediately reach the end of the input and basically terminate. (We should discuss whether we want parallelism in ooni/probe-cli or not; for now it is not implemented because it would have complicated too much the review of this diff, but clearly a subsequent PR can improve the situation, if we want.)

```
   • OnProgress: 0.100000 - starting the test
   0.10% starting the test        
[engine] net_test: running next measurement
   • OnProgress: 0.100000 - Processing input: https://www.iconnecthere.com/
   0.10% Processing input: https://www.iconnecthere.com/
[engine] net_test: creating entry
[engine] net_test: calling setup
[engine] net_test: running with input https://www.iconnecthere.com/
[engine] web_connectivity: starting dns_query for www.iconnecthere.com
[engine] net_test: running next measurement
[engine] net_test: reached end of input
[engine] net_test: running next measurement
[engine] net_test: reached end of input
```

See how Web Connectivity is performed for a single URL (with some snipping):

```
[engine] getaddrinfo('www.iconnecthere.com') => error: code=0, reason=''
[engine] dns_test: got response!
[engine] dns_test: callbacking
[engine] web_connectivity: starting tcp_connect
[engine] resolve_hostname: 163.44.187.153
[engine] resolve_hostname: is valid ipv4
[engine] connect_base 163.44.187.153:443
[engine] connect() in progress...
[engine] dns_test: callback called
[engine] connect time: 0.315201
[engine] web_connectivity: success to connect to 163.44.187.153:443
[engine] web_connectivity: starting http_request to https://www.iconnecthere.com/
[engine] Requesting url https://www.iconnecthere.com/
[engine] resolve_hostname: www.iconnecthere.com
[engine] resolve_hostname: ipv4...
[engine] getaddrinfo('www.iconnecthere.com') => error: code=0, reason=''
[engine] resolve_hostname: ipv4... done
[engine] resolve_hostname: ipv6...
[engine] getaddrinfo('www.iconnecthere.com') => error: code=0, reason=''
[engine] resolve_hostname: ipv6... done
[engine] connect_base 163.44.187.153:443
[engine] connect() in progress...
[engine] connect time: 0.271237
[engine] ssl: creating SSL_CTX with bundle: '/Users/sbasso/.ooni/assets/ca-bundle.pem'
[engine] Re-enabling SSLv2 and SSLv3
[engine] SSL version: TLSv1.2
[engine] ssl: handshake...
[engine] ssl: handshake... complete
[engine] Allowing dirty SSL shutdown
[engine] > GET / HTTP/1.1
[engine] > User-Agent: Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36
[engine] > Accept-Language: en-US;q=0.8,en;q=0.5
[engine] > Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
[engine] > Host: www.iconnecthere.com
[engine] >
[engine] http: started reading response
[engine] http: passing read data to parser
[engine] < HTTP/1.1 200 OK
[engine] < Server: nginx
[engine] < Date: Sun, 16 Jun 2019 07:54:49 GMT
[engine] < Content-Type: text/html; charset=UTF-8
[engine] < Transfer-Encoding: chunked
[engine] < Connection: keep-alive
[engine] < Link: <https://www.iconnecthere.com/wp-json/>; rel="https://api.w.org/", <https://wp.me/P9whHu-Bz>; rel=shortlink
[engine] < X-XSS-Protection: 1; mode=block
[engine] < X-Content-Type-Options: nosniff
[engine] < X-Nginx-Cache: HIT
[engine] <
[engine] http: continue reading for the response
[engine] http: passing read data to parser
...
[engine] http: received error 0 on connection
[engine] web_connectivity: doing control request
[engine] Using backend https://c.web-connectivity.th.ooni.io:443
[engine] resolve_hostname: c.web-connectivity.th.ooni.io
[engine] resolve_hostname: ipv4...
[engine] getaddrinfo('c.web-connectivity.th.ooni.io') => error: code=0, reason=''
[engine] resolve_hostname: ipv4... done
[engine] resolve_hostname: ipv6...
[engine] getaddrinfo('c.web-connectivity.th.ooni.io') => error: code=0, reason=''
[engine] resolve_hostname: ipv6... done
[engine] connect_base 37.218.241.164:443
[engine] connect() in progress...
[engine] connect time: 0.564001
[engine] SSL version: TLSv1.2
[engine] ssl: handshake...
[engine] ssl: handshake... complete
[engine] > POST / HTTP/1.1
[engine] > Content-Type: application/json
[engine] > Host: c.web-connectivity.th.ooni.io
[engine] > Content-Length: 350
[engine] >
[engine] http: started reading response
[engine] http: passing read data to parser
[engine] < HTTP/1.1 200 OK
[engine] < Date: Sun, 16 Jun 2019 07:54:52 GMT
[engine] < Content-Length: 559
[engine] < Content-Type: application/json
[engine] < Server: cyclone/1.1
[engine] <
[engine] http: received error 0 on connection
[engine] web_connectivity: comparing control with experiment
[engine] web_connectivity: comparing http_requests
[engine] web_connectivity: body proportion 1.000000
[engine] web_connectivity: comparing status codes
[engine] web_connectivity: checking headers
[engine] web_connectivity: checking HTML titles
[engine] web_connectivity: comparing dns_queries
[engine] web_connectivity: comparing tcp_connect
[engine] web_connectivity: exp,ctrl http failure determination
[engine] web_connectivity: checking if we got expected web page
[engine] web_connectivity: determining blocking reason
[engine] web_connectivity: no blocking detected
```

See how we're tearing down the nettest and leaving MK:

```
[engine] net_test: tearing down
[engine] net_test: written entry
[engine] net_test: running next measurement
[engine] net_test: reached end of input
   • OnProgress: 0.950000 - ending the test
   0.95% ending the test          
   • OnProgress: 1.000000 - test complete
   1.00% test complete            
```

Now the control's back to ooni/probe-cli and ooni/probe-engine. What we're doing now is submitting the measurement to the collector.

```
[engine] jsonapi: request body: {
    "content": {
        "annotations": {
            "engine_name": "libmeasurement_kit",
            "engine_version": "0.10.4",
            "engine_version_full": "v0.10.4",
            "platform": "macos"
        },
        "data_format_version": "0.2.0",
        "id": "8139bf0d-53b6-400b-8f3a-39a020cfd046",
        "input": "https://www.iconnecthere.com/",
        "measurement_start_time": "2019-06-16 07:54:47",
        "probe_asn": "AS30722",
        "probe_cc": "IT",
        "probe_ip": "127.0.0.1",
        "report_id": "",
        "software_name": "ooniprobe-desktop",
        "software_version": "3.0.0-beta.3",
        "test_helpers": {
            "backend": {
                "address": "https://c.web-connectivity.th.ooni.io:443",
                "type": "https"
            }
        },
        "test_keys": {
            "accessible": true,
            "agent": "redirect",
            "blocking": false,
            "body_length_match": true,
            "body_proportion": 1,
            "client_resolver": "173.194.169.102",
            "control": {
                "dns": {
                    "addrs": [
                        "163.44.187.153"
                    ],
                    "failure": null
                },
                "http_request": {
                    "body_length": 554651,
                    "failure": null,
                    "headers": {
                        "Content-Type": "text/html; charset=UTF-8",
                        "Date": "Sun, 16 Jun 2019 07:54:53 GMT",
                        "Link": "<https://www.iconnecthere.com/wp-json/>; rel=\"https://api.w.org/\", <https://wp.me/P9whHu-Bz>; rel=shortlink",
                        "Server": "nginx",
                        "X-Content-Type-Options": "nosniff",
                        "X-Nginx-Cache": "HIT",
                        "X-XSS-Protection": "1; mode=block"
                    },
                    "status_code": 200,
                    "title": " | "
                },
                "tcp_connect": {
                    "163.44.187.153:443": {
                        "failure": null,
                        "status": true
                    }
                }
            },
            "control_failure": null,
            "dns_consistency": "consistent",
            "dns_experiment_failure": null,
            "headers_match": true,
            "http_experiment_failure": null,
            "queries": [
                {
                    "answers": [
                        {
                            "answer_type": "CNAME",
                            "hostname": "www.iconnecthere.com",
                            "ttl": null
                        },
                        {
                            "answer_type": "A",
                            "ipv4": "163.44.187.153",
                            "ttl": null
                        }
                    ],
                    "engine": "system",
                    "failure": null,
                    "hostname": "www.iconnecthere.com",
                    "query_type": "A",
                    "resolver_hostname": null,
                    "resolver_port": null
                }
            ],
            "requests": [
                {
                    "failure": null,
                    "request": {
                        "body": "",
                        "headers": {
                            "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
                            "Accept-Language": "en-US;q=0.8,en;q=0.5",
                            "User-Agent": "Mozilla/5.0 (Windows NT 6.1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/47.0.2526.106 Safari/537.36"
                        },
                        "method": "GET",
                        "tor": {
                            "exit_ip": null,
                            "exit_name": null,
                            "is_tor": false
                        },
                        "url": "https://www.iconnecthere.com/"
                    },
                    "response": {
                        "body": "...",
                        "code": 200,
                        "headers": {
                            "Connection": "keep-alive",
                            "Content-Type": "text/html; charset=UTF-8",
                            "Date": "Sun, 16 Jun 2019 07:54:49 GMT",
                            "Link": "<https://www.iconnecthere.com/wp-json/>; rel=\"https://api.w.org/\", <https://wp.me/P9whHu-Bz>; rel=shortlink",
                            "Server": "nginx",
                            "Transfer-Encoding": "chunked",
                            "X-Content-Type-Options": "nosniff",
                            "X-Nginx-Cache": "HIT",
                            "X-XSS-Protection": "1; mode=block"
                        },
                        "response_line": "HTTP/1.1 200 OK"
                    }
                }
            ],
            "retries": null,
            "socksproxy": null,
            "status_code_match": true,
            "tcp_connect": [
                {
                    "ip": "163.44.187.153",
                    "port": 443,
                    "status": {
                        "blocked": false,
                        "failure": null,
                        "success": true
                    }
                }
            ],
            "title_match": false
        },
        "test_name": "web_connectivity",
        "test_runtime": 7.021033048629761,
        "test_start_time": "2019-06-16 07:54:47",
        "test_version": "0.0.1"
    },
    "format": "json"
}
[engine] jsonapi: method: POST
[engine] jsonapi: URL: https://c.collector.ooni.io:443/report/20190616T075447Z_AS30722_uYWBiL1DmELYYLTFKUmixgZhNwfCQxHm6Nlhj5Q4Wk8DMYJ3Ir
[engine] http: connection to 37.218.242.210:443 ready; sending request
```

Note that we're again reusing the open connection.

```
[engine] > POST /report/20190616T075447Z_AS30722_uYWBiL1DmELYYLTFKUmixgZhNwfCQxHm6Nlhj5Q4Wk8DMYJ3Ir HTTP/1.1
[engine] > Content-Length: 595830
[engine] > Host: c.collector.ooni.io:443
[engine] > Content-Type: application/json
[engine] > User-Agent: ooniprobe-desktop/3.0.0-beta.3
[engine] > Accept-Encoding: gzip
[engine] >
[engine] http: request sent; waiting for response
[engine] http: start receiving response
[engine] < HTTP/1.1 200 OK
[engine] < Date: Sun, 16 Jun 2019 07:54:55 GMT
[engine] < Content-Type: application/json
[engine] < Content-Length: 21
[engine] < Connection: keep-alive
[engine] < Keep-Alive: timeout=120
[engine] < Server: nginx/1.6.2
[engine] <
[engine] jsonapi: response body: {"status": "success"}
```

And this is it for the first URL, now we're going over the second URL:

```
   • Fetching: 0 &{
    32 web_connectivity 2019-06-16 07:54:47.781657 +0000 UTC
    7.814443 true true false { false} false { false} false {... true} {198 true} {0 false} {
    false false}  13 /.../ooni283065313/msmt-websites.WebConnectivity-2019-06-16T075447.134412Z.jsonl}
   • status.measurement_start 
[engine] measurementkit: settings: {
    "inputs":["http://adultfriendfinder.com/"],
    "log_level":"DEBUG",
    "name":"WebConnectivity",
    "options":{
...
}}
```

So, it should be clear now that we're not performing a geolookup for every URL and we're not opening a report for every URL. Rather, these functionality have been moved from the C++ engine to the Go engine. This is part of the plan to move as much functionality as possible to the Go engine, to make the C++ engine unnecessary in the future.

Do you agree with this analysis? If so, I believe the correct follow-up is to address all the issues that I've found while reading the logs before merging.

bassosimone:
(It seems I broke Travis in my attempts to fix it for probe-engine.)

bassosimone:
> (It seems I broke Travis in my attempts to fix it for probe-engine.)

For reference, to unbreak it and get back the webook, go to `travis-ci.org/organizations/ooni/repositories`, disable and re-enable the affected repo.

bassosimone:
It also seems I did broke the Linux build. Let me check this.

bassosimone:
Bug filed: https://github.com/measurement-kit/script-build-linux/issues/6

nettests/websites/web_connectivity.go
hellais: Yes what you say makes sense and apologies for jumping to conclusions. I took note of this last night mostly so I wouldn't forget to look into with more calm.

Thank you for your very detailed and thorough analysis which seems very reasonable.

What tripped me off was that I was seeing in the logs:
```
[engine] Your public IP address: 188.152.18.76
[engine] Your country: IT
[engine] Your ISP identifier: AS30722
[engine] Your ISP name: Vodafone Italia S.p.A.
   • OnProgress: 0.050000 - geoip lookup
   7.05% geoip lookup
```

Which made me believe that a geoip lookup was being done each time (this log line did not appear in previous versions). Perhaps we should remove the line saying "geoip lookup" from the logs in this case.

Similarly for report creation I was tipped off by the log line saying:
```
   6.10% open report
```

on each URL.

Perhaps we can cleanup the output a bit to show these messages only when the report is opened for the first time and the geoip lookup operation is being performed?

bassosimone: Indeed! In fact, I removed these logs from MK v0.10.4 precisely because I thought they were rather confusing :-).

bassosimone:
Fixed in these PRs, which require some light review:

- https://github.com/measurement-kit/script-build-linux/pull/7

- https://github.com/ooni/probe-engine/pull/27

When those are merged, we can resume the work here.

nettests/nettests.go
bassosimone: ping @hellais

bassosimone:
# TODO list of actions to be performed before merging

@hellais here's the initial list I identified based on our exchanges above. Please add to this list if you think there's something we should address here or by opening new issues.

- [ ] add and use `DefaultProbeASNString` in the engine (https://github.com/ooni/probe-cli/pull/46#discussion_r292825808)

- [ ] lookup the DNS address in the engine and pass it down to MK (https://github.com/ooni/probe-cli/pull/46#discussion_r294074083, https://github.com/measurement-kit/measurement-kit/issues/1863)

- [ ] make sure IP scrubbing is performed somewhere (it's currently unclear whether it is performed and I'm actually keep on doing it a higher rather than on a lower level)

nettests/nettests.go
hellais: I guess you should use the type which is most comfortable. I would advise against defining another representation for it called `model.DefaultProbeASNString` as that would also be inconsistent with the naming of other variables.

If you are fine with doing the `fmt.Sprintf` calls then go for it. It was just something that made me think enough to write a comment while reviewing it.

I am fine with whatever solution you go for (even just leaving it as-is).

bassosimone: ACK, thank you!

nettests/websites/web_connectivity.go
bassosimone: @hellais:

> Edit: I should also add that I noticed that the runtime of the run websites command on this branch appeared to be slower than on the current master.

The first explanation that comes to mind is that MK uses parallelism `3` whereas the code I've added in here is strictly sequential. As you know, we've discussed a number of times what to do in terms of parallelism, and it seems on Monday with @FedericoCeratto and you we concluded that we're fine with parallelism `1` by default. Hence, this initial iteration of the code has no provision for parallelism and all is sequential.

bassosimone: I have made sure we have actionable issues for every aspect mentioned in this branch of the conversation, so I am going to resolve it, to simplify my understanding of progress.

bassosimone:
With the merging of https://github.com/ooni/probe-engine/pull/37 we have enough pieces in place to move forward this PR. While the `resolver_ip` is not properly handled by the probe engine, it will be handled at a later stage. Also, because we're still on MK v0.10.4 with the engine, that does not matter, _because_ MK writes the `resolver_ip` anyway. The main issue is that it performs a resolver lookup for each input, and that will be solved with MK v0.10.5 and a subsequent engine.

@hellais can you please take a final look to this branch?

hellais:
This branch looks good to me, however I think that the `resolver_ip` is still a bit of a blocker on cutting a new release on probe-cli.

For me it's good to go ahead with merging this PR, but it would be good to prioritise the `resolver_ip` lookup and writing functionality as it's pretty crucial.

hellais:
Adding some notes from the discussion we had privately with @bassosimone.

It appears that adding the `client_resolver` key into the `test_keys` is not so seamless to do in the client. This friction highlights how this is most likely an issue in the specification and the most logical place to put it would in theory be in the top level keys.

However doing that introduces problems in data processing and normalisation which we aren't really sure how to handle.

In the best interest of moving forward I suggest that for the time being we add the `client_resolver` to the top level keys so that at least it's there and then we figure out what is the best strategy going forward for this prior to a stable launch/release.

PR #53, opened 2019-08-15
Implement the show command

hellais:
The show commands allows you to view the JSON of a measurement from the CLI.


bassosimone:
@hellais CI is failing and it seems it has been running for over 50 min. Can you take a look? Could it be caused by the recent changes in the test-lists API?

hellais:
Yes you were right, it was failing due to it running over the full test list. I am adding support for passing in a limit.

hellais:
It seems like the linux build is now failing due to:
```
g++: error: /usr/lib/libssl.a: No such file or directory
g++: error: /usr/lib/libcrypto.a: No such file or directory
g++: error: /lib/libz.a: No such file or directory
```

I checked the image for `openobservatory/mk-alpine` (https://github.com/measurement-kit/script-build-linux) and I do not find static versions of those libraries inside of it.

These are referenced in here: https://github.com/ooni/probe-engine/blob/master/measurementkit/task_cgo.go#L30

hellais:
Should now be fixed. Let's wait for the CI to be happy.

.travis.yml
```diff
@@ -12,6 +12,10 @@ matrix:
     env:
     - OS_NAME: macos
     osx_image: xcode10.2
+    cache:
+    directories:
+      - $HOME/Library/Caches/Homebrew
+      - /usr/local/Homebrew
```
bassosimone: ah, well done, this is a trick that we can use also in other places. (is it intended to have `cache:` and `directories:` at the same level of nesting?)

hellais: Ah you are right, we need to nest it one extra level.

I also didn't fully understand if what I am doing here is going to be entirely effective. This SO answer https://stackoverflow.com/questions/39930171/cache-brew-builds-with-travis-ci makes this thing look much more complicated than I think it should be.

Looking at that above post I concluded these are most likely the main paths we want to cache.

.travis.yml
```diff
@@ -12,6 +12,10 @@ matrix:
     env:
     - OS_NAME: macos
     osx_image: xcode10.2
+    cache:
+      directories:
+        - $HOME/Library/Caches/Homebrew
+        - /usr/local/Homebrew
```
bassosimone: @hellais today [the build broke](https://travis-ci.org/ooni/probe-cli/jobs/582595891#L2446). Could this be related to attempts to create an homebrew cache? Can we confirm it works without cache?

hellais: Yeah that would be a good thing to try out or as you were suggesting on #ooni-dev dropping macOS CI support entirely.

@bassosimone do you think you can pick this branch up and get it in a working state?

bassosimone: Apart from dropping the macOS CI, anything else is required?

bassosimone: I've filed #54 to address this specific issue in the PR.

bassosimone:
CI is now green, starting the code review soon

config/settings.go
```diff
@@ -42,6 +42,7 @@ type NettestConfig struct {
 // Websites test group
 type Websites struct {
 	EnabledCategories []string `json:"enabled_categories"`
+	Limit			  int      `json:"limit"`
```
bassosimone: Did you run `go fmt`?

go.sum
```diff
@@ -141,12 +141,16 @@ github.com/onsi/ginkgo v1.8.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+W
...
+github.com/ooni/probe-engine v0.0.0-20190828175829-9e908f8e3a3a h1:e3SiEPrbtsB76tPm+sFLI5dtKBd7sa+vjtdfvSMKLqk=
+github.com/ooni/probe-engine v0.0.0-20190828175829-9e908f8e3a3a/go.mod h1:zovuewNdJp8exKnBlpCgZXNu/SN+Cs4acj2KgGZOP9s=
 github.com/oschwald/geoip2-golang v1.3.0 h1:D+Hsdos1NARPbzZ2aInUHZL+dApIzo8E0ErJVsWcku8=
 github.com/oschwald/geoip2-golang v1.3.0/go.mod h1:0LTTzix/Ao1uMvOhAV4iLU0Lz7eCrP94qZWBTDKf0iE=
 github.com/oschwald/maxminddb-golang v1.3.0 h1:oTh8IBSj10S5JNlUDg5WjJ1QdBMdeaZIkPEVfESSWgE=
 github.com/oschwald/maxminddb-golang v1.3.0/go.mod h1:3jhIUymTJ5VREKyIhWm66LJiQt04F0UCDdodShpjWsY=
 github.com/oschwald/maxminddb-golang v1.3.1 h1:kPc5+ieL5CC/Zn0IaXJPxDFlUxKTQEU8QBTtmfQDAIo=
 github.com/oschwald/maxminddb-golang v1.3.1/go.mod h1:3jhIUymTJ5VREKyIhWm66LJiQt04F0UCDdodShpjWsY=
+github.com/oschwald/maxminddb-golang v1.4.0 h1:5/rpmW41qrgSed4wK32rdznbkTSXHcraY2LOMJX4DMc=
+github.com/oschwald/maxminddb-golang v1.4.0/go.mod h1:3jhIUymTJ5VREKyIhWm66LJiQt04F0UCDdodShpjWsY=
```
bassosimone: Can you `go mod tidy`? It seems surprising that `maxminddb-golang` is removed from `go.mod` and readded here?

internal/cli/run/run.go
```diff
@@ -50,20 +48,20 @@ func init() {
...
-		fmt.Sprintf("the nettest group to run. Supported tests are: %s, or nothing to run them all",
-			strings.Join(nettestGroupNamesBlue, ", "))).String()
-
 	noCollector := cmd.Flag("no-collector", "Disable uploading measurements to a collector").Bool()
 	collectorURL := cmd.Flag("collector-url", "Specify the address of a custom collector").String()
 	bouncerURL := cmd.Flag("bouncer-url", "Specify the address of a custom bouncer").String()
 
-	cmd.Action(func(_ *kingpin.ParseContext) error {
-		ctx, err := root.Init()
+	cmd.Action(func(kp *kingpin.ParseContext) error {
```
bassosimone: It seems `kp` is still unused?

internal/cli/run/run.go
```diff
@@ -112,17 +110,34 @@ func init() {
...
-				if err := runNettestGroup(tg, ctx, network); err != nil {
-					log.WithError(err).Errorf("failed to run %s", tg)
-				}
+	websitesCmd := cmd.Command("websites", "")
+	websitesCmd.Action(func(_ *kingpin.ParseContext) error {
+		return runNettestGroup("websites", ctx, network)
+	})
+	imCmd := cmd.Command("im", "")
+	imCmd.Action(func(_ *kingpin.ParseContext) error {
+		log.Info("calling im code path")
```
bassosimone: This seems a debug leftover?

internal/cli/run/run.go
```diff
@@ -112,17 +110,34 @@ func init() {
...
+	middleboxCmd := cmd.Command("middlebox", "")
+	middleboxCmd.Action(func(_ *kingpin.ParseContext) error {
+		return runNettestGroup("middlebox", ctx, network)
+	})
+	allCmd := cmd.Command("all", "").Default()
+	allCmd.Action(func(_ *kingpin.ParseContext) error {
+		log.Infof("Running %s tests", color.BlueString("all"))
+		for tg := range groups.NettestGroups {
+			if err := runNettestGroup(tg, ctx, network); err != nil {
+				log.WithError(err).Errorf("failed to run %s", tg)
```
bassosimone: Do we want this to happen? Would we rather want the probe to continue and then eventually `exit(1)`. This seems to me a problem similar to `make` vs `make -k`. I'd say we discuss this now, then this code is fine anyway, and if we decide that the default is to continue, we file an issue for later. (I'd rather not introduce a `-k`, but maybe you disagree?)

internal/cli/run/run.go
```diff
@@ -50,20 +48,20 @@ func init() {
...
-
 	noCollector := cmd.Flag("no-collector", "Disable uploading measurements to a collector").Bool()
 	collectorURL := cmd.Flag("collector-url", "Specify the address of a custom collector").String()
 	bouncerURL := cmd.Flag("bouncer-url", "Specify the address of a custom bouncer").String()
 
-	cmd.Action(func(_ *kingpin.ParseContext) error {
-		ctx, err := root.Init()
+	cmd.Action(func(kp *kingpin.ParseContext) error {
+		var err error
+		ctx, err = root.Init()
```
bassosimone: I have a more scratching-my-head-like question: why is `ctx` initialized here and used in later commands?

internal/cli/show/show.go
```diff
@@ -4,19 +4,26 @@ import (
...
+		ctx, err := root.Init()
 		if err != nil {
 			log.WithError(err).Error("failed to initialize root context")
 			return err
 		}
-		log.Error("this function is not implemented")
-
+		msmt, err := database.GetMeasurementJSON(ctx.DB, *msmtID)
+		if err != nil {
+			log.Errorf("error: %v", err)
```
bassosimone: We probably want to `return err` here

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+
+func GetMeasurementJSON(sess sqlbuilder.Database, measurementID int64) (map[string]interface{}, error) {
+	var (
+		measurement MeasurementURLNetwork
+		msmtJSON map[string]interface{}
+	)
+
+	req := sess.Select(
+		db.Raw("urls.*"),
+		db.Raw("measurements.*"),
```
bassosimone: What does this mean? Select any field from these two tables?

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+	).From("measurements").
+		LeftJoin("urls").On("urls.url_id = measurements.url_id").
+		Where("measurements.measurement_id= ?", measurementID)
+
+	if err := req.One(&measurement); err != nil {
+		log.Errorf("failed to run query %s: %v", req.String(), err)
+		return nil, err
+	}
+	reportFilePath := measurement.Measurement.ReportFilePath
+	if (measurement.URL.URL.Valid == false) {
```
bassosimone: You should add a comment explaining why you're performing this check. (I assume this has to do with the fact that the measurement may or may not already be in explorer.)

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+		b, err := ioutil.ReadFile(reportFilePath)
+		if err != nil {
+			return nil, err
+		}
+		json.Unmarshal(b, &msmtJSON)
```
bassosimone: Please, check the return value and return it. It may never fail but a little bit of robustness here costs nothing (and in particular it does not even require us adding extra lines of code.)

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+		return msmtJSON, nil
+	}
+
+	url := measurement.URL.URL.String
```
bassosimone: Oh, I am also confused by `.URL.URL`: why twice?

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+	file, err := os.Open(reportFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	log.Infof("URL is %s", url)
```
bassosimone: This is also probably a debugging leftover. Can we either have a more informative message or silence? (or Debug?)

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+	reader := bufio.NewReader(file)
+
+	for {
+		line, err := util.ReadLine(reader)
+		if (err == io.EOF) {
+			break
+		}
```
bassosimone: Please, make sure we return an error if `err != nil`

internal/database/actions.go
```diff
@@ -48,20 +106,18 @@ func GetResultTestKeys(sess sqlbuilder.Database, resultID int64) (string, error)
...
 		// it could be that these keys come out as initializes with the default
 		// values.
 		// XXX we may want to change this behaviour by adding `omitempty` to the
 		// struct definition.
-		if msmt.TestName == "ndt" || msmt.TestName == "dash" {
-			if err := json.Unmarshal([]byte(msmt.TestKeys), &tk); err != nil {
-				log.WithError(err).Error("failed to parse testKeys")
-				return "{}", err
-			}
+		if (msmt.TestName != "ndt" && msmt.TestName != "dash") {
```
bassosimone: please remove the unnecessary parentheses

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
...
+		}
+		json.Unmarshal([]byte(line), &msmtJSON)
+		if (msmtJSON["input"].(string) == url) {
+			return msmtJSON, nil
+		}
+	}
+	return nil, errors.New("Could not find measurement")
+}
+
 // GetResultTestKeys returns a list of TestKeys for a given measurements
```
bassosimone: Please fix the grammar in the documentation comment

internal/database/actions.go
```diff
@@ -48,20 +106,18 @@ func GetResultTestKeys(sess sqlbuilder.Database, resultID int64) (string, error)
 		tk   PerformanceTestKeys
 	)
 	for res.Next(&msmt) {
-		if msmt.TestName == "web_connectivity" {
-			break
-		}
```
bassosimone: Historically, why was this code here?

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
 	return measurements, nil
 }
 
+type MeasurementWithInput struct {
+	Input string `json:"input"`
+}
+
+func GetMeasurementJSON(sess sqlbuilder.Database, measurementID int64) (map[string]interface{}, error) {
```
bassosimone: Please add documentation

internal/database/actions.go
```diff
@@ -38,6 +42,60 @@ func ListMeasurements(sess sqlbuilder.Database, resultID int64) ([]MeasurementUR
 	return measurements, nil
 }
 
+type MeasurementWithInput struct {
+	Input string `json:"input"`
+}
```
bassosimone: Where is this struct used? I cannot immediately see it

internal/output/output.go
```diff
@@ -11,6 +11,14 @@ import (
 	"github.com/ooni/probe-cli/internal/util"
 )
 
+// MeasurementJSON prints the JSON of a measurement
+func MeasurementJSON(j map[string]interface{}) {
```
bassosimone: I was about to say "hey, this is not a good name", then I read the package name!

internal/util/util.go
```diff
@@ -112,3 +113,18 @@ func WrapString(s string, lim uint) string {
...
 }
+
+
+// ReadLine will read a single line from a bufio.Reader
+func ReadLine(r *bufio.Reader) (string, error) {
+	var (
+		isPrefix bool
+		err error
+		line, ln []byte
+	)
```
bassosimone: I think you can write more compact code using `:=` below, or I'm missing why you're not using it.

internal/util/util.go
```diff
@@ -112,3 +113,18 @@ func WrapString(s string, lim uint) string {
...
+	for isPrefix && err == nil {
+		line, isPrefix, err = r.ReadLine()
+		ln = append(ln, line...)
```
bassosimone: Why are you using `line...` here? What is the effect?

config/settings.go
hellais: No. I need to check why my editor is not doing this on save anymore.

go.sum
hellais: Maybe because it's a different version?

internal/cli/run/run.go
hellais: This is the parent action, so it's going to be called anytime you call the child actions. The initialisation of ctx is done in here so that we don't have to duplicate the code for every sub-action.

internal/cli/run/run.go
hellais: That is true, I suppose it can be underscored.

internal/cli/run/run.go
hellais: I am not sure I understand entirely. Do you mean why don't we instead exit when the first test of a group fails?

I don't think that would be desirable as we still wait to run the remaining tests inside of the group even if a single instance of it is to fail.

Maybe it would be wise to in some way record this and return a different exit code when the process terminates later on to indicate the failure, but I don't think we should be exiting early here.

internal/database/actions.go
hellais: Select ~any~**all** fields from these two tables

internal/database/actions.go
hellais: The `Valid` method is part of the standard `Nullable` `sql` objects. It's used to be able to make a query on a column where the value may be set to a particular value or to `NULL`.

This is checking if the measurement in question has a URL input. If there is no input then we don't need to seek to the right line offset because it's a single entry report file and we can just read it and return it.

internal/database/actions.go
hellais: It is twice because `URL`[0] is the `url` table, while `URL`[1] is the `url` column inside of the `url` table.

internal/database/actions.go
hellais: I was under the impression I had done this to be equivalent, but I actually see there may be something different happening. I will take a look at this logic again when I go back to my editor.

internal/util/util.go
hellais: Because `line` is a byte array and I unpack it before appending it to the `ln` byte array.

internal/util/util.go
hellais: Not sure what you are suggesting. Can you provide a diff?

internal/database/actions.go
hellais: So I checked and actually given the comment I see, this is the correct behaviour. I suppose that if we start enriching that testKeys at this level for clients we may want to go back and change this, but for the moment I say it stays like this.

Note: the if statement below is more stringent than it was before.

internal/database/actions.go
hellais: Only here you don't like them eh? :P

internal/cli/run/run.go
bassosimone: Ah, reading more carefully, `Errorf` does not exit. This is what I'd like to happen.

hellais: 😕 https://godoc.org/github.com/apex/log#Entry.Errorf

internal/cli/run/run.go
bassosimone: While it is true that we don't have to duplicate the code, a much less confusing implementation would be to factor the initialisation code in a function that returns whatever we need, and call it. The current implementation is magic and, because of this reason, I don't like that [*]. I would be happier if we could have more plain code here. Can you please refactor the initialisation code and have it return what you need in every context?

[*] This is, in a way, similar to using a global variable, except that a programmer does not expect this to happen because they are usually more accustomed to global variables being dangerous, not to stack variables. My issue with this usage of a stack variable is that it confuses my brain when I reason about scoping.

internal/cli/show/show.go
bassosimone: (As a reminded, can you let me resolve conversations when I think I am satisfied with the fix, which is actually the case here? It makes my review process more streamlined and pleasant.)

internal/database/actions.go
bassosimone: Thanks, I was in doubt!

internal/database/actions.go
bassosimone: > This is checking if the measurement in question has a URL input. If there is no input then we don't need to seek to the right line offset because it's a single entry report file and we can just read it and return it.

Yes, this is the part that needs to be documented. It is a bit of information that otherwise I'd be missing.

internal/database/actions.go
bassosimone: In general, I find it bad to have a `.Foo.Foo` chaining where it's unclear to me what `Foo` and `Foo` are. At the same time, I believe the naming cannot change at this point, and changing it would be non-practical :-(.

internal/database/actions.go
bassosimone: Gotcha!

internal/util/util.go
bassosimone: No, I can't, it seems the code cannot be simplified further.

internal/util/util.go
bassosimone: So, here you're basically trying the deal with the fact that we have very long lines, correct? Can you perhaps document this as the reason why we need to wrap ReadLine? (Otherwise a naive person reading the code could wonder why we're wrapping ReadLine)

hellais: I suppose, I can do that though the docs of `bufio.Reader.ReadLine` pretty clearly say:

> ReadLine tries to return a single line, not including the end-of-line bytes. If the line was too long for the buffer then isPrefix is set and the beginning of the line is returned. 

How about I rename the function to `ReadLongLine`?

internal/database/actions.go
hellais: I added docs already: https://github.com/ooni/probe-cli/pull/53/commits/e0e12cdbca06aa5b5193467f51077af193d712c9#diff-33dda6bf2d32805bdafa67dc4c47250eR65

internal/database/actions.go
hellais: If you look at the `MeasurementURLNetwork` model inside of `internal/database/models.go` I think it will improve your understanding of what `URL` and `URL` are (one is the inline definition of the table, while the other is the attribute mapped to the column in the table).

I don't think it's that uncommon to have this sort of database schema where you have a column in the table named like the table itself.

internal/database/actions.go
bassosimone: 👍

internal/cli/run/run.go
hellais: This was also done in the interest of making changes part of this PR minimal. Had I done what you suggest I probably would have had to make more substantial changes in how subcommands are handled.

I agree that this can maybe be improved. Would you like to maybe document this as an issue and work on refactoring this code?

internal/database/actions.go
bassosimone: I mean, if you use goreportcard.com and include parentheses in this context, it will reduce your score and suggest to use `gofmt -s` to simplify the codebase.

internal/util/util.go
bassosimone: Yes, that part of the docs is indeed why I guessed we were in such case. I am not necessarily confused by the function name, but I'd like to see the _reason_ why we expect long lines to be documented somewhere.

internal/cli/show/show.go
hellais: Hum, it's that then it's harder for me to track what I did fix or didn't. I suppose I can write a comment saying that it's fixed.

I generally leave them unresolved if I believe there is value in having the reviewer look at the fix to validate if it's OK. In this case the fix was very trivial and I did not consider it necessary for you to look at it.

internal/util/util.go
hellais: > I'd like to see the reason why we expect long lines to be documented somewhere

I mean we should be prepared for anything. Tbh I find it a major oversight from the golang stdlib developers to make a functions that is called `ReadLine` and doesn't actually do what it's suppose to do.

How about I add a docstring to the docstring: "Just like the golang ReadLine, except it actually does what it should do.".

I would actually argue that the golang stdlib function should be called `ReadLineOrAPartOfIt` :P

internal/cli/run/run.go
bassosimone: This change is adding technical debt. Can we avoid that now rather than postponing?

internal/database/actions.go
bassosimone: Sure, if I knew better the code I'd not be surprised. But that the code is structured in a way that it is a bit surprising is a bummer. I guess this is a side effect of using an ORM. Also, it may be that, when using a IDE, it's easier to know what is what. Instead, when doing a review only looking at a diff, this looks much worst. (But maybe this is a feature).

In any case, I don't think there's much we can do about that 🤷‍♂

bassosimone: > I don't think it's that uncommon to have this sort of database schema where you have a column in the table named like the table itself.

I have seen `URLS.URL` and `URL.URL`. There are arguments for both designs. We have a design with `URL.URL` and we probably don't want to change it.

internal/cli/run/run.go
hellais: Sure, feel free to push code to this branch as well.

internal/util/util.go
bassosimone: It seems to me we're using a deprecated interface [1]. Have you perhaps checked whether we could use bufio.Scanner [2] or bufio.Reader.ReadString [3] instead? (Both solutions seem to overcome the issue that ReadLine is bounded to a fixed size slice.)

[1] https://twitter.com/davecheney/status/604837853344989184?lang=en

[2] https://golang.org/pkg/bufio/#Scanner

[3] https://golang.org/pkg/bufio/#Reader.ReadString

internal/cli/run/run.go
hellais: I migrated this to an issue: https://github.com/ooni/probe-cli/issues/55

hellais: I have migrated this to issue #55

internal/util/util.go
bassosimone: https://github.com/ooni/probe-cli/issues/57

PR #60, opened 2019-10-02
Upgrade to ooni/probe-engine@v0.1.1

bassosimone:
This PR upgrades to ooni/probe-engine@v0.1.0. As part of that release, I've started hiding some previously exposed fields, to favour APIs and interfaces. My goal is to have a coarse coupling between this codebase and probe-engine, where possible, as explained in https://github.com/ooni/probe-engine/issues/75. So, rather than per-using fields, this diff goes in the direction of using interfaces internally, and abstracting our needs. Now we don't depend anymore on the definition of probe-engine's `model.LocationInfo`. The same information could be accessed from a `Session` instance, for which we have a convenient internal abstraction for testability.


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/lib/pq v1.1.1 // indirect
 	github.com/mattn/go-colorable v0.1.2
 	github.com/mattn/go-sqlite3 v1.10.0 // indirect
-	github.com/ooni/probe-engine v0.0.0-20190828175829-9e908f8e3a3a
+	github.com/ooni/probe-engine v0.1.1
```
bassosimone: This is the result of `go get github.com/ooni/probe-engine@v0.1.1`, i.e. the command to pin a version.

go.sum
```diff
@@ -1,4 +1,5 @@
 github.com/AndreasBriese/bbloom v0.0.0-20190306092124-e2d15f34fcf9/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
```
bassosimone: This is the "lockfile" of Go. It is machine generated.

internal/cli/run/run.go
```diff
@@ -89,7 +89,7 @@ func init() {
 			log.WithError(err).Error("Failed to lookup the location of the probe")
 			return err
 		}
-		network, err = database.CreateNetwork(ctx.DB, ctx.Session.Location)
+		network, err = database.CreateNetwork(ctx.DB, ctx.Session)
```
bassosimone: We don't expose `Location` anymore since v0.1.0. It may be `null` and should be a private field. Plus, we can obtain the same info in a safer way by using the `ctx.Session` API.

internal/database/actions.go
```diff
@@ -256,14 +255,14 @@ func CreateResult(sess sqlbuilder.Database, homePath string, testGroupName strin
 }
 
 // CreateNetwork will create a new network in the network table
-func CreateNetwork(sess sqlbuilder.Database, location *model.LocationInfo) (*Network, error) {
+func CreateNetwork(sess sqlbuilder.Database, loc util.LocationProvider) (*Network, error) {
```
bassosimone: This is an interface rather than `*session.Session` to simplify mocking and testing.

internal/database/actions_test.go
```diff
@@ -3,14 +3,46 @@ package database
...
 	db "upper.io/db.v3"
 )
 
+type locationInfo struct {
+	asn         uint
+	countryCode string
+	ip          string
+	networkName string
+	resolverIP  string
+}
```
bassosimone: This implements the `util.LocationProvider` interface and allows to write mocked tests.

internal/util/util.go
```diff
@@ -128,3 +128,14 @@ func ReadLine(r *bufio.Reader) (string, error) {
...
+// LocationProvider is an interface that returns the current location. The
+// github.com/ooni/probe-engine/session.Session implements it.
+type LocationProvider interface {
+	ProbeASN() uint
+	ProbeASNString() string
+	ProbeCC() string
+	ProbeIP() string
+	ProbeNetworkName() string
+	ResolverIP() string
+}
```
bassosimone: This is the interface of a thing that tells you about its location. (We basically have replaced a structure exported by another repo with a private interface.) Note that, while it may be tempting to say "export the interface from the probe-engine package", that would not be idiomatic Go. Usually structures are used an interfaces are only used for mocking, internally, or when they are really needed (see `net.Conn`).

bassosimone: Xref: https://twitter.com/davecheney/status/942593128355192832

bassosimone: And https://github.com/golang/go/wiki/CodeReviewComments#interfaces

internal/database/actions.go
sarathms: Despite this comment, I had a tough time understanding that this is indeed a reference to `probe-engine:session.Session`. But that was my poor knowledge of how things work in Go. It does make sense now. Thanks for walking me through it.

bassosimone:
> I have left a question about why you chose to declare the interface where you did.

Good point. I'll see if I can find a better place for it tomorrow!

> The PR title says probe-engine@v0.1.0

Fixed, thank you!

PR #62, opened 2019-10-03
Publish code used to generate v3.0.0-beta.4

bassosimone:


bassosimone:
Will self merge when travis completed because @hellais agreed with this no-fancy-stuff-for-now principle on Slack and suggested to remove goreleaser for now.

PR #63, opened 2019-10-03
Write DESIGN document for OONI Probe CLI v3.0.0

### LINKED ISSUE #56: Write DESIGN document

bassosimone:
This specifies what we exactly expect from the CLI, from the OONI Desktop PoV.

Closes #56


sarathms:
@bassosimone For `probe-desktop` to be able to display NDT measurements properly, it would be optimal to expose the `server_address` field in `test_keys`. Right now, it shows all the other 9 properties we need to show the same info shown by the mobile app.

I know that I can get this information from the output of `ooniprobe show <msmtID>`. But it doesn't make sense to change the pattern I use in `probe-desktop` for all the tests just to fill in this one gap in the `ooni list` command.

bassosimone:
@sarathms 

> Right now, it shows all the other 9 properties we need to show the same info shown by the mobile app.

This a good reminder: ndt7 won't have most fields. Maybe better to start zapping the fields we'll remove in a month or so right now?

sarathms:
The rule of thumb is that we keep the desktop and mobile features as close as possible, at least right now. So let's do such ndt7 changes of they are soon going to be done in `probe-mobile` as well. If those potentially deprecated fields doesn't include the server location, then lets make it work on desktop first and roll out ndt7 when we do.

bassosimone:
@sarathms:

> The rule of thumb is that we keep the desktop and mobile features as close as possible, at least right now. So let's do such ndt7 changes if they are soon going to be done in probe-mobile as well. 

It's even more complex than I remember it. New ndt5 (=== NDT) servers already have dropped these fields so on mobile we show a bunch of zeroes.

See https://github.com/ooni/probe/issues/878

> If those potentially deprecated fields doesn't include the server location, then lets make it work on desktop first and roll out ndt7 when we do.

We've not deprecated server location!

Thanks!

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+
+1. discover the user IP, country, and network
+
+2. list the results stored on disk
+
+3. delete one or more results
+
+4. run any group of experiments (websites, instant messaging,
+performance, middleboxes), or run them all
+
```
sarathms: We should also add `run a failed experiment` here 
ref: ooni/probe-desktop#13

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+```
+
+where:
+
+- `anomaly_count` and `total_count` are the the total number of measurements
+emitted and the number of these that were anomalies.
+
+- `type` identifies this as a measurement summary message.
+
+- all other fields have already-describes semantics, which in this case of
```
sarathms:
```suggestion
- all other fields have already-described semantics, which in this case of
```

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+}
+```
+
+where `"upload"` is the upload speed in kbit/s, `"download"` is the
+download speed in kbit/s, and `"ping"` is the minimum RTT in ms.
+
+For the experiments in the `"middlebox"` group:
+
+```JSON
+{}
```
sarathms: Is this unintended?

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+The OONI CLI will be updated by the desktop app.
+
+The OONI CLI will automatically download and store into
+a configuration/state directory any resource that it
+may require to perform its operations. This directory
+is by default `$HOME/.ooni` on all systems; yet, you can
+override it using the `OONI_HOME` environment variable.
+In the following we will call this directory the OONI home,
+state/configuration directory, or, with a stretch of
+semantics, `OONI_HOME`.
```
hellais: I would add here that the CLI application needs to take into account the fact that a OONI Probe legacy installation MAY exist at this path and it needs to handle this case (it's already handled with a prompt that asks the user to either delete it or make a backup of it).

It's also worth noting that in the desktop installation this is set to another path specific for the desktop app, namely `~/Library/Application\ Support/OONI\ Probe/ooni_home/` on macOS and `XXX find what it is on windows` on windows

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+measurements are saved is below the `OONI_HOME`.
+
+The OONI CLI is a binary called `ooniprobe[.exe]`.
+
+## Batch command line interface
+
+This section describes the batch command line interface, i.e., the
+one that should be used by the desktop app, or by any other tool, to
+drive the OONI CLI. These tools SHOULD set the `OONI_HOME`
+environment variable to avoid clashing with ordinary CLI usage.
```
hellais: Maybe what I said above can be referenced in here

DESIGN.md
hellais: I believe that since we only have to display a boolean of "middlebox detected" or "no middlebox detected" we just need the `is_anomaly` field

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+### Getting a measurement JSON
+
+This command:
+
+```
+ooniprobe --batch show <id>
+```
+
+emits in output the measurement identified by `<id>`. The data
+format is described at https://github.com/ooni/spec.
```
hellais: It would be cool if this was done by fetching it from the API like we do on mobile, though this is maybe something post-MVP.

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+}
+```
+
+where:
+
+- `sharing` configures what information to include in measurements
+and whether to automatically upload measurements
+
+- `test_settings.websites` allows to configure the enabled
+categories and how many URLs test at a time (`0` means no limit)
```
hellais: Maybe you want to mention what the `enabled_categories` field does?

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+    },
+    "middlebox": {
+      "enabled_tests": [
+        "http-invalid-request-line",
+        "http-header-field-manipulation"
+      ]
+    }
+  }
+}
+```
```
hellais: Some settings are missing here, namely:
```
    "send_crash_reports": true,
    "collector_url": "",
    "bouncer_url": "https://bouncer.ooni.io"
```

DESIGN.md
```diff
@@ -0,0 +1,818 @@
...
+reason why we recommend using another `OONI_HOME` when running the
+CLI from another program/app. We do this when running the CLI from
+OONI Desktop, and you should do the same in your use case.
+
+## Configuration file
+
+The configuration file has the following structure:
+
+```JSON
+{
```
hellais: Some fields are missing here, namely:

```
  "_version": 0,
  "_informed_consent": true,
  "_is_beta": true,
  "auto_update": true,
```

hellais:
I took a look at the document and it looks good to me.

I do think it's a bit incomplete in that several functional aspects are missing. I am not sure it's 100% useful to work on ensuring this is fully up to date as I fear it's going to drift out of sync pretty quickly.

Off the top of my heads things that are missing are:
* That data usage is being tracked and recorded to the filesystem
* How logging is handled and written to disk
* The database schema and logic for accessing and updating it
* How database migrations should be handled
* The logic for download and update of geoip data files
* How crash reporting is handled 

I am not suggesting we update the document to include these, but I just thought I would bring these up as I think they are relevant to the implementation.
The reason why I don't recommend we include them in the document is that: 
a. the effort required to specify these in a detailed way is pretty high
b. these implementation details are likely going to change as we move forward with the app development and I think it's going to get tedious pretty fast to have to update code + docs for these low level things

May I suggest that maybe we have this design document focus on the `--batch` JSON API for exposing events and data to the probe-desktop app and keep the docs and discussions related to the above in github issues or other less formal documents?

DESIGN.md
bassosimone: > I would add here that the CLI application needs to take into account the fact that a OONI Probe legacy installation MAY exist at this path and it needs to handle this case (it's already handled with a prompt that asks the user to either delete it or make a backup of it).

On second thought, we should not be breaking existing installations. A better usage of our time seems to use a different directory. I would also give the new binary a different name, e.g., `ooniprobe3`, and just leave legacy installations alone. I bet people that has a working installation of ooniprobe legacy and want to just try out `ooniprobe3` to give us suggestions as it evolves would be very pleased by this arrangement. Also, this allows us to tell people to keep them back to back until we have replaced all features (i.e. less pressure on us).

> It's also worth noting that in the desktop installation this is set to another path specific for the desktop app, namely ~/Library/Application\ Support/OONI\ Probe/ooni_home/ on macOS and XXX find what it is on windows on windows

This is an implementation detail of the probe-desktop, I won't add it.

DESIGN.md
bassosimone: Yes. Also, this is an implementation detail. As far as the probe-desktop is concerned, this is the command to show a measurement and it doesn't matter how it's done.

DESIGN.md
bassosimone: Are these settings something that the desktop app can set? If so, then they are missing, otherwise, they're just stuff we're promising to have before we're sure it's stable.

DESIGN.md
bassosimone: These are internal fields. I am not sure we want to document them. Probably not. I am also confused by the `auto_update` field. This is a setting for probe-desktop, as far as I get the design, so I believe it should not really be the probe-cli concern.

bassosimone:
> [...]
>
> May I suggest that maybe we have this design document focus on the --batch JSON API for exposing events and data to the probe-desktop app and keep the docs and discussions related to the above in github issues or other less formal documents?

Yes, good suggestion. My current view of a design document is that we should only specify what is he design in light of relevant use cases. Because the current use case is the desktop app, I am in agreement with you that the only thing that matters is the desktop app.

When/if we will decide that we want to expose an interface for running as a daemon under Unix (in whatever form we want to do that, including systemd), then this would be another interface to specify. Also in such case, I'd only focus on what a machine would expected from probe-cli. I do not want to go down the path of promising the human readable output won't change.

> Off the top of my heads things that are missing are:
>
> That data usage is being tracked and recorded to the filesystem
> How logging is handled and written to disk
> The database schema and logic for accessing and updating it
> How database migrations should be handled
> The logic for download and update of geoip data files
> How crash reporting is handled

Yeah, all of this strikes me as an implementation detail, I'll lave this out.

DESIGN.md
hellais: `send_crash_reports` they are currently not exposed in the desktop app. Ideally they would be exposed as settings in the desktop app as they were part of the previous version of OONI Probe.

They are implemented in probe-cli, so it's just a matter of exposing them in the UI code.

I will leave it up to you and @sarathms if you believe it's reasonable to include them or not.

DESIGN.md
hellais: > These are internal fields. I am not sure we want to document them

I guess so, though the desktop app needs to be aware of the `_informed_consent` field in order to know if onboarding should happen or not.

It's not 100% clear to me then what is the goal of this design document.

> I am also confused by the auto_update field. This is a setting for probe-desktop, as far as I get the design, so I believe it should not really be the probe-cli concern.

Yes you are right this is only used by the desktop app, though the settings are shared by both probe-cli and probe-desktop as in they both use the same configuration file.

bassosimone: > It's not 100% clear to me then what is the goal of this design document.

The goal is that, if I give this document to a person, they could write a replacement for OONI Probe CLI that the OONI Probe Desktop could use via the `--batch` interface.

> Yes you are right this is only used by the desktop app, though the settings are shared by both probe-cli and probe-desktop as in they both use the same configuration file.
>
> [..]
>
> I guess so, though the desktop app needs to be aware of the _informed_consent field in order to know if onboarding should happen or not.

I was not aware of this design where the two tools share the same config file. I think I can see the advantages and the drawbacks of this design. I have not a good feeling about this design, and I'd like to spend more time thinking about whether it could serve us well or not.

(Returning to the original topic of the goal of writing a design document, I'd like to add that this is also an opportunity for further discussing and writing down existing assumptions.)

bassosimone:
Lest I forget, further feedback from @sarathms:

1. modify CLI on-boarding to write/check-the-existence-of a separate file, to separate config from state, and clearly specify that the CLI does not write into its state

2. now the CLI cannot write the configuration, hence we're safe from any case where we have two writers for config that diverge causing the CLI to break the desktop app

3. to avoid disrupting existing installations that may still be useful, use `~/.ooni-beta/v3` rather than `~/.ooni` for the time being and until we have replaced all the previous functionality

4. `ooniprobe --batch retry <measurement-id>`

5. `ooniprobe --batch upload <measurement-id>`

bassosimone:
This design document effort did not get enough traction and love. I am going to close this PR.

PR #65, opened 2019-10-03
build.sh: produce same release as before

### LINKED ISSUE #64: Build script changes break download script on probe-desktop
The released assets now have an additional segment in the file names viz. `$(git describe)`
image

These files are now not downloadable using the release version tag in the [download script](https://github.com/ooni/probe-desktop/blob/master/scripts/download-bin.js#L8) on `probe-desktop`

The file ` ooniprobe_checksums.txt` is also missing from the downloadable assets. which was being used by the download script.

Are these changes intended and to be reflected in `probe-desktop` too?

bassosimone:
Closes #64

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -27,13 +27,19 @@ elif [ "$1" = "release" ]; then
   set -x
   v=`git describe --tags`
   $0 linux
-  tar -czf ooniprobe-linux-amd64-$v.tar.gz ./dist/linux/amd64
+  tar -czf ooniprobe_${v}_linux_amd64.tar.gz LICENSE.md Readme.md              \
+    -C ./dist/linux/amd64 ooniprobe
+  shasum -a 256 ooniprobe_${v}_linux_amd64.tar.gz > ooniprobe_checksums.txt
   $0 macos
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -27,13 +27,19 @@ elif [ "$1" = "release" ]; then
...
+    -C ./dist/macos/amd64 ooniprobe
+  shasum -a 256 ooniprobe_${v}_darwin_amd64.tar.gz >> ooniprobe_checksums.txt
   $0 windows
-  tar -czf ooniprobe-windows-amd64-$v.tar.gz ./dist/windows/amd64
+  tar -czf ooniprobe_${v}_windows_amd64.tar.gz ./dist/windows/amd64            \
+    -C dist/windows/amd64 ooniprobe.exe
+  shasum -a 256 ooniprobe_${v}_windows_amd64.tar.gz >> ooniprobe_checksums.txt
   echo ""
-  echo "Now sign the packages and upload them to GitHub"
+  echo "Now sign ooniprobe_checksums.txt packages and upload all to GitHub"
```
sarathms:
```suggestion
  echo "Now sign ooniprobe_checksums.txt and upload it along with tarballs to GitHub"
```

Issue #25, opened 2018-12-11
Mention not implemented config options


anadahz:
The config file should mention options that have no effect or not currently used similar to the help page message: `this function is not implemented`.

hellais:
Currently it's probably easier to document what is implemented (in terms of settings), than what is not implemented.

These are the settings that should work:
* `sharing` -> all of them
* `advanced.send_crash_reports`
* `advanced.bouncer_url`
* `advanced.collector_url` (will work as a failover when bouncer communication does not work)

bassosimone:
@hellais do we want to implement all config options for the release? Are there any feature that we may want to not ship to let the release happen sooner?

hellais:
I think the most high priority config options are documented in: https://github.com/ooni/probe-cli/issues/39

PR #73, opened 2019-11-14
Remove all settings which are not implemented

hellais:
To focus on a very minimal MVP.

This also fixes: https://github.com/ooni/probe-cli/issues/25.

What I still need to do in this branch is handle the migration from a beta user and a stable release user.

I suggest how we handle the migration is simply zapping the old configuration and starting fresh when we notice the presence of `_is_beta: true`. We should maybe also consider how to handle config file migrations in the future.

### DIFF
--- a/config/parser.go
+++ b/config/parser.go
@@ -57,12 +57,8 @@ type Config struct {
 	InformedConsent bool   `json:"_informed_consent"`
 	IsBeta          bool   `json:"_is_beta"` // This is a boolean flag used to indicate this installation of OONI Probe was a beta install. These installations will have their data deleted across releases.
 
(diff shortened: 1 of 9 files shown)


hellais:
Another thing we should decide is the default settings for `websites_url_limit`. It currently defaults to 0, which means test all the things, yet this means that the runtime of the websites test is going to be several hours.

Maybe we should set it to something more sane like 100 or 200?

@agrabeli what do you think?

hellais:
We discussed this and maybe if we are able to make the test cancellable we should keep the default of the full list, but allow users to cancel the test mid-way.

This should be possible by just doing `chilprocess.kill()` on the `ooniprobe` CLI.

Issue #37, opened 2019-05-21
Add support for specifying custom test lists


hellais:
Currently it's not possible to pass your own test-list to probe-cli.

PR #82, opened 2019-11-29
Add document outlining the differences between probe-cli and probe-le…

hellais:
…gacy

### DIFF
--- a/docs/OONIProbeLegacyCompatibility.md
+++ b/docs/OONIProbeLegacyCompatibility.md
@@ -58,7 +58,7 @@ Priority: medium
 
 `-O, --pcapfile PATH_TO_PCAPFILE` Write a PCAP of the ooniprobe session to this filename.
 Supported: ❌
-Priority: wontfix, we don't have packet capture support in probe-cli
+Priority: low
 
 `-f, --configfile PATH_TO_CONFIG` Specify a path to the ooniprobe configuration file.
 Supported: ✅
@@ -109,7 +109,7 @@ Priority: high
 Supported: ❌
 Priority: high
 
-* Write custom tests like OONI test templates
+* Public API for writing OONI tests (see: https://github.com/ooni/probe-legacy/blob/master/docs/source/writing_tests.rst)
 Supported: ❌
 Priority: medium
 
--- a/docs/OONIProbeLegacyCompatibility.md
+++ b/docs/OONIProbeLegacyCompatibility.md
@@ -61,13 +61,13 @@ Supported: ✅
 probe-cli equivalent: set the `OONI_HOME` environment variable
 
 `-a, --annotations key:value[,key2:value2]` Annotate the report with a key:value[, key:value] format.
-Supported: ✅
+Supported: ❌
 Priority: high
 
 `-P, --preferred-backend onion|https|cloudfront` Set the preferred backend to use when submitting results and/or communicating with test helpers. Can be either onion, https or cloudfront
 Supported: ❌
 Priority: wontfix, we don't support any other backend beyond https, yet we will
-and yet we would prefer to have the logic of reporting be managed by the probe
+and we would prefer to have the logic of reporting be managed by the probe
 itself and not expose this setting.
 
 # Features
--- a/docs/OONIProbeLegacyCompatibility.md
+++ b/docs/OONIProbeLegacyCompatibility.md
@@ -77,7 +77,7 @@ Supported: ✅
 probe-cli equivalent: we now call a test deck a test group and we have them
 coded into the logic of the client
 
-* Run an individual netttest
+* Run an individual nettest
 Supported: ❌
 Priority: medium
 
--- a/docs/OONIProbeLegacyCompatibility.md
+++ b/docs/OONIProbeLegacyCompatibility.md
@@ -4,6 +4,14 @@
 Supported: ✅
 probe-cli equivalent: `-h, --help`
 
+`--version` Display the version exit
+Supported: ✅
+probe-cli equivalent: `ooniprobe version`
+
+`-v, --verbose` Enable verbose log output
+Supported: ✅
+probe-cli equivalent: `ooniprobe -v`
+
 `-n, --no-collector` Disable writing to collector
 Supported: ❌
 Priority: high
@@ -69,6 +77,10 @@ Supported: ❌
 Priority: wontfix, we don't support any other backend beyond https, yet we will
 and we would prefer to have the logic of reporting be managed by the probe
 itself and not expose this setting.
+
+`-Q, --queue`, AMQP Queue URL. amqp://user:pass@host:port/vhost/queue
+Supported: ❌
+Priority: wontfix
 
 # Features
 
--- a/docs/OONIProbeLegacyCompatibility.md
+++ b/docs/OONIProbeLegacyCompatibility.md
@@ -1,39 +1,39 @@
 # Command line flags
 
 `-h, --help` Display help and exit
-Supported: yes
+Supported: ✅
 probe-cli equivalent: `-h, --help`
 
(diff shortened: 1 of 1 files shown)


docs/OONIProbeLegacyCompatibility.md
```diff
@@ -0,0 +1,118 @@
...
+Supported: ✅
+probe-cli equivalent: `--config`
+
+`-d, --datadir` Specify a path to the ooniprobe data directory.
+Supported: ✅
+probe-cli equivalent: set the `OONI_HOME` environment variable
+
+`-a, --annotations key:value[,key2:value2]` Annotate the report with a key:value[, key:value] format.
+Supported: ✅
+Priority: high
```
bassosimone: Confusing: why add priority if supported? Also: you sure it's supported?

bassosimone: Suggestion: `-a key:value -a key2:value2` is easier to handle than what we had previously. (MK uses `-A` for that precisely because the probe-legacy behaviour is not implemented.)

bassosimone: (xref: https://github.com/ooni/probe-engine/issues/132)

docs/OONIProbeLegacyCompatibility.md
```diff
@@ -0,0 +1,118 @@
...
+itself and not expose this setting.
+
+# Features
+
+* Run a test deck
+Supported: ✅
+probe-cli equivalent: we now call a test deck a test group and we have them
+coded into the logic of the client
+
+* Run an individual netttest
```
bassosimone: remove the third `t`, please

docs/OONIProbeLegacyCompatibility.md
```diff
@@ -0,0 +1,118 @@
...
+
+* Test an individual URL with web_connectivity
+Supported: ❌
+Priority: high
+
+* Run tests automatically like `ooniprobe-agent`
+Supported: ❌
+Priority: high
+
+* Write custom tests like OONI test templates
```
bassosimone: I don't understand what you mean here

docs/OONIProbeLegacyCompatibility.md
```diff
@@ -0,0 +1,118 @@
...
+Supported: ❌
+Priority: medium
+
+* Packet captures
```
bassosimone: I understood from above it was wontfix?

docs/OONIProbeLegacyCompatibility.md
```diff
@@ -0,0 +1,118 @@
...
+probe-cli equivalent: set the `OONI_HOME` environment variable
+
+`-a, --annotations key:value[,key2:value2]` Annotate the report with a key:value[, key:value] format.
+Supported: ✅
+Priority: high
+
+`-P, --preferred-backend onion|https|cloudfront` Set the preferred backend to use when submitting results and/or communicating with test helpers. Can be either onion, https or cloudfront
+Supported: ❌
+Priority: wontfix, we don't support any other backend beyond https, yet we will
+and yet we would prefer to have the logic of reporting be managed by the probe
```
anadahz: This seems like a syntax error: `yet we will and yet we would prefer`

docs/OONIProbeLegacyCompatibility.md
hellais: Ah yes you are right, this is not supported.

docs/OONIProbeLegacyCompatibility.md
hellais: I mean that we don't have a public API for writing tests like we had for OONI Probe 2.0.0, see: https://github.com/ooni/probe-legacy/blob/master/docs/source/writing_tests.rst

docs/OONIProbeLegacyCompatibility.md
hellais: Yeah there was an extra yet probably. I removed it.

docs/OONIProbeLegacyCompatibility.md
hellais: Maybe, I would not exclude it a-priori though. What do you think?

docs/OONIProbeLegacyCompatibility.md
```diff
@@ -0,0 +1,118 @@
...
+Supported: ❌
+Priority: medium
+
+* Measurement quota to limit the disk usage
+Supported: ❌
+Priority: high
+
+* Failover strategies for uploading measurements (use https then onion then cloudfront)
+Supported: ❌
+Priority: high
```
anadahz: Doesn't this contradict with?: https://github.com/ooni/probe-cli/pull/82/files#diff-a7b5bc713d5d56af7ba54f408fd8a4d7R67-R71
>wontfix, we don't support any other backend beyond https

anadahz:
Some options that are not included in the outline:

* -v, --verbose           Show more verbose information
* -Q, --queue=           AMQP Queue URL,  amqp://user:pass@host:port/vhost/queue
* --version                 Display the ooniprobe version and exit.

Comments:
* Are the configuration options (in config file) in scope of this outline?
* Can you create/run your own custom test-list (not deck) in probe-cli?
* It will be very useful to see the available tests or tests that are not going to be implemented in probe-cli

docs/OONIProbeLegacyCompatibility.md
hellais: We will support failover strategies, but we will not expose settings for choosing this as the logic is to be handled by the probe itself, see:
> we would prefer to have the logic of reporting be managed by the probe itself and not expose this setting.

hellais:
> Some options that are not included in the outline:

Thanks added these in: https://github.com/ooni/probe-cli/pull/82/commits/e4ff1fe89a7a5a29ee48d80dab9b649f692309d3

> Are the configuration options (in config file) in scope of this outline?

Yes config options are in scope. I believe the missing relevant ones should all be documented inside of the features section.

If you think some are missing, let me know.

I did not document stuff which is not relevant like the measurement and reporting concurrency, but if you think it's important to document those too, feel free to open a PR with that too. They are likely `wontfix`, unless there is a strong argument in their favour.

> Can you create/run your own custom test-list (not deck) in probe-cli?

No that is not supported. See: https://github.com/ooni/probe-cli/issues/37 and in this file:
> * Test an individual URL with web_connectivity

I added an extra item just for test lists from files.

> It will be very useful to see the available tests or tests that are not going to be implemented in probe-cli

You can find all the supported tests in here: https://github.com/ooni/probe-engine/blob/master/experiment.go#L269.

Basically it's all the tests inside of the mobile app + psiphon. You can find the list of tests we have plans for in here: https://github.com/ooni/probe-engine/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3A%22new+experiment%22+ or https://github.com/ooni/spec/issues?q=is%3Aissue+is%3Aopen+label%3A%22new+test%22.

If there is a test you think is especially important to implement give the issue a thumbs up! 👍

docs/OONIProbeLegacyCompatibility.md
bassosimone: Yeah, but then you should not flag it as wontfix above?

docs/OONIProbeLegacyCompatibility.md
bassosimone: Gotcha! Can you then maybe rephrase this to indicate we want to have a public API for writing tests?

docs/OONIProbeLegacyCompatibility.md
hellais: Ok I will put the above one as not wontfix then.

hellais:
@bassosimone I have addressed you feedback and will now merge. Thanks for the review!

PR #83, opened 2019-12-02
Refactoring to reduce package count

hellais:
* Consolidate util and utils into the same package
* Move internal/onboard into internal/cli/onboard
* Move maybeOnboard into the onboard package
* Add code coverage checks

What I am looking for in a review is:
1. If it seems like a good idea to go for this approach of reducing the package count and making the directory structure a bit more flat
2. If anything pops out as odd or unusual in the refactoring

No new code has been added (beyond adding new tests and code coverage checks).


internal/cli/onboard/onboard.go
```diff
@@ -1,14 +1,164 @@
...
+		if c.IsBatch == true {
+			return errors.New("cannot run onboarding in batch mode")
+		}
+		if err := Onboarding(c.Config); err != nil {
+			return errors.Wrap(err, "onboarding")
+		}
+	}
+	return nil
+}
+
```
hellais: This has just been moved into here from the separate onboarding package.

nettests/facebook_messenger.go
```diff
@@ -1,15 +1,11 @@
-package im
-
-import (
-	"github.com/ooni/probe-cli/nettests"
-)
+package nettests
```
hellais: All nettests have been flattened into the root nettests package. I think eventually we should move all the nettest specific code into probe-engine as there is currently so many levels of indirection going on here.

nettests/nettests_test.go
```diff
@@ -0,0 +1,46 @@
...
+	network, err := database.CreateNetwork(ctx.DB, ctx.Session)
+	if err != nil {
+		t.Fatal(err)
+	}
+	res, err := database.CreateResult(ctx.DB, ctx.Home, "middlebox", network.ID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	nt := HTTPInvalidRequestLine{}
+	ctl := NewController(nt, ctx, res)
```
hellais: This is a very basic smoke test.

nettests/facebook_messenger.go
bassosimone: Yes, I believe we should aim for slowly moving code in probe-engine.

Issue #76, opened 2019-11-19
Make it possible to cleanly stop a currently running test session


hellais:
Currently if you hit ctrl-c while a test is running, it will not shutdown cleanly and you will end up with a bunch of test results in the `is_done: false` state which will mean they are not shown in the result listing.

This is about adding support for handling the `SIGINT` signal and doing a clean shutdown so as to close the current report which is open and put the measurements in a consistent state.

bassosimone:
Yeah, we need to modify the way in which `probe-cli` work to have a cancellable context that is cancelled when the user hits `^C`. There is a bunch of cases where the cancellation cannot probably be immediate, unless we further refactor code inside of probe-engine to cope with that. In principle, most Go code is cancellable, if you use a `context.Context`. We do that in `probe-engine` but it may be that some operations are not immediately cancellable anyway (e.g. blocked for I/O), thus there is need to wrap them in a goroutine to make them fully cancellable.

hellais:
This has been implemented in #84.

PR #84, opened 2019-12-19
Handle the SIGINT and SIGTERM signals to support stopping a test cleanly

hellais:
This fixes: #76

### DIFF
--- a/internal/cli/run/run.go
+++ b/internal/cli/run/run.go
@@ -27,7 +27,7 @@ func listenForSignals(ctx *ooni.Context) {
 
 	go func() {
 		<-s
-		log.Debugf("caught a signal, shutting down cleanly")
+		log.Info("caught a stop signal, shutting down cleanly")
 		ctx.Terminate()
 	}()
 }
--- a/internal/cli/run/run.go
+++ b/internal/cli/run/run.go
@@ -2,6 +2,9 @@ package run
 
 import (
 	"errors"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/alecthomas/kingpin"
 	"github.com/apex/log"
@@ -12,6 +15,20 @@ import (
 	"github.com/ooni/probe-cli/internal/database"
 	"github.com/ooni/probe-cli/nettests"
 )
+
+// listenForSignals will listen for SIGINT and SIGTERM. When it receives those
+// signals it will set isTerminated to true, which will cleanly shutdown the
+// test logic
+func listenForSignals(ctx *ooni.Context) {
+	s := make(chan os.Signal, 1)
+	signal.Notify(s, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-s
+		log.Debugf("caught a signal, shutting down cleanly")
+		ctx.IsTerminated = true
+	}()
+}
 
 func runNettestGroup(tg string, ctx *ooni.Context, network *database.Network) error {
 	group, ok := nettests.NettestGroups[tg]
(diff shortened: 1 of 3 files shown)


internal/cli/run/run.go
```diff
@@ -13,6 +16,20 @@ import (
...
+// signals it will set isTerminated to true, which will cleanly shutdown the
+// test logic
+func listenForSignals(ctx *ooni.Context) {
+	s := make(chan os.Signal, 1)
+	signal.Notify(s, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-s
+		log.Debugf("caught a signal, shutting down cleanly")
+		ctx.IsTerminated = true
```
bassosimone: Please, use `atomic` to make sure we're accessing the variable with atomic semantics. I am thinking at something like `atomic.AddInt64(&ctx.IsTerminated)` to stop and `atomic.LoadInt64(&ctx.IsTerminated)` to fetch it.

Unfortunately there is no atomic boolean. If you wish to retain a boolan type, otherwise use a `sync.Mutex` but I have a mild preference for atomic semantics because it's more robust than explicitly using mutexes.

internal/cli/run/run.go
```diff
@@ -13,6 +16,20 @@ import (
 	"github.com/ooni/probe-cli/nettests"
 )
 
+// listenForSignals will listen for SIGINT and SIGTERM. When it receives those
+// signals it will set isTerminated to true, which will cleanly shutdown the
+// test logic
```
bassosimone: Please, document that the proper way to attain this objective would instead be to use a `context.Context` and cancel it when a signal is delivered, but this kind of refactoring is too complex for now. It would perhaps be top to have an issue that suggests to use a cancellable context rather than just adding a TODO comment to the code.

hellais:
>  I have a request to use some form of atomic semantic for the flag variable and I would also like to see a follow-up issue describing using a cancellable context.Context.

I agree with you that using a `context.Context` object would probably be the optimal pattern, yet I think we would need to do quite some more refactoring in order to obtain that, so I suggest we defer that to future work.

Regarding the atomic semantics for the flag, I am not sure I properly understand what race condition or inconsistency you are concerned about.
The `IsTerminated` bool variable is initialised once with the Context and is only every written once when the signal handler channel is triggered: https://github.com/ooni/probe-cli/pull/84/files#diff-be745cb99b81f008f99c3b674683ef40R29.

Throughout the rest of the code this variable is only ever read, so I am not really sure what we are gaining by using an `atomic` construct. For example if there is a "race" when the `IsTerminated` is written just in the moment that it's being read in https://github.com/ooni/probe-cli/pull/84/files#diff-be745cb99b81f008f99c3b674683ef40R49 or https://github.com/ooni/probe-cli/pull/84/files#diff-eacd810feae41332a4946c2af72b0b5aR110, that is not a problem. We will just do an extra loop and terminate cleanly on the next iteration.

Since the setup is that of a single writer, I don't see it possible for a race condition and I think using the atomic construct is unnecessary.

hellais:
Following a private chat with @bassosimone I was enlightened by the magic that happens in the golang memory model, see: https://golang.org/ref/mem.

Some key takeways from that text include:

>  Worse, there is no guarantee that the write to done will ever be observed by main, since there are no synchronization events between the two threads. The loop in main is not guaranteed to finish.

and

>  The Go memory model specifies the conditions under which reads of a variable in one goroutine can be guaranteed to observe values produced by writes to the same variable in a different goroutine.

Moreover a minimal version of our code like this:
```
package main
import "time"
type Context struct {
        done bool
}
func main() {
	s := make(chan bool, 1)
	ctx := new(Context)
	go func() {
		<-s
		ctx.done = true
	}()
	go func() {
		time.Sleep(3 * time.Second)
		s <- true
	}()
	for true {
		if (ctx.done == true) {
			break
		}
		time.Sleep(1 * time.Second)
	}
}
```

results in a race being detected by the golang race detector.

hellais:
I pushed changes that implement the recommended atomic patterns.

PR #92, opened 2019-12-28
Release 3.0.0-rc.5

bassosimone:


nettests/psiphon.go
```diff
@@ -0,0 +1,33 @@
...
+	)
+	if err != nil {
+		return err
+	}
+	return ctl.Run(builder, []string{""})
+}
+
+// PsiphonTestKeys contains the test keys
+type PsiphonTestKeys struct {
+	IsAnomaly bool `json:"-"`
```
hellais: I think we should also expose as a test key:
```
        "bootstrap_time": 6.076582013,
        "failure": null,
```

internal/log/handlers/cli/results.go
```diff
@@ -69,6 +69,13 @@ var summarizers = map[string]func(uint64, uint64, string) []string{
 			"",
 		}
 	},
+	"circumvention": func(totalCount uint64, anomalyCount uint64, ss string) []string {
+		return []string{
+			fmt.Sprintf("Detected: %v", anomalyCount > 0),
```
hellais: This should probably say something like: "%d tested" and "%d blocked".

nettests/psiphon.go
bassosimone: https://github.com/ooni/probe-cli/issues/96

internal/log/handlers/cli/results.go
bassosimone: https://github.com/ooni/probe-cli/issues/97

nettests/psiphon.go
bassosimone: See #96

PR #101, opened 2020-01-27
Use ~/.ooniprobe as the home directory

hellais:
Remove all probe-legacy related to code since there is no more conflict
between the two
Fixes: ooni/probe#972


.gitignore
```diff
@@ -1,6 +1,7 @@
 /dist
 /ooniprobe
 /coverage.cov
+/ooni.cov
```
bassosimone: Why do we have two differently named files for coverage?

hellais: I suspect `coverage.cov` is how it used to be called at some point in the past. Maybe we should look into dropping that one from the `.gitignore` or maybe just do `.gitignore` for `*.cov`?

bassosimone: I am confused by what I see, tho:

```bash
sbs@synchronicity probe-cli % git grep ooni.cov
sbs@synchronicity probe-cli % git grep coverage.cov
.gitignore:/coverage.cov
build.sh:  $0 __docker go test -v -coverprofile=coverage.cov -coverpkg=./... ./...
build.sh:          -coverprofile=coverage.cov -service=travis-ci
build.sh:  go test -v -race -coverprofile=coverage.cov -coverpkg=./... ./...
sbs@synchronicity probe-cli % git branch 
* issue/973
  master
```

bassosimone: (`issue/973` is branched off the current `master`)

hellais:
```
probe-cli [fix/ooni_home] % ls -l ooni.cov
-rw-r--r--  1 x  staff  16043 Dec  2 14:17 ooni.cov
````

Hum ok, maybe this was an old file I had lying around from a time prior to some refactor?

bassosimone: ah, could be!

bassosimone:
I'm going to squash and merge (need to merge another PR soon and tag a release).

PR #102, opened 2020-01-28
One measurement, One file

hellais:
This implements what is described in: ooni/probe#949.

What is still missing from this PR is how to handle the case in which on disk we have report_file (many measurements in a file) and we are trying to access it.

Ideas for it can be:

* (1) We do the migration to a single measurement per entry by going through the file line by line
* (2a) We don't care about them and we just delete them from the database, but keep them on disk
* (2b) We don't care about them and we just delete them from the database **and** from disk
* (3) We keep in place the code that reads stuff from a single measurement and stuff that reads from a report_file. Eventually we tell a user to either upload them or we will delete them from disk.

I am more leaning towards the options in the 2-3 range, but still need to research this a bit more to actually understand how complex it is to do a full migration of the files. The tricky thing is that we have to match the measurement row to a specific line in the file which may not be matched.

The only problem exists for websites test as the IM and other tests already have just a single row.


bassosimone:
@hellais I suggest to ignore them since technically this is not released code.

hellais:
> I suggest to ignore them since technically this is not released code

Can you clarify what you mean exactly by "ignore them"? 
Do we delete them from the database and on disk, do we keep them, but don't show them, do we keep them, but crash when you try to access them or ??

hellais:
As discussed in private, I have added a check to see if the `measurement_file_path` is valid and if not ask the user to run `ooniprobe reset`.

This is PR is now ready for review.

build.sh
```diff
@@ -4,6 +4,14 @@ set -e
 buildtags=""
 ldflags="-s -w"
 
+if [ "$1" = "bindata" ];then
```
bassosimone: Missing space after `;`

build.sh
```diff
@@ -4,6 +4,14 @@ set -e
...
+    GO_BINDATA_V=$(go-bindata -version | grep go-bin | cut -d ' ' -f2)
+    [[ "$GO_BINDATA_V" == "3.2.0" ]] && echo "Updating bindata" || exit "Wrong go-bindata-version"
```
bassosimone: I think `[[` is a bashism and we're using `/bin/sh` here. How about using `[` instead?

internal/database/actions.go
```diff
@@ -55,44 +55,22 @@ func GetMeasurementJSON(sess sqlbuilder.Database, measurementID int64) (map[stri
...
+		log.Error("backup your OONI_HOME and run `ooniprobe reset`")
+		return nil, errors.New("cannot access measurement file")
 	}
-	// When the URL is a string then we need to seek until we reach the
-	// measurement line in the file that matches the target input
-	url := measurement.URL.URL.String
-	file, err := os.Open(reportFilePath)
+	measurementFilePath := measurement.Measurement.MeasurementFilePath.String
+	// TODO handle the case in which we have MeasurementFilePath == NULL because
+	// it's a beta measurement
```
bassosimone: Is this comment still applicable?

build.sh
hellais: Oh, why do we want to use `/bin/sh`?

internal/database/actions.go
hellais: Ah right, it isn't.

build.sh
hellais: Or put differently, you really have no guarantee what `/bin/sh` is going to be, so why not just use `#!/bin/bash` which is going to be present on most systems that are going to be used to build a CLI golang tool?

hellais: I am not sure the extension change is really necessary as the shebang has changed. Anyways since the reason why I changed it is so small I will just use the more portable way of doing the same pattern and revert this.

PR #103, opened 2020-01-28
Add example launchd service for running OONI Probe automatically ever…

hellais:
…y hour

You should place the file inside of: `$HOME/Library/LaunchAgents` and then enable by running `launchctl load org.ooni.probe.cli.plist`

It assumes you have a `ooniprobe` binary in `~/.ooniprobe/ooniprobe` and a special config file named `/Users/USERNAME/.ooniprobe/config-100sites.json` with a URL limit of 100 sites per run:
```
  "nettests": {
    "websites_url_limit": 100
  },
```


scripts/org.ooni.probe.cli.plist
```diff
@@ -0,0 +1,34 @@
...
+    <key>Label</key>
+    <string>org.ooni.probe.daily-run</string>
+
+    <key>KeepAlive</key>
+    <false/>
+    <key>RunAtLoad</key>
+    <true/>
+
+    <key>Program</key>
+    <string>/Users/USERNAME/.ooniprobe/ooniprobe</string>
```
bassosimone: I don't understand what is this location

scripts/org.ooni.probe.cli.plist
```diff
@@ -0,0 +1,34 @@
...
+    <key>ProgramArguments</key>
```
bassosimone: Isn't this duplicated?

scripts/org.ooni.probe.cli.plist
```diff
@@ -0,0 +1,34 @@
...
+    <key>ProgramArguments</key>
+    <array>
+        <string>--config="/Users/USERNAME/.ooniprobe/config-100sites.json"</string>
```
bassosimone: Does this exists by default?

scripts/org.ooni.probe.cli.plist
```diff
@@ -0,0 +1,34 @@
...
+        <string>--batch</string>
```
bassosimone: Why `--batch`?

scripts/org.ooni.probe.cli.plist
```diff
@@ -0,0 +1,34 @@
...
+        <string>run</string>
+    </array>
+
+    <key>StartInterval</key>
+    <integer>3600</integer>
```
bassosimone: The main concern that I have is that subsequent experiments may overlap

scripts/org.ooni.probe.cli.plist
hellais: It's the location of your `ooniprobe` binary. It's meant to be an example that you should edit prior to use.

scripts/org.ooni.probe.cli.plist
hellais: Oops, yeah. For some reason it works even if it has a duplicated key. Will fix it though.

scripts/org.ooni.probe.cli.plist
hellais: Nope, as mentioned in the commit you should make a special config for this automatic test which limits to 100 websites, so that it's more likely to converge in 1h.

scripts/org.ooni.probe.cli.plist
hellais: To make the output into the log file a bit more easy to parse.

scripts/org.ooni.probe.cli.plist
hellais: The should not happen if you set the limit to 100 URLs, but some more testing is probably required to validate this.

On my network, which is pretty fast, I run this consistently in under 10 minutes, but actually closer to 5 minutes.

hellais:
Following your comments I figured it probably makes most sense to place this configuration inside of the `Readme.md` rather than have a file on disk. I have done that now.

PR #106, opened 2020-02-07
Explicitly handle SIGINT

hellais:
Tentative fix for: https://github.com/ooni/probe/issues/1005


bassosimone:
Tomorrow I should be able to test on Windows.

hellais:
@bassosimone also check the backlog of the #ooni-probe channel. @sarathms did some debugging of this issue on windows too.

sarathms:
From what I tried, the `probe-cli` part of the interaction works fine on Windows. I am able to run a test and do a 'Ctrl+C' which gets caught and initiates a clean shutdown. The part that fails is `probe-desktop` doing the equivalent of a `Ctrl+C` because `child_process.kill('SIGINT')` doesn't generate the expected signal on Windows.

PR #112, opened 2020-02-14
build.sh: pass `-tags ooni` to build

bassosimone:
See https://github.com/ooni/probe-engine/pull/333

Closes https://github.com/ooni/probe-engine/issues/267


bassosimone:
This is in draft status until https://github.com/ooni/probe-engine/pull/333 lands.

PR #116, opened 2020-02-26
deps: latest upper.io/db and pin probe-engine

bassosimone:
We're specifically pinning a commit of probe-engine that is using
MK v0.10.11, so that we address the following issues:

1. we use db-ip.com for the country database as opposed to using
the increasingly stale MaxMind database (see https://github.com/ooni/probe-engine/issues/334)

2. we're using an implementation of WhatsApp that does not suffer
from the super-old CIDR bug (see https://github.com/ooni/probe-engine/issues/341)

3. we're not linking to libcurl anymore on Windows and Linux, thanks to
this new version of MK where we can optionally disable libcurl; we are
still linking to libcurl on macOS, but that has no impact on the binary
size since on macOS libcurl is part of the system

This should be enough, from my side to bless a new release of the
probe-cli (see https://github.com/ooni/probe/issues/1028).


.dockerignore
```diff
@@ -1,2 +1,4 @@
 /dist
+/ooniprobe
+/ooniprobe.exe
```
bassosimone: Send less useless context to Docker daemon

Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM openobservatory/mk-alpine:latest
+FROM openobservatory/mk-alpine:20200226
```
bassosimone: Reproducible builds

build.sh
```diff
@@ -22,7 +22,7 @@ fi
 if [ "$1" = "windows" ]; then
   set -x
   CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++                         \
-    CGO_ENABLED=1 GOOS=windows GOARCH=amd64                                    \
+    CGO_LDFLAGS_ALLOW='-fstack-.*' CGO_ENABLED=1 GOOS=windows GOARCH=amd64     \
```
bassosimone: Now, sadly, we need to white list this flag to correctly link on Windows

build.sh
```diff
@@ -48,7 +48,7 @@ elif [ "$1" = "release" ]; then
     -C ./dist/macos/amd64 ooniprobe
   shasum -a 256 ooniprobe_${v}_darwin_amd64.tar.gz >> ooniprobe_checksums.txt
   $0 windows
-  tar -czf ooniprobe_${v}_windows_amd64.tar.gz ./dist/windows/amd64            \
+  tar -czf ooniprobe_${v}_windows_amd64.tar.gz LICENSE.md Readme.md            \
```
bassosimone: While there, fix the distribution generation on Windows, which was broken

PR #117, opened 2020-03-02
Initial packaging

FedericoCeratto:
Set Travis to use Bionic
Create a simple Debian package
Do not upload into bintray yet


debian/changelog
```diff
@@ -0,0 +1,5 @@
+ooniprobe-cli (3.0.0-rc.9-1) UNRELEASED; urgency=medium
```
bassosimone: This is probably `-rc.10`

debian/ooniprobe-cli.install
```diff
@@ -0,0 +1 @@
+dist/linux/amd64/ooniprobe usr/sbin
```
bassosimone: Why `sbin`? It strikes me more as a tool for `bin`

Readme.md
```diff
@@ -1,5 +1,7 @@
 # OONI Probe CLI
 
+![Build packages](https://github.com/ooni/probe-cli/workflows/Build%20packages/badge.svg)
```
bassosimone: This strikes me as stray, since here we're using travis

.travis.yml
```diff
@@ -1,8 +1,15 @@
...
+dist: bionic
 language: minimal
 services:
 - docker
 script:
-- ./build.sh _travis-${TRAVIS_OS_NAME}
-- ./scripts/travis_test.sh
+- ./build.sh linux
+- find . -name ooniprobe -type f
+- ./debian/build_package.sh
```
bassosimone: I think this should only be invoked when on a tag, unless we're talking about nightly packages here

debian/control
```diff
@@ -0,0 +1,17 @@
...
+Vcs-Git: https://salsa.debian.org/go-team/packages/probe-cli.git
+Homepage: https://github.com/ooni/probe-cli
+Rules-Requires-Root: no
+XS-Go-Import-Path: github.com/ooni/probe-cli
+
+Package: ooniprobe-cli
+Architecture: any
+Depends: ${misc:Depends}, ${shlibs:Depends}
+Built-Using: ${misc:Built-Using}
+Description: Next generation OONI Probe CLI
```
bassosimone: I think we can omit "next generation" since it's not so useful for users?

debian/control
```diff
@@ -0,0 +1,17 @@
+Source: ooniprobe-cli
+Maintainer: Federico Ceratto <federico@openobservatory.org>
+Section: devel
+Priority: optional
+Build-Depends: dpkg-dev, debhelper-compat (= 12)
+Standards-Version: 4.5.0
+Vcs-Browser: https://salsa.debian.org/go-team/packages/probe-cli
+Vcs-Git: https://salsa.debian.org/go-team/packages/probe-cli.git
```
bassosimone: These two URLs strike me as unexpected

debian/control
```diff
@@ -0,0 +1,17 @@
...
+Homepage: https://github.com/ooni/probe-cli
+Rules-Requires-Root: no
+XS-Go-Import-Path: github.com/ooni/probe-cli
+
+Package: ooniprobe-cli
+Architecture: any
```
bassosimone: Is there an architecture that debian supports and golang does not support?

.travis.yml
hellais: I guess it doesn't hurt to check if we can build the package on every PR, no?

debian/changelog
hellais: Is there a way to generate this from the git tags? Is there some pattern for doing that in debian land?

debian/control
hellais: Yes I am also not sure why these are here. Maybe it's some debian internal thing.

debian/control
hellais: I guess that was taken from the github description field.

Maybe we should agree on what we use as a description for the package everywhere and @agrabeli can maybe help with this copy.

debian/copyright
```diff
@@ -0,0 +1,40 @@
...
+Copyright: 2018 Open Observatory of Network Interference (OONI)
+License: BSD-3-clause
+
+Files: debian/*
+Copyright: 2020 Federico Ceratto <federico@openobservatory.org>
+License: BSD-3-clause
+Comment: Debian packaging is licensed under the same terms as upstream
+
+License: BSD-3-clause
+ Copyright 2018 Open Observatory of Network Interference (OONI), The Tor Project
```
hellais: This should maybe be 2020.

debian/ooniprobe-cli.install
hellais: Yeah I would also have expected that too.

bassosimone (approved):
Thank you @FedericoCeratto! You did an awesome job here and I am very excited that now we have a working Debian package that we can soon sponsor to people! 🐳

bassosimone:
For the records:

```
# cat /etc/apt/sources.list.d/ooni.list 
deb [trusted=yes] https://dl.bintray.com/ooni/ooniprobe-debian/ unstable main
```

PR #119, opened 2020-03-12
nettests/ndt.go: use ndt7 rather than ndt5

bassosimone:
Part of https://github.com/ooni/probe/issues/969


nettests/ndt.go
```diff
@@ -19,73 +19,59 @@ func (n NDT) Run(ctl *Controller) error {
...
-	IsAnomaly  bool    `json:"-"`
+	Upload         float64 `json:"upload"`
+	Download       float64 `json:"download"`
+	Ping           float64 `json:"ping"`
+	MaxRTT         float64 `json:"max_rtt"`
+	AvgRTT         float64 `json:"avg_rtt"`
+	MinRTT         float64 `json:"min_rtt"`
+	MSS            float64 `json:"mss"`
+	RetransmitRate float64 `json:"retransmit_rate"`
+	IsAnomaly      bool    `json:"-"`
```
bassosimone: cc: @sarathms @hellais 

The above diff summarises what changed in the emitted message migrating from ndt5 to ndt7.

bassosimone: RetransmitRate is conceptually similar to PacketLoss, but PacketLoss was too much of a lie to keep the naming.

PR #131, opened 2020-05-20
Add support for extracting tor test targets

hellais:
Implements: https://github.com/ooni/probe/issues/1130


hellais:
Closing this due to: https://github.com/ooni/probe/issues/1130#issuecomment-632683317

PR #133, opened 2020-05-29
Use ooni/probe-engine 0.12.0+patches

bassosimone:
The main patch we're pulling on top of 0.12.0 allows us to build on 32 bit systems.

Also, as mentioned in e061880's commit message:

> The [probe-engine] API has changed. Methods that used to change bits of the session have been removed. Now the session is more immutable than before.
>
> As such, we need to completely fill the config before using it.

This explains most of the diff. The rest is automatic changes caused by updating deps.


bassosimone:
TODO(@bassosimone): backport https://github.com/ooni/probe-cli/pull/134#pullrequestreview-422982498

PR #137, opened 2020-06-08
Expose the STUNReachability experiment

bassosimone:
Closes https://github.com/ooni/probe/issues/1185.


nettests/psiphon.go
```diff
@@ -27,22 +27,22 @@ type PsiphonTestKeys struct {
...
-		err error
-		ok  bool
+		ok       bool
+		testKeys PsiphonTestKeys
 	)
-	testKeys := PsiphonTestKeys{IsAnomaly: false, Failure: ""}
 	if tk["failure"] != nil {
 		testKeys.IsAnomaly = true
-		testKeys.Failure, ok = tk["failure"].(string)
+		failure, ok := tk["failure"].(*string)
```
bassosimone, line 35: I found out that this pattern needed to be changed when testing stunreachability

hellais: Is the same change perhaps needed also for BootstrapTime?

nettests/groups.go
```diff
@@ -39,6 +39,7 @@ var NettestGroups = map[string]NettestGroup{
 	"circumvention": {
 		Label: "Circumvention Tools",
 		Nettests: []Nettest{
+			STUNReachability{},
```
hellais, line 42: Doesn't this mean that we are going to be in fact recording (and displaying) the result for the STUNReachability test in the UI as well?

I thought the plan was to start collecting data for this test, but without presenting it in the UI.

nettests/psiphon.go
bassosimone: No: `BootstrapTime` is not nullable. The issue here is that casting to `*string` and casting to `string` is not the same and we need to use the precise type. As for `BootstrapTime` it's `float64` so we're good.

nettests/groups.go
bassosimone: This runs the test as part of the nettest group. We will show the results when running from the command line, but I don't know whether the desktop app will show this results and in which way. I would assume it will skip results it doesn't know?

bassosimone:
@hellais I think it's fine that CLI users see it. What do desktop user see?

nettests/groups.go
hellais: I believe the desktop app is going to either display an error in the row where this is shown or potentially even crash.

@sarathms perhaps know better the behavior.

I think the easiest way to go about implementing this is to hide completely the output of the test also from probe-cli, that way we don't have to worry about the integration with the desktop app as the output is entirely identical.

bassosimone: I would like to avoid writing hackish code for running this experiment. Ideally the desktop app should only process the measurement it understands. Imma try and run circumvention with a OONI Probe that supports stun_reachability and see what happens.

hellais:
I tested this in the desktop app and it appears that it will in fact ignore the stun test result, but the counter for the number of tools is inconsistent (it says 3 instead of 2):

Screenshot 2020-06-11 at 14 51 20

I think the best way to go about doing this is to omit writing any sort of information about the results of this test to database at all and ensure that the output is identical to the output without this test.

bassosimone:
We'll put this PR on hold until we fix the issue of bad names in the database.

bassosimone:
@hellais I've changed the PR so that we don't run STUNReachability. So, we can get the benefits of the latest probe-engine, but we are not running the test. (Deleting the existing code or opening a new PR looked like not needed to me; I'm open to create a new PR if you consider instead that doing that would be more proper form for proceeding.)

nettests/groups.go
bassosimone: Resolving this conversation: we're not running the experiment until we have a UI.

hellais:
I created a new epic and child issues for this: 
https://github.com/ooni/probe/issues/1202
This involves:

* [Designing UI for STUN test](ooni/probe#1119)
* [Add support for running STUN test in probe-cli](ooni/probe#1200)
* [Adding probe-desktop support for STUN test UI](ooni/probe#1201)

bassosimone:
This branch now just contains the STUNReachability code. The part related to updating the version of ooni/probe-engine and to fixing Psiphon code landed as part of https://github.com/ooni/probe-cli/pull/138

bassosimone:
I have added STUN to the experimental nettest group in https://github.com/ooni/probe-cli/pull/243. This PR is now obsolete.

PR #162, opened 2020-11-13
Debian package: run as a daemon

FedericoCeratto:


debian/ooniprobe-cli.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=OONI CLI Probe
+Requires=ooniprobe-cli.service
+
+[Timer]
+Unit=ooniprobe-cli.service
+OnCalendar=daily
+RandomizedDelaySec=24h
```
hellais, line 8: Are we sure we would like to do 24h intervals? I was thinking it might be better to do it more frequently like every 6 hours so we can cover the day better.

Moreover, do we maybe want to add a random delay somewhere in here?

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,53 @@
...
+ReadOnlyDirectories=/
+ReadWriteDirectories=-/proc
+ReadWriteDirectories=-/var/log/ooniprobe
+ReadWriteDirectories=-/var/lib/ooniprobe
+ReadWriteDirectories=-/var/run
+RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6
+RestrictNamespaces=yes
+RestrictRealtime=yes
+SystemCallArchitectures=native
+SystemCallFilter=@system-service
```
hellais, line 50: Nice!

hellais:
Should it perhaps be called `ooniprobe` like the old package or do we want to avoid conflicts with the two?

debian/ooniprobe-cli.timer
bassosimone: My understanding is that `RandomizedDelaySec` is already random. I have a feeling we should start with a very conservative value here for packages and then see what happens and choose how to proceed?

bassosimone: > My understanding is that RandomizedDelaySec is already random [...]

This is literally my understanding from reading the man page but I have not tested it. Though, I assume that `OnCalendar` plus `RandomizedDelaySec` gives us a single run per day w/ random delay uniform in [0..24h].

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,53 @@
+# OnionBalance systemd target
```
bassosimone, line 1: @FedericoCeratto I would recommend to set `OONI_HOME` to `/var/lib/ooniprobe`. This seems to be the correct way of forcing `ooniprobe` to use a specific directory as home directory. This approach is better than just letting `ooniprobe` choose based on the value of `$HOME` in that we don't create a `.ooniprobe` subfolder.

bassosimone (changes requested):

debian/ooniprobe-cli.timer
FedericoCeratto: According to the logs it's randomized as described.

debian/ooniprobe-cli.service
FedericoCeratto: Done

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,51 @@
+[Unit]
+Description=OONI CLI Probe
```
bassosimone, line 2:
```suggestion
Description=OONI Probe CLI
```

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,51 @@
...
+Documentation=man:ooniprobe-cli
+#Documentation=file:///usr/share/doc/ooniprobe-cli/html/index.html
```
bassosimone, line 4:
```suggestion
```

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,51 @@
...
+Documentation=https://ooni.org/
+After=network.target tor.target
+Wants=network-online.target
+ConditionPathExists=/etc/ooniprobe/ooniprobe.conf
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/ooniprobe --config=/etc/ooniprobe/ooniprobe.conf run all
```
bassosimone, line 12:
```suggestion
ExecStart=/usr/bin/ooniprobe --config=/etc/ooniprobe/ooniprobe.conf run unattended
```

We don't want to run performance tests in background. (The `unattended` command is implemented in 3.1.0)

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,51 @@
...
+Environment="OONI_HOME=/var/lib/ooniprobe"
+User=ooniprobe
+PermissionsStartOnly=true
+Restart=on-abnormal
+RestartSec=2s
+LimitNOFILE=65536
+WorkingDirectory=/var/lib/ooniprobe
+
+# Sandboxing
+CapabilityBoundingSet=CAP_NET_BIND_SERVICE
```
bassosimone, line 25: Could this be the empty set of capabilities? (AFAIK we don't bind any privileged port)

debian/ooniprobe-cli.service
```diff
@@ -0,0 +1,51 @@
...
+ProtectHostname=yes
+ProtectKernelLogs=yes
+ProtectKernelModules=yes
+ProtectKernelTunables=yes
+ProtectSystem=full
+ReadOnlyDirectories=/
+ReadWriteDirectories=-/proc
+ReadWriteDirectories=-/var/log/ooniprobe
+ReadWriteDirectories=-/var/lib/ooniprobe
+ReadWriteDirectories=-/var/run
```
bassosimone, line 43: My systemd-fu is such that I don't understand the `-` here. The [manual to which I have access](https://www.commandlinux.com/man-page/man5/systemd.exec.5.html) reads:

> Paths in ReadOnlyDirectories= and InaccessibleDirectories= may be prefixed with "-", in which case they will be ignored when they do not exist. Note that using this setting will disconnect propagation of mounts from the service to the host (propagation in the opposite direction continues to work). This means that this setting may not be used for services which shall be able to install mount points in the main mount namespace.

I'm clearly missing some knowledge here, please help :-).

debian/ooniprobe-cli.timer
bassosimone: Thanks!

debian/ooniprobe.conf.disabled
```diff
@@ -0,0 +1,17 @@
+{
+  "_": "",
+  "_version": 1,
+  "_informed_consent": true,
+  "sharing": {
+    "include_ip": false,
+    "include_asn": true,
```
bassosimone, line 7:
```suggestion
```

Not needed in 3.1.0.

debian/rules
```diff
@@ -1,10 +1,17 @@
...
 	dh $@
 
+override_dh_auto_clean:
+	rm -f debian/ooniprobe.1
+	dh_auto_clean
+
 override_dh_auto_build:
-	true
+	# The ooniprobe binary is provided
+	./CLI/linux/amd64/ooniprobe --help-man > debian/ooniprobe.1
```
bassosimone, line 14: This is very neat!

bassosimone (changes requested):
Thank you very much for your excellent work in making a Debian package! I am super happy that we can soon proceed with a new package 🥳!

I have a bunch of requests for changes, which are mainly focused around updating the package to the way in which ooniprobe 3.1.0 works.

In the meanwhile, I'll spin up a Vagrant box and test more thoroughly.

🎸

debian/ooniprobe-cli.service
FedericoCeratto: If those directories are missing for any reason, the systemd unit file will not break. E.g. /var/log/ooniprobe is not being created and that's ok.

bassosimone: What confuses me is that the manual page that I linked to explicitly states that this behaviour is only occurring for ReadOnlyDirectories and InaccessibleDirectories, but above it's used for ReadWriteDirectories. It seems the [Debian manual page for that topic](https://manpages.debian.org/stretch/systemd/systemd.exec.5.en.html) says that also ReadWriteDirectories can be used along with the `-` modifier.

debian/ooniprobe-cli.timer
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=OONI CLI Probe
```
bassosimone, line 2:
```suggestion
Description=OONI Probe CLI
```

bassosimone (approved):

bassosimone:
Thanks again @FedericoCeratto! I am going to merge when all the CIs are green. There are future improvements we can work on, including maybe finding a way to unify the stable and the testing builds file, whose diff is currently very limited. This diff being so limited, though, gives me confidence that the stable build is okay. For the records this is the diff:

```diff
% diff -u .github/workflows/linux-debian-packages{,-release}.yml
--- .github/workflows/linux-debian-packages.yml	2020-12-15 12:23:21.000000000 +0100
+++ .github/workflows/linux-debian-packages-release.yml	2020-12-15 12:21:54.000000000 +0100
@@ -1,10 +1,10 @@
-# Build a Debian package and publish on a test/internal archive
+# Build a Debian package only when a relase tag is applied
+# and publish it on the public/release archive
 name: linux-debian-packages
 on:
-  pull_request:
   push:
-    branches:
-      - master
+    tags:
+      - '*'
 jobs:
   build:
     runs-on: "ubuntu-20.04"
@@ -18,11 +18,10 @@
       - run: find . -name ooniprobe -type f -executable
       - run: sudo apt-get update -q
       - run: sudo apt-get build-dep -y --no-install-recommends .
-      # Use <probe version>~<github build number> as package version
+      # Use probe version as package version
       - run: |
-          VER=$(./CLI/linux/amd64/ooniprobe version)
-          DVER="${VER}~${GITHUB_RUN_NUMBER}"
-          dch -v $DVER "test version"
+          DVER=$(./CLI/linux/amd64/ooniprobe version)
+          dch -v $DVER "New release"
           dpkg-buildpackage -us -uc -b
           find ../ -name "*.deb" -type f
           DEB="../ooniprobe-cli_${DVER}_amd64.deb"
@@ -35,4 +34,4 @@
           BT_APIUSER: federicoceratto
           BT_ORG: ooni
           BT_PKGNAME: ooniprobe
-          BT_REPO: ooniprobe-debian-test
+          BT_REPO: ooniprobe-debian
```

PR #177, opened 2020-11-26
feat: use ooni/probe-engine@286613b74e6c and cleanup

bassosimone:
1. zap unused configuration settings from the config file but do not
bump the version number because doing that _may_ interact in unexpected
ways with probe-desktop (hence https://github.com/ooni/probe/issues/1297)
and also because we've just _removed_ stuff for now, therefore any
previous configuration file will continue to work, except that we'll
be ignoring a bunch of options. In a future version of probe-cli I'll
spend some time to further improve config file management.

2. accordingly, make sure all current configuration files that are around
in the tree are current and only feature supported options.

3. update to ooni/probe-engine@286613b74e6c, which contains a bunch of
APIs that should allow us to simplify the interaction between the cli and
the engine, by sharing code more cleverly.

4. zap GetTestKeys because now we use code in probe-engine instead.

5. zap LogSummary because it was not being used.

6. the main change related to cleaning up the config and to the update
to the latest probe-engine is that include_{cc,asn,ip} settings are
gone and we now share the CC and the ASN and we never share the IP addr.

Reference issue: https://github.com/ooni/probe/issues/1283.

After this change is landed, there's a bunch more work to do to further
unify cli and engine. The final state will be that the cli uses ~the code
used by miniooni, so it will have a bunch of desirable options.


bassosimone:
It seems I need to update `bindata` before merging. Will do that soon.

PR #179, opened 2020-11-30
refactor(webconnectivity): use engine.InputLoader

bassosimone:
This diff pins to ooni/probe-engine@3049779878bf6858def6514119b392c19f09d985
and starts using the recently introduced probe-engine APIs.

Namely, here, we use the InputLoader for loading URLs.

I've confirmed manually everything is still working as intended.

Part of https://github.com/ooni/probe/issues/1283.

(In particular, the InputLoader is the abstraction allowing us to load
input from several sources, including command line flags and external
files.)


bassosimone:
Merging: the macos build has been building for more than twenty minutes now without any output on the console. I stopped it and restarted it and again there was no output on the console. I'm going to flag this as a GHA glitch.

PR #181, opened 2020-12-01
feat: implement syslog logging

bassosimone:
With this functionality in tree, macOS users could easily access ooniprobe logs by filtering by process name in the console app. See https://github.com/ooni/probe-cli/pull/181#issuecomment-736508139 for a nice screenshot.

Part of https://github.com/ooni/probe/issues/1289

### DIFF
--- a/cmd/ooniprobe/main.go
+++ b/cmd/ooniprobe/main.go
@@ -1,9 +1,8 @@
 package main
 
 import (
-	// commands
-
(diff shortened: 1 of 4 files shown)


bassosimone:
The following screenshot shows how messages appear on macOS:

Screenshot 2020-12-01 at 12 59 53

The corresponding command was: `./ooniprobe --log-handler=syslog run im`.

PR #184, opened 2020-12-04
go get -v github.com/ooni/probe-engine@f8284475147

bassosimone:


bassosimone:
It seems macOS build is stuck. Let's merge!

PR #187, opened 2021-01-07
Add support for tracking the informed consent via a file-based flag

hellais:
Add support for migrating informed consent from probe-legacy into
probe-cli

### DIFF
--- a/internal/ooni/ooni.go
+++ b/internal/ooni/ooni.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"sync/atomic"
 	"syscall"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/ooni/probe-cli/internal/database"
 	"github.com/ooni/probe-cli/internal/enginex"
 	"github.com/ooni/probe-cli/internal/utils"
+	"github.com/ooni/probe-cli/internal/utils/homedir"
 	engine "github.com/ooni/probe-engine"
 	"github.com/pkg/errors"
 	"upper.io/db.v3/lib/sqlbuilder"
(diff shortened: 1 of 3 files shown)


internal/ooni/ooni.go
```diff
@@ -239,6 +252,23 @@ func MaybeInitializeHome(home string) error {
...
+// MaybeMigrateLegacyConsent takes the informed consent from a legacy version of
+// OONI Probe and migrates it to the new version
+func MaybeMigrateLegacyConsent(ooniHome string) error {
+	userHome, err := homedir.Dir()
+	if err != nil {
+		return err
+	}
+	path := filepath.Join(filepath.Join(userHome, ".ooni"), "initialized")
+	if utils.FileExists(path) {
+		_, err := os.Create(utils.InformedConsentPath(ooniHome))
```
hellais, line 264: I should use the function to Accept the consent in here.

hellais: Actually it's maybe not so much of a good idea since all the functions in this file take as argument the `ooniHome` so it's unexpected for this one to not behave in that way. I think leaving it as-is, since it's anyways legacy compatibility code that will eventually be removed, is safer.

internal/utils/paths.go
```diff
@@ -41,6 +46,13 @@ func DBDir(home string, name string) string {
 	return filepath.Join(home, "db", fmt.Sprintf("%s.sqlite3", name))
 }
 
+func FileExists(path string) bool {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return false
+	}
+	return true
```
bassosimone, line 53:
```suggestion
	stat, err := os.Stat(path)
	return err == nil && stat.Mode().IsRegular()
```

I think we should treat the file as missing if there is _any_ error. For example, if the directory is not readable, we want to return false here. Also, we want to make sure the target path is a regular file.

internal/ooni/ooni.go
bassosimone: sure!

hellais:
This PR is superseded by https://github.com/ooni/probe-cli/pull/188

PR #190, opened 2021-01-12
chore: update deps (probe-engine 0.22.0 et al.)

bassosimone:
The CloseReport method is gone. We don't need to close reports
anymore with the new OONI backend.

The InputsRequired flags now is InputsOrQueryTestLists.


bassosimone:
See https://github.com/ooni/probe-engine/issues/1125

PR #192, opened 2021-01-13
feat: implement darwin launch agent

bassosimone:
Part of https://github.com/ooni/probe/issues/1289

### DIFF
--- a/internal/autorun/autorun_darwin.go
+++ b/internal/autorun/autorun_darwin.go
@@ -1,6 +1,4 @@
-// +build darwin
-
-package periodic
+package autorun
 
 import (
 	"bytes"
@@ -69,7 +67,7 @@ func (managerDarwin) mustNotHavePlist() error {
 	log.Infof("exec: test -f %s && already_registered()", plistPath)
 	if utils.FileExists(plistPath) {
 		// This is not atomic. Do we need atomicity here?
-		return errors.New("periodic: service already registered")
+		return errors.New("autorun: service already registered")
 	}
 	return nil
 }
@@ -96,7 +94,6 @@ func (managerDarwin) start() error {
 	return runQuiteQuietly("launchctl", "bootstrap", domainTarget, plistPath)
 }
 
-// Start starts running periodically.
 func (m managerDarwin) Start() error {
 	operations := []func() error{m.mustNotHavePlist, m.writePlist, m.start}
 	for _, op := range operations {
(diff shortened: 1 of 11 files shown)


internal/cli/run/run.go
```diff
@@ -74,6 +76,13 @@ func init() {
 
 	unattendedCmd := cmd.Command("unattended", "")
 	unattendedCmd.Action(func(_ *kingpin.ParseContext) error {
+		if runtime.GOOS == "darwin" {
+			// Until we have enabled the check-in API we're called every
+			// hour on darwin and we need to self throttle.
+			// TODO(bassosimone): switch to check-in and remove this hack.
+			const veryFew = 10
```
hellais, line 83: @FedericoCeratto do you think this is a reasonable number? Should we bump it to something higher?

hellais:
I took a look at the code and it looks good to me! I only left a small comment for something I think we should discuss with the rest of the team. I am now going to do a bit of testing of it locally.

hellais:
I made the suggestion to rename the command name from `periodic` to `autorun`

hellais:
When I run `log show` I get the following error:
```
./CLI/darwin/amd64/ooniprobe periodic log show
   • exec: log show --info --debug --process ooniprobe --style compact
log: unrecognized option `--process'
```

Running `log stream` works OTOH.

internal/cli/run/run.go
bassosimone: He says ten is reasonable. We could also possibly use twenty. Because we will have check-in in 2-3 weeks, I think it's fine here to be conservative and just keep ten, which seems "low" to everyone. What do you think?

hellais: Ok let's air on the safe side and keep it to 10. We can always make a new release. As we have seen homebrew users do update fairly regularly.

internal/autorun/autorun_darwin.go
```diff
@@ -0,0 +1,178 @@
...
+	return major, nil
+}
+
+var errNotImplemented = errors.New(
+	"autorun: command not implemented in this version of macOS")
+
+func (managerDarwin) LogShow() error {
+	major, _ := darwinVersionMajor()
+	if major < 20 /* macOS 11.0 Big Sur */ {
+		return errNotImplemented
```
hellais, line 79: 👍

hellais (approved):
🐳

bassosimone:
I've started testing this functionality extensively. To do the same:

```
cd src/github.com/ooni/probe-cli
git checkout master
git pull
go install -v ./cmd/ooniprobe
~/go/bin/ooniprobe autorun start
```

PR #193, opened 2021-01-18
feat(build.sh): build windows/386 and linux/386

bassosimone:
See https://github.com/ooni/probe/issues/1315


bassosimone:
We still need to figure out some bits. Stopping for today, but we're close to landing this branch!

PR #197, opened 2021-01-22
Fixup the rightpad on the list command

hellais:
before
Screenshot 2021-01-22 at 12 47 50
after
Screenshot 2021-01-22 at 12 47 30


bassosimone (approved):
🐳

PR #199, opened 2021-02-01
Test armhf build

FedericoCeratto:


.github/workflows/linux.yml
```diff
@@ -22,3 +22,72 @@ jobs:
...
+          # Not required, but speeds up builds by storing container images in
+          # a GitHub package registry.
+          githubToken: ${{ github.token }}
+
+          run: |
+            uname -a
+            apt-get -qq update
+            apt-get install -qq -y --no-install-recommends gcc git linux-headers-armmp wget musl-dev ca-certificates libc6-dev
+            wget --no-verbose https://golang.org/dl/go1.16rc1.linux-armv6l.tar.gz
+            tar xfz go1.16rc1.linux-armv6l.tar.gz
```
bassosimone, line 48: Would this run on alpine? I'd rather use Alpine and their latest `go` version that pinning to a specific version. Also, by building on Alpine we can truly static link and hence make a binary that works on any distro.

bassosimone (commented):
Thank you! I have a question regarding whether this would work on Alpine Linux.

.github/workflows/linux.yml
```diff
@@ -22,3 +22,72 @@ jobs:
...
+            if [[ ! $GITHUB_REF =~ ^refs/tags/* ]]; then
+              VER="${VER}~${GITHUB_RUN_NUMBER}"
+              dch -v $VER "New test version"
+              BT_REPO="${BT_REPO}-test"
+            else
+              dch -v $VER "New release"
+            fi
+            dpkg-buildpackage -us -uc -b
+            find ../ -name "*.deb" -type f
+            DEB="../ooniprobe-cli_${VER}_armhf.deb"
```
bassosimone, line 83:
```suggestion
            DEB="../ooniprobe-cli_${VER}_armhf.deb"
            echo no | sudo dpkg -i $DEB
```

If someone edits the templates file and makes a mistake with the stanzas, the debian package will build but it will fail to install. This is not optimal. We should instead be failing the build here. The best way to do so, in my view, is to install the package and make sure the installation works.

.github/workflows/linux.yml
```diff
@@ -22,3 +22,72 @@ jobs:
         with:
           name: ooniprobe-linux-amd64
           path: ./CLI/linux/amd64/ooniprobe
+
```
bassosimone, line 25: I would be grateful if you could move this code to the `debian.yml` workflow :-). This is the workflow that we use for building for Debian, and I find it surprising that we build for Debian here. (I know we have too many workflows and so it's not immediately easy to understand which one to use, sorry!)

.github/workflows/linux.yml
```diff
@@ -22,3 +22,72 @@ jobs:
...
+    runs-on: "ubuntu-latest"
+    name: Build on armhf Buster
+    steps:
+      - uses: actions/checkout@v2.1.0
+      - uses: uraimo/run-on-arch-action@v2.0.5
+        name: Build ooniprobe
+        id: runcmd
+        with:
+          arch: armv7
+          distro: buster
```
bassosimone, line 36: Why not building on Alpine? The `alpine_latest` target supports all the arm platforms we care about. By doing that, we produce a statically linked executable that has no `libc` issues whatsoever. (Alternatively, see my suggestion below, but I would be happier with running this code inside Alpine.)

.github/workflows/linux.yml
```diff
@@ -22,3 +22,72 @@ jobs:
...
+          # distro: ubuntu20.04 # no linux-headers-arm64
+
+          # Not required, but speeds up builds by storing container images in
+          # a GitHub package registry.
+          githubToken: ${{ github.token }}
+
+          run: |
+            uname -a
+            apt-get -qq update
+            apt-get install -qq -y --no-install-recommends gcc git linux-headers-armmp wget musl-dev ca-certificates libc6-dev
```
bassosimone, line 46: Why installing `musl-dev` here? I admit not knowing much about what `musl-dev` could do on Debian, but I suspect that, if we want the compiler to link against it, then we need other compiler flags. (Am I missing something?)

.github/workflows/linux.yml
```diff
@@ -22,3 +22,72 @@ jobs:
...
+            wget --no-verbose https://golang.org/dl/go1.16rc1.linux-armv6l.tar.gz
+            tar xfz go1.16rc1.linux-armv6l.tar.gz
+            export PATH=$PATH:$(pwd)/go/bin
+            go version
+            go env
+            go build -tags netgo -ldflags='-s -w -extldflags -static' ./cmd/ooniprobe
```
bassosimone, line 52: If you don't want to switch to Alpine, this is fine but, then, please omit `-static`. Statically linking with `glibc` is not a good idea because `glibc` insists on loading some modules dynamically. So, this is dangerous from the point of view of portability and can cause unexpected issues when running on different systems.

.github/workflows/linux.yml
bassosimone: I am going to resolve this conversation, since I have expanded what I mean in greater detail elsewhere.

bassosimone (commented):
Good day, @FedericoCeratto! Since we spoke yesterday, I took some more time looking harder into this PR and noticed that a few small tweaks could give us what we want (i.e. statically linking using Alpine or not-statically linking using Debian.)

I am very excited to see this code landing in `master` soon! I know it has been a lot of research and grunt work to make this build happen because I have seen it happening. I am grateful that you took on this important task :-). This :beers: is a voucher with which you can reclaim a couple of beers with me when the pandemic is over and we can travel again :crossed_fingers:

PR #202, opened 2021-02-02
fix: add missing files causing tests to fail

bassosimone:
See https://github.com/ooni/probe/issues/1335#issuecomment-771559718


bassosimone:
Okay, we can move forward. The failures we see here are ~same as the ones we see in probe-engine. See https://github.com/ooni/probe-cli/runs/1814490513. Basically, "KibiBytes sent" (https://github.com/ooni/probe-engine/issues/1120) and RiseupVPN (which is having some random/transient issue).

PR #204, opened 2021-02-02
refactor: integrate more probe-engine workflows

bassosimone:
This diff is part of https://github.com/ooni/probe/issues/1335.

We are moving more probe-engine workflows to toplevel.

The general idea here is to migrate all possible workflows and to
delete the ones that we cannot use in this repo (if any).


bassosimone:
We know that there are test failures. They are the usual suspects. At this stage, I would like to move forward and continue integrating ooni/probe-engine into ooni/probe-cli. Later today I will deal with the errors by skipping the relevant tests.

PR #205, opened 2021-02-02
refactor: start building an Android package

bassosimone:

This seems also a good moment to move some packages out of the
engine, e.g., oonimkall. This package, for example, is a consumer
of the engine, so it makes sense it's not _inside_ it.


bassosimone:
We are now able to build the Android AAR from the `probe-cli` repository. We still see two failures: (1) "unexpected KibiBytesSent" and (2) RiseupVPN. Those failures are ~okay at this stage since they also occurred for the probe-engine repo.

PR #207, opened 2021-02-03
refactor: move more commands to internal/cmd

bassosimone:

We would like all commands to be at the same level of engine
rather than inside engine (now that we can do it).


bassosimone:
Tests do not fix themselves of course. But I have some more refactoring to do before tending to that.

PR #209, opened 2021-02-03
refactor: enable automatic iOS builds

bassosimone:
Part of https://github.com/ooni/probe/issues/1335


bassosimone:
Okay, we can build and publish an iOS package! More progress!

PR #212, opened 2021-02-03
fix(riseupvpn): skip currently failing tests

bassosimone:

Tracked also by https://github.com/ooni/probe/issues/1338.


bassosimone:
Good news: we're green again!

PR #221, opened 2021-02-04
oonimkall: add wrappers for /test-list/urls

lorenzoPrimi:
Implementing https://github.com/ooni/probe/issues/1331

### DIFF
--- a/pkg/oonimkall/session.go
+++ b/pkg/oonimkall/session.go
@@ -5,9 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net/url"
 	"runtime"
-	"strings"
 	"sync"
 
 	engine "github.com/ooni/probe-cli/v3/internal/engine"
@@ -435,22 +433,11 @@ func (sess *Session) FetchURLList(ctx *Context, config *URLListConfig) (*URLList
 		Limit:       config.Limit,
 	}
 
-	query := url.Values{}
-	if cfg.CountryCode != "" {
-		query.Set("country_code", cfg.CountryCode)
-	}
-	if cfg.Limit > 0 {
-		query.Set("limit", fmt.Sprintf("%d", cfg.Limit))
-	}
-	if len(cfg.Categories) > 0 {
-		query.Set("category_codes", strings.Join(cfg.Categories, ","))
-	}
-	var response URLListResult
-	err = psc.GetJSONWithQuery(ctx.ctx, "/api/v1/test-list/urls", query, &response)
+	result, err := psc.FetchURLList(ctx.ctx, cfg)
 	if err != nil {
 		return nil, err
 	}
 	return &URLListResult{
-		Results: response.Results,
+		Results: result,
 	}, nil
 }
--- a/pkg/oonimkall/session.go
+++ b/pkg/oonimkall/session.go
@@ -385,6 +385,11 @@ type URLListConfig struct {
 	Limit       int64    // Max number of URLs (<= 0 means no limit)
 }
 
+// URLListResult contains the URLs returnet from the FetchURL API
+type URLListResult struct {
+	Results []model.URLInfo
+}
+
 // AddCategory adds category code to the array in URLListConfig
 func (ckw *URLListConfig) AddCategory(cat string) {
 	ckw.Categories = append(ckw.Categories, cat)
@@ -408,13 +413,7 @@ func (ckw *URLListResult) Size() int64 {
 	return int64(len(ckw.Results))
 }
 
-// URLListResult contains the URLs returnet from the FetchURL API
-type URLListResult struct {
-	Results []model.URLInfo
-}
-
 // FetchURLList function is called to fetch the urls list to test
-// In the config argument a probe can specify the Categories to test, limit urls and send the Probe's CC
 func (sess *Session) FetchURLList(ctx *Context, config *URLListConfig) (*URLListResult, error) {
 	sess.mtx.Lock()
 	defer sess.mtx.Unlock()
--- a/pkg/oonimkall/session.go
+++ b/pkg/oonimkall/session.go
@@ -413,7 +413,7 @@ func (ckw *URLListResult) Size() int64 {
 	return int64(len(ckw.Results))
 }
 
-// FetchURLList function is called to fetch the urls list to test
+// FetchURLList fetches the list of URLs to test
 func (sess *Session) FetchURLList(ctx *Context, config *URLListConfig) (*URLListResult, error) {
 	sess.mtx.Lock()
 	defer sess.mtx.Unlock()
--- a/pkg/oonimkall/session.go
+++ b/pkg/oonimkall/session.go
@@ -380,8 +380,9 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
 
 // URLListConfig contains configuration for fetching the URL list.
 type URLListConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/oonimkall/session.go
```diff
@@ -375,3 +377,81 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
...
+
+// Size returns the number of URLs.
+func (ckw *URLListResult) Size() int64 {
+	return int64(len(ckw.Results))
+}
+
+// URLListResult contains the URLs returnet from the FetchURL API
+type URLListResult struct {
+	Results []model.URLInfo
+}
```
bassosimone, line 414: Please, move the struct definition before its methods

pkg/oonimkall/session.go
```diff
@@ -375,3 +377,81 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
...
+
+// FetchURLList function is called to fetch the urls list to test
```
bassosimone, line 416:
```suggestion
// FetchURLList fetches the list of URLs to test
```

pkg/oonimkall/session.go
```diff
@@ -375,3 +377,81 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
...
+// In the config argument a probe can specify the Categories to test, limit urls and send the Probe's CC
```
bassosimone, line 417: I don't think there is need to describe the config argument here.

pkg/oonimkall/session.go
```diff
@@ -375,3 +377,81 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
...
+		query.Set("country_code", cfg.CountryCode)
+	}
+	if cfg.Limit > 0 {
+		query.Set("limit", fmt.Sprintf("%d", cfg.Limit))
+	}
+	if len(cfg.Categories) > 0 {
+		query.Set("category_codes", strings.Join(cfg.Categories, ","))
+	}
+	var response URLListResult
+	err = psc.GetJSONWithQuery(ctx.ctx, "/api/v1/test-list/urls", query, &response)
```
bassosimone, line 450: This code does not strike me as wrong, but there should be already code in `probeservices` for doing that. I'm ~sure you are duplicating existing code here. Can you perhaps could check?

pkg/oonimkall/session_integration_test.go
```diff
@@ -419,6 +419,67 @@ func TestCheckInNoParams(t *testing.T) {
 	}
 }
 
+func TestFetchURLListSucc(t *testing.T) {
```
bassosimone, line 422:
```suggestion
func TestFetchURLListSuccess(t *testing.T) {
```

pkg/oonimkall/session_integration_test.go
```diff
@@ -419,6 +419,67 @@ func TestCheckInNoParams(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	ctx := sess.NewContext()
+	config := oonimkall.URLListConfig{
+		Limit: 10,
+	}
+	config.AddCategory("NEWS")
+	config.AddCategory("CULTR")
+	result, err := sess.FetchURLList(ctx, &config)
+
```
bassosimone, line 434:
```suggestion
```

pkg/oonimkall/session_integration_test.go
```diff
@@ -419,6 +419,67 @@ func TestCheckInNoParams(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error: %+v", err)
+	}
+	if result == nil || result.Results == nil {
+		t.Fatal("got nil result")
+	}
+	if len(result.Results) < 1 {
```
bassosimone, line 441: `len(result.Results)` is zero when `result.Results` is `nil`, so you can write more compact code above.

pkg/oonimkall/session_integration_test.go
```diff
@@ -419,6 +419,67 @@ func TestCheckInNoParams(t *testing.T) {
...
+	config.AddCategory("NEWS")
+	config.AddCategory("CULTR")
+	result, err := sess.FetchURLList(ctx, &config)
+	if err != nil {
+		t.Fatalf("unexpected error: %+v", err)
+	}
+	if result == nil || result.Results == nil {
+		t.Fatal("got nil result")
+	}
+	if len(result.Results) < 1 {
```
bassosimone, line 470: Same comment as before

bassosimone (changes requested):
Thanks! I feel like we should be reusing more code in `probeservices` here!

pkg/oonimkall/session.go
lorenzoPrimi: Yes there is similar code in probeservices/urls.go, but I thought we needed to duplicate the code we need for the wrappers

pkg/oonimkall/session_integration_test.go
lorenzoPrimi: Is there a way to differentiate between nil and empty array? I don't know how you want the code to behave

bassosimone: In this case an empty array or a nil array is wrong, so I don't think we need to distinguish.

bassosimone (approved):
🐳 

The code looks good to me! I am going to test locally and merge later today!

Thank you!

pkg/oonimkall/session.go
```diff
@@ -375,3 +375,69 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
...
 }
+
+// URLListConfig contains configuration for fetching the URL list.
+type URLListConfig struct {
+	Categories  []string // Categories to query for (empty means all)
+	CountryCode string   // CountryCode is the optional country code
+	Limit       int64    // Max number of URLs (<= 0 means no limit)
+}
+
+// URLListResult contains the URLs returnet from the FetchURL API
```
bassosimone, line 386:
```suggestion
// URLListResult contains the URLs returned from the FetchURL API
```

pkg/oonimkall/session.go
```diff
@@ -375,3 +375,69 @@ func (sess *Session) CheckIn(ctx *Context, config *CheckInConfig) (*CheckInInfo,
...
+type URLListResult struct {
+	Results []model.URLInfo
+}
+
+// AddCategory adds category code to the array in URLListConfig
+func (ckw *URLListConfig) AddCategory(cat string) {
+	ckw.Categories = append(ckw.Categories, cat)
+}
+
+// At gets the URLInfo at position idx from CheckInInfoWebConnectivity.URLs
```
bassosimone, line 396:
```suggestion
// At gets the URLInfo at position idx from CheckInInfoWebConnectivity.URLs. It returns
// nil if you are using an outs of bound index.
```

pkg/oonimkall/session_integration_test.go
```diff
@@ -419,6 +419,59 @@ func TestCheckInNoParams(t *testing.T) {
...
+	found := false
+	for _, entry := range result.Results {
+		if entry.CountryCode == "IT" {
+			found = true
+		}
+	}
+	if !found {
+		t.Fatalf("not found url for country code: IT")
+	}
+}
```
bassosimone, line 474:
```suggestion
}

```

bassosimone:
I am going to merge this PR as soon as the automatic tests succeed.

PR #226, opened 2021-02-11
fix(http body saver): properly deal with EOF terminated bodies

bassosimone:
See https://github.com/ooni/probe-engine/issues/1191

### DIFF
--- a/internal/engine/netx/httptransport/saver.go
+++ b/internal/engine/netx/httptransport/saver.go
@@ -2,6 +2,7 @@ package httptransport
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


hellais (approved):
LGTM!

internal/engine/netx/httptransport/saver.go
```diff
@@ -139,6 +141,22 @@ func (txp SaverBodyHTTPTransport) RoundTrip(req *http.Request) (*http.Response,
...
+// ignoreExpectedEOF converts an error signalling the end of the body
+// into a success. We know that we are in such condition when the
+// resp.Close hint flag is set to true. (Thanks, stdlib!)
+//
+// See https://github.com/ooni/probe-engine/issues/1191 for an analysis
+// of how this error was impacting measurements and data quality.
+func ignoreExpectedEOF(err error, resp *http.Response) error {
+	if err == nil {
+		return nil
+	}
```
hellais, line 153: Isn't this redundant?

hellais: If the err is `nil` you should be able to return it as `nil` nonetheless, no?

bassosimone: yes, I prefer to have it explicit in the code

PR #228, opened 2021-02-12
Disable sentry based crash reporting in probe-cli

hellais:
The existing crash reports are very few and not actionable. Better off just disabling them altogether since we are no longer in beta testing stage.


bassosimone (approved):
🐳

PR #231, opened 2021-02-16
Debian package: measurement deletion

FedericoCeratto:
Also set version 3.5.2 ; few other changes

### DIFF
--- a/debian/ooniprobe-cli.service
+++ b/debian/ooniprobe-cli.service
@@ -9,6 +9,7 @@ ConditionPathExists=/etc/ooniprobe/ooniprobe.conf
 
 [Service]
 Type=simple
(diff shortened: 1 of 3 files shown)


debian/ooniprobe-cli.service
```diff
@@ -9,6 +9,7 @@ ConditionPathExists=/etc/ooniprobe/ooniprobe.conf
 
 [Service]
 Type=simple
+ExecStartPre=-/bin/rm -rf /var/lib/ooniprobe/msmts/*
```
hellais, line 12: There is no longer the need to do this as it's done automatically by probe-cli since version 3.2.0

FedericoCeratto:
Dropping the PR as discussed.

PR #234, opened 2021-02-23
engine/ooapi: autogenerated API with login and caching

bassosimone:


hellais (approved):
We discussed these changes with @bassosimone and it seems like a great approach! 🐳

PR #235, opened 2021-02-26
feat: use go1.16 and resources embedding

bassosimone:
We want to embed everything that can be easily embedded. We should, at a
minimum, replace the downloading of resources and bindata.

Ref: https://github.com/ooni/probe/issues/1367.


internal/cmd/getresources/getresources.go
```diff
@@ -0,0 +1,51 @@
...
+
+func getit(name string, ri *resources.ResourceInfo) error {
+	workDir := filepath.Join("internal", "engine", "resourcesmanager")
+	URL, err := url.Parse(resources.BaseURL)
+	if err != nil {
+		return err
+	}
+	URL.Path = ri.URLPath
+	log.Println("fetching", URL.String())
+	resp, err := http.DefaultClient.Get(URL.String())
```
bassosimone, line 33:
```suggestion
	resp, err := http.Get(URL.String())
```

PR #241, opened 2021-03-08
fix(webconnectivity): allow measuring https://1.1.1.1

bassosimone:
See https://github.com/ooni/probe/issues/1376


internal/engine/experiment/webconnectivity/dnslookup.go
```diff
@@ -37,6 +37,7 @@ func DNSLookup(ctx context.Context, config DNSLookupConfig) (out DNSLookupResult
 			}
 			if answer.IPv6 != "" {
 				out.Addrs[answer.IPv6] = answer.ASN
+				continue
```
bassosimone, line 40: No functional change, but having `continue` here is more consistent

PR #244, opened 2021-03-08
chore: release 3.7.0

bassosimone:
This comes just a few days after 3.6.0. It contains small
improvements required by ooni/probe-desktop.

For this reason, I am going to skeep the normal release
process and I am just bumping the version number.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -3,5 +3,5 @@ package version
 
 const (
 	// Version is the software version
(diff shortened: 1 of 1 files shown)


bassosimone:
The only failure I see here is caused by the API being out of sync with the probe. It should be fixed with the next deployment of the API, whenever that happens (cc: @FedericoCeratto). I am going to proceed with merging.

PR #248, opened 2021-03-09
feat(ooapi): add toplevel client and simplify API

bassosimone:
This diff should simplify using ooapi from other packages by
adding more abstraction that wraps the existing code.

Part of https://github.com/ooni/probe/issues/1355 and https://github.com/ooni/probe/issues/1405.


internal/engine/ooapi/cloners.go
```diff
@@ -1,18 +1,18 @@
...
 
 package ooapi
 
 //go:generate go run ./internal/generator -file cloners.go
 
-// PsiphonConfigCaller represents any type exposing a method
-// like PsiphonConfigAPI.WithToken.
-type PsiphonConfigCloner interface {
-	WithToken(token string) PsiphonConfigCaller
+// callerForPsiphonConfigAPI represents any type exposing a method
```
bassosimone, line 8: We need to fix this comment

internal/engine/ooapi/integration_test.go
```diff
@@ -41,7 +41,7 @@ func TestWithRealServerDoCheckIn(t *testing.T) {
 		},
 	}
 	httpClnt := &VerboseHTTPClient{t: t}
-	api := &CheckInAPI{
```
bassosimone, line 44: We should rewrite this file to use the new public API

internal/engine/ooapi/internal/generator/clientcall.go
```diff
@@ -0,0 +1,104 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+// TODO(bassosimone): update package docs
```
bassosimone, line 9: We should address this TODO

internal/engine/ooapi/internal/generator/clientcall.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (d *Descriptor) clientMakeAPIBase(sb *strings.Builder) {
+	// TODO(bassosimone): this can probably be automated
```
bassosimone, line 12: We should address this TODO

internal/engine/ooapi/clientcall_test.go
```diff
@@ -0,0 +1,898 @@
...
+
+func (h *handleClientCallTestHelpers) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ff := fakeFill{}
+	defer h.mu.Unlock()
+	h.mu.Lock()
+	if h.count > 0 {
+		w.WriteHeader(400)
+		return
+	}
+	h.count++
```
bassosimone, line 320: TODO(@bassosimone): This should be atomic (not a big deal in this test but better to design correct types). There is similar code when testing the simple APIs, so we should also change that code.

bassosimone: Ah, no, we already have a mutex.

bassosimone:
I am going to merge this one into master! If there are post-merge comments, I'll go and create high-pri issues for them!

PR #249, opened 2021-03-09
feat(oobackend): dynamically change the backend we use

bassosimone:
Part of https://github.com/ooni/probe/issues/1355

### DIFF
--- a/internal/engine/internal/oobackend/client.go
+++ b/internal/engine/internal/oobackend/client.go
@@ -71,10 +71,15 @@ func (c *Client) do(req *http.Request) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
-	sgs := c.readstrategies(time.Now().UnixNano())
+	state := c.readstatemaybedefault(time.Now().UnixNano())
 	me := multierror.New(ErrBackend)
-	defer c.writestrategies(sgs) // update the datastore
-	for _, sg := range sgs {
+	defer c.writestate(state) // update the datastore
+	for _, si := range state.Strategies {
+		sg, err := c.makestrategy(si)
+		if err != nil {
+			me.Add(err)
+			continue
+		}
 		// Note: we pass to sg.Do a copy of the request so it's free
 		// to apply any required transformations. Because sending a
 		// request consumes the request body, we need to recreate the
--- a/internal/engine/internal/oobackend/client.go
+++ b/internal/engine/internal/oobackend/client.go
@@ -1,0 +1,107 @@
+package oobackend
+
+import (
+	"bytes"
+	"errors"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/ooni/probe-cli/v3/internal/engine/internal/multierror"
+)
+
+// TODO(bassosimone): apply similar changes to sessionresolver. In
+// particular, let's avoid writing state in sessionresolver.Resolver
+// by making the KVStore mandatory, which is a small change.
+
+// Client is a client to speak to the OONI backend. In its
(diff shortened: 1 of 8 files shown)


internal/engine/internal/oobackend/client.go
```diff
@@ -0,0 +1,112 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"github.com/ooni/probe-cli/v3/internal/engine/internal/multierror"
+)
+
+// TODO(bassosimone): apply similar changes to sessionresolver. In
+// particular, let's avoid writing state in sessionresolver.Resolver
+// by making the KVStore mandatory, which is a small change.
```
bassosimone, line 15: We need to address this TODO before merging.

internal/engine/internal/oobackend/state.go
```diff
@@ -0,0 +1,164 @@
...
+	var ss strategyState
+	if err := c.jsonCodec().Decode(data, &ss); err != nil {
+		return nil, err
+	}
+	return &ss, nil
+}
+
+// errNoEntries indicates that no entry remained after we pruned
+// all the available entries in readstateandprune.
+var errNoEntries = errors.New("sessionresolver: no available entries")
```
bassosimone, line 64: This error message is wrong

internal/engine/internal/oobackend/client.go
```diff
@@ -0,0 +1,112 @@
...
+	// the state and in which we write the state.
+	KVStore KVStore
+
+	// HTTPClientWithProxy is an optional HTTP client that is
+	// configured to always use a proxy. When this client is
+	// present, we always and only use it, and we do not update
+	// our strategy dynamically. In fact, when this client is
+	// present, it means the user has requested the probe to
+	// use a specific proxy explicitly.
+	HTTPClientWithProxy HTTPClient
```
bassosimone, line 32: I am wondering whether a better design would be to have a boolean indicating we're using a proxy. Another possible design here could be that we don't use this client in case we are using a proxy, but we use the real HTTP client directly.

internal/engine/internal/oobackend/client.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// HTTPClientDefault is the HTTP client we use by default
+	// when HTTPClientWithProxy is not set. If HTTPClientDefault
+	// is not set, we use http.DefaultClient.
+	HTTPClientDefault HTTPClient
```
bassosimone, line 37: We probably need another HTTP client for http3 here. (Another possibility is that we modify the design of the HTTP client to be able to dynamically change what protocol to use with a context, as opposed to creating a new client for each specific need. This means that we probably need to use a different type of client here?)

bassosimone: Also, in this moment we're currently vulnerable to deadlock after the round trip, because reading the body is not bound to the context life cycle. This may be a reason to use a different HTTP client type than the one used by the stdlib?

internal/engine/internal/oobackend/client.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		return resp, err
+	}
+	return nil, me
+}
+
+// readbody reads the request body, if any.
+func (c *Client) readbody(req *http.Request) ([]byte, error) {
+	if req.Body == nil {
+		return nil, nil // nil is an empty []byte
```
bassosimone, line 100: This comment is not very clear

DecFox:
I am going to close this since we already have a series of backends in the engine currently we use as failovers and the backends in this diff no longer exist for us.

PR #256, opened 2021-03-15
chore: bless 3.8.0

bassosimone:


bassosimone:
I've created an issue for the only failing test: https://github.com/ooni/probe/issues/1409. Will merge now, because the failure seems to depend on the interaction between different tests.

PR #260, opened 2021-03-22
refactor: redesign how we import assets

bassosimone:
Here we're fixing the following issues:

1. that `pkg.go.dev` does not build and `go get` is broken because `go run ./internal/cmd/getresources` is required (https://github.com/ooni/probe/issues/1421)

2. that there is no need for us to copy resources https://github.com/ooni/probe/issues/1372 

Thanks to @hellais for this very neat idea!


bassosimone:
# TODO

- [ ] update the PR description to mention all the fixed issues
- [ ] search for more opportunity for simplification

bassosimone (commented):
I did a first pass at looking at the diff and flagging extra work to do. A second pass is still needed, because there are probably extra opportunities to simplify the codebase.

cmd/ooniprobe/internal/ooni/ooni.go
```diff
@@ -199,7 +201,6 @@ func (p *Probe) NewSession() (*engine.Session, error) {
 		return nil, errors.Wrap(err, "creating engine's kvstore")
 	}
 	return engine.NewSession(engine.SessionConfig{
-		AssetsDir:       utils.AssetsDir(p.home),
```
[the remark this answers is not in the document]
bassosimone: No, we cannot because we SHOULD be cleaning up the assets dir.

internal/engine/geolocate/geolocate.go
```diff
@@ -96,30 +89,17 @@ type probeIPLookupper interface {
...
-	// ASNDatabasePath returns the path of the ASN database.
-	ASNDatabasePath() string
-
-	// CountryDatabasePath returns the path of the country database.
-	CountryDatabasePath() string
-
-	// MaybeUpdateResources ensures that the required resources
-	// have been downloaded and are current.
-	MaybeUpdateResources(ctx context.Context) error
-}
```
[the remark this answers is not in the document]
bassosimone: done!

internal/engine/netx/archival/archival.go
```diff
@@ -431,16 +431,16 @@ func NewDNSQueriesList(begin time.Time, events []trace.Event, dbpath string) []D
 func (qtype dnsQueryType) ipoftype(addr string) bool {
 	switch qtype {
 	case "A":
-		return strings.Contains(addr, ":") == false
+		return !strings.Contains(addr, ":")
 	case "AAAA":
-		return strings.Contains(addr, ":") == true
+		return strings.Contains(addr, ":")
```
bassosimone, line 436: These were `staticcheck` warnings that I fixed. While reading again this diff, I've found that the tests for this function could be improved, so I rewrote them in a better way (see below).

internal/engine/netx/archival/archival_internal_test.go
```diff
@@ -0,0 +1,41 @@
+package archival
+
+import "testing"
+
+func TestDNSQueryIPOfType(t *testing.T) {
```
bassosimone, line 5: This is the test I rewrote (see above note).

internal/engine/netx/archival/archival_test.go
```diff
@@ -1009,13 +985,6 @@ func TestNewFailure(t *testing.T) {
 	}
 }
 
-func TestDNSQueryTypeInvalidIPOfType(t *testing.T) {
-	qtype := archival.DNSQueryType("ANTANI")
-	if qtype.IPOfType("8.8.8.8") != false {
-		t.Fatal("unexpected return value")
-	}
-}
-
```
bassosimone, line 1018: The test I rewrote

internal/engine/netx/archival/archival_test_internal.go
```diff
@@ -1,8 +0,0 @@
-package archival
```
bassosimone, line 1: Wrong naming for this file. I renamed it `archival_internal_test.go`.

bassosimone: And then I added there internal tests.

pkg/oonimkall/session.go
```diff
@@ -124,7 +127,6 @@ func NewSession(config *SessionConfig) (*Session, error) {
 		})
 	}
 	engineConfig := engine.SessionConfig{
-		AssetsDir:              config.AssetsDir,
```
bassosimone, line 127: Maybe we should continue propagating `AssetsDir` and cleanup the assets dir in the session?

bassosimone (commented):
I did another pass. The only remaining aspect to address seems whether we want to remove the content of the assets directory (or just the two large files that are there). Everything else looks good to me after this second pass. I could not find any other place where to cleanup.

bassosimone:
On second though there are actually three unaddressed issues:

- [ ] we MUST store somewhere stable the resources. We're currently publishing branches with the resources but there is no guarantee that we will keep these branches forever. We already have a place in which we store the asn.mmdb, but we are missing a place where to store db-ip.com's database, which probably we want to do.

- [ ] if we store resources at a stable place, then we _also_ need to address the problem of allowing someone in the future to reconstruct the resources that were used for a specific release _after_ the vendor branches are gone.

- [ ] as mentioned above, we want to AT LEAST remove the `*.mmdb` files from disk.

We cannot merge this PR until all these points have been fixed.

pkg/oonimkall/session.go
bassosimone: No, I've taken another approach

bassosimone:
Interesting: it seems https://github.com/ooni/probe-cli/pull/260/commits/5d18e33b33501c84fd72440afe0d87de2ff34edb has broken `miniooni` because the creation of the assets dir was _also_ creating the parent directory :facepalm:

bassosimone:
I am going to move forward with this PR. Because we publish binaries at GitHub, we can always recover the resources used by a given release by extracting them from the published binaries.

PR #263, opened 2021-03-25
feat(ExperimentOrchestraClient): add CheckIn

bassosimone:
We use ExperimentOrchestraClient in several places to help us
calling probe-services APIs. We need to call CheckIn because we
want to use CheckIn in InputLoader.

(We also want to remove the URLs API, but that is not something
doable now, since the mobile app is still using this API via
the wrappers at pkg/oonimkall.)

Work part of https://github.com/ooni/probe/issues/1299.

### DIFF
--- a/pkg/oonimkall/task_integration_test.go
+++ b/pkg/oonimkall/task_integration_test.go
@@ -315,7 +315,7 @@ func TestMaxRuntime(t *testing.T) {
 	// In case there are further timeouts, e.g. in the sessionresolver, the
 	// time used by the experiment will be much more. This is for example the
 	// case in https://github.com/ooni/probe-engine/issues/1005.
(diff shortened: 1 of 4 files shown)


pkg/oonimkall/task_integration_test.go
```diff
@@ -315,7 +315,7 @@ func TestMaxRuntime(t *testing.T) {
 	// In case there are further timeouts, e.g. in the sessionresolver, the
 	// time used by the experiment will be much more. This is for example the
 	// case in https://github.com/ooni/probe-engine/issues/1005.
-	if time.Now().Sub(begin) > 10*time.Second {
+	if time.Since(begin) > 10*time.Second {
```
bassosimone, line 318: This is a warning from `staticcheck` that I fixed while working on this diff.

PR #267, opened 2021-03-29
feat(inputloader): use check-in to fetch URLs

bassosimone:


internal/libminiooni/libminiooni.go
```diff
@@ -140,10 +145,6 @@ func init() {
 	)
 }
 
-func fatalWithString(msg string) {
-	panic(msg)
-}
-
```
bassosimone, line 146: unused

PR #278, opened 2021-03-31
feat(ooniprobe): discard lists not in selected categories

bassosimone:
One day we may make an integration mistake and for any reason
we may end up with URLs that do not belong to the categories
originally selected by the user. If that happens, it's nice to
have a safety net where we remove URLs that do not belong to
the right category before proceeding with testing.

This diff was conceived while discussing the robustness of
https://github.com/ooni/probe/issues/1299 with @hellais.

### DIFF
--- a/cmd/ooniprobe/internal/nettests/web_connectivity.go
+++ b/cmd/ooniprobe/internal/nettests/web_connectivity.go
@@ -9,6 +9,26 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/engine/model"
 )
 
+// preventMistakes makes the code more robust with respect to any possible
+// integration issue where the backend returns to us URLs that don't
+// belong to the category codes we requested.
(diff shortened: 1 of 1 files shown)


cmd/ooniprobe/internal/nettests/web_connectivity.go
```diff
@@ -9,6 +9,26 @@ import (
...
+	for _, entry := range input {
+		var found bool
+		for _, cat := range categories {
+			if entry.CategoryCode == cat {
+				found = true
+				break
+			}
+		}
+		if len(categories) > 0 && !found {
+			log.Warnf("URL %+v not in %+v; skipping", entry, categories)
```
bassosimone, line 25: A `continue` is missing here. Will fix and add a unit test.

bassosimone:
I'm gonna go ahead and merge. I'll wait for feedback asynchronously.

PR #279, opened 2021-03-31
release 3.9.0 process: reduce warnings

bassosimone:
This basically means going over the Go report card at https://goreportcard.com/report/github.com/ooni/probe-cli and reducing the number of warnings that are listed there. This work is part of the check list at https://github.com/ooni/probe/issues/1369.


bassosimone:
It seems we have a bunch of failures while running `alltests`. I am going to investigate these failures locally and fix them as part of a subsequent pull request, rather than attempting to fix then now. I believe it's reasonable to expect changes that I would like to merge separately for clarity and to better document what has been changed.

PR #280, opened 2021-03-31
release process: update dependencies

bassosimone:
Part of the check-list at https://github.com/ooni/probe/issues/1369


bassosimone:
I am going to merge this notwithstanding the build errors. I will try to make everything green in a subsequent PR.

internal/engine/netx/quicdialer/connectionstate.go
```diff
@@ -1,5 +1,3 @@
-// +build !go1.15
-
```
bassosimone, line 2: I removed backwards compatibility with go1.14 because now we're on go1.16.

internal/engine/netx/quicdialer/connectionstate.go
```diff
@@ -10,5 +8,5 @@ import (
 
 // ConnectionState returns the ConnectionState of a QUIC Session.
 func ConnectionState(sess quic.EarlySession) tls.ConnectionState {
-	return tls.ConnectionState{}
+	return sess.ConnectionState().TLS.ConnectionState
```
bassosimone, line 11: v0.20.x of lucas-clemente/quic-go moved `.ConnectionState` inside a `.TLS` subobject.

The function definition is now:

```Go
type Session interface {
	// ...
	ConnectionState() ConnectionState
}
```

In turn, `ConnectionState` is defined as

```Go
type ConnectionState struct {
	TLS               handshake.ConnectionState
	SupportsDatagrams bool
}
```

and `handshake.ConnectionState` is defined as

```Go
type ConnectionState = qtls.ConnectionState
```

and `qtls.ConnectionState` is

```Go
type ConnectionState = qtls.ConnectionStateWith0RTT
```

and `qtls.ConnectionStateWith0RTT` is

```Go
type ConnectionStateWith0RTT struct {
	ConnectionState
	Used0RTT bool // true if 0-RTT was both offered and accepted
}
```

and `ConnectionState` is

```Go
type ConnectionState = tls.ConnectionState
```

and `tls.ConnectionState` is the real `tls.ConnectionState`, which is not an interface.

Because none of these is an interface, the above code is fine: `TLS` cannot be `nil` and `ConnectionState` cannot be `nil`. (The issue we have here is that `sess` could be `nil` but this means the caller has called `ConnectionState()` on the wrong object &mdash; so I think all of this is fine.)

PR #282, opened 2021-04-01
fix(oohelper): make sure the CI is green

bassosimone:
Hopefully this helps with https://github.com/ooni/probe/issues/1409.

Part of https://github.com/ooni/probe/issues/1369


bassosimone:
The workflows are now green. It's a bit sad that I cannot figure out what was the original reason why that piece of code was failing yet. But, I suppose it makes sense to be pragmatic and go ahead with the release. I will eventually figure out :grimacing: :crossed_fingers:

PR #284, opened 2021-04-02
refactor: remove model.ExperimentOrchestraClient

bassosimone:
The main idea of this pull request is to refactor the way with which we fetch Psiphon configuration, so that we can introduce a private, optional package containing the Psiphon config. The easiest way to achieve this goal is to reduce the coupling between our Psiphpon wrappers and the probeservices package, by letting the Session make more work. Once this code has been written, I figured out it was generally a good idea to get rid of ExperimentOrchestraClient entirely and use the Session interface as the, well, interface between what experiment needs and what the probeservices package provides. This change is also instrumental to enable our future plan of further restructuring such a package. On the way, I also checked whether I could improve existing tests and generally increase our code coverage.

The reference issue for this work is https://github.com/ooni/probe/issues/985.

### DIFF
--- a/internal/engine/experiment/tor/tor_test.go
+++ b/internal/engine/experiment/tor/tor_test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/engine/legacy/oonitemplates"
 	"github.com/ooni/probe-cli/v3/internal/engine/model"
 	"github.com/ooni/probe-cli/v3/internal/engine/netx/errorx"
-	"github.com/ooni/probe-cli/v3/internal/engine/probeservices"
 )
 
 func TestNewExperimentMeasurer(t *testing.T) {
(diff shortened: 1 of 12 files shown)


internal/engine/experiment/tor/tor_test.go
```diff
@@ -929,3 +898,45 @@ func TestSummaryKeysWorksAsIntended(t *testing.T) {
...
+	if tk.DirPortAccessible != 1 {
+		t.Fatal("unexpected DirPortAccessible")
+	}
+	if tk.ORPortDirauthTotal != 2 {
+		t.Fatal("unexpected ORPortDirauthTotal")
+	}
+	if tk.ORPortDirauthAccessible != 1 {
+		t.Fatal("unexpected ORPortDirauthAccessible")
+	}
+}
```
bassosimone, line 942: I added two extra small unit test to make sure we cover all lines in the file. We soon need to refactor this package, so this feels like a good moment to ensure we have full coverage.

PR #286, opened 2021-04-03
feat: create tunnel inside NewSession

bassosimone:
We want to create the tunnel when we create the session. This change
allows us to nicely ignore the problem of creating a tunnel when we
already have a proxy, as well as the problem of locking. Everything is
happening, in fact, inside of the NewSession factory.

Modify miniooni such that --tunnel is just syntactic sugar for
--proxy, at least for now. We want, in the future, to teach the
tunnel to possibly use a socks5 proxy.

Because starting a tunnel is a slow operation, we need a context in
NewSession. This causes a bunch of places to change. Not really a big
deal except we need to propagate the changes.

Make sure that the mobile code can create a new session using a
proxy for all the APIs we support.

Make sure all tests are still green and we don't loose coverage of
the various ways in which this code could be used.

This change is part of https://github.com/ooni/probe/issues/985.


bassosimone:
This work is still incomplete. We have not adjusted all the tests. We have not added mobile support for propagating the proxy correctly. I will continue working on this after lunch, by implementing what is missing and adjusting what is broken.

internal/engine/session.go
```diff
@@ -119,25 +142,42 @@ func NewSession(config SessionConfig) (*Session, error) {
...
 		torArgs:                 config.TorArgs,
 		torBinary:               config.TorBinary,
 	}
+	proxyURL := config.ProxyURL
+	if proxyURL != nil {
+		switch proxyURL.Scheme {
+		case "psiphon", "tor":
+			tunnel, err := tunnel.Start(ctx, tunnel.Config{
+				Name:    proxyURL.Scheme,
+				Session: sess,
```
bassosimone, line 158: 🚨 This code is, sadly, wrong: `tunnel.Start` will ask the session to give it an HTTP transport, which we do not have yet and so we need to start the real tunnel in a deferred way. Or, we need to refactor the tunnel to be lazy. This is a bit of a blocker for moving forward. I need to think about another way or organizing the code.

bassosimone: I have fixed this circular dependency now.

PR #287, opened 2021-04-03
refactor: merge psiphonx and torx into tunnel

bassosimone:
This is a case where it seems that merging these three packages into
a single package will enable us to better the implementation.

The goal is still https://github.com/ooni/probe/issues/985.

The roadblock I'm trying to overcome is
https://github.com/ooni/probe-cli/pull/286#pullrequestreview-627460104.


bassosimone:
Now: the next step is probably to reduce the coupling between a tunnel and a session. We're now a bit impaired because the tunnel depends on the session, and we want instead the session to depend on the tunnel. The way to go is probably to avoid passing a Session directly to the tunnel config, but rather just the information that the tunnel requires. At this point, we will be in a position where we can start a tunnel without having a Session-like instance. This will allow us to resume https://github.com/ooni/probe-cli/pull/286.

PR #309, opened 2021-04-08
feat(ios): adapt build to post bintray world

bassosimone:
Part of https://github.com/ooni/probe/issues/1437


bassosimone:
I'm going to merge with failing tests because the specific failure seems useful to keep. The failure may be related with a rare problem with the API, which may depend on the API or on our code.

PR #312, opened 2021-04-22
Add support for tracking the is_uploaded status in the results table

hellais:
* Add unit tests for measurement upload status

This implements: https://github.com/ooni/probe/issues/1457


bassosimone:
:eyes: Thank you! I will review and merge this diff after 3.10.0-beta is out. (With 3.10.0-beta, we aim to solve https://github.com/ooni/probe/issues/1444, which is an Android-only issue; with that fixed, we can then focus on probe-cli and integrate this diff.)

cmd/ooniprobe/internal/database/actions.go
```diff
@@ -204,6 +204,45 @@ func DeleteResult(sess sqlbuilder.Database, resultID int64) error {
...
+	req := tx.Select(
+		db.Raw("SUM(measurements.measurement_is_uploaded)"),
+		db.Raw("COUNT(*)"),
+	).From("results").
+		Join("measurements").On("measurements.result_id = results.result_id").
+		Where("results.result_id = ?", result.ID)
+
+	err = req.One(&uploadedTotal)
+	if err != nil {
+		log.WithError(err).Error("failed to retrieve total vs uploaded counts")
```
bassosimone, line 225:
```suggestion
		log.WithError(err).Error("failed to retrieve total vs uploaded counts")
		return err
```

cmd/ooniprobe/internal/database/actions.go
```diff
@@ -57,7 +57,7 @@ func GetMeasurementJSON(sess sqlbuilder.Database, measurementID int64) (map[stri
 		log.Errorf("failed to run query %s: %v", req.String(), err)
 		return nil, err
 	}
-	if measurement.IsUploaded {
+	if measurement.Measurement.IsUploaded {
```
bassosimone, line 60: Now both `Measurement` and `Result` have `IsUploaded`. To disambiguate we need to explicitly specify that we want to select the `IsUploaded` attribute of the `Measurement`.

PR #314, opened 2021-04-26
Switch .deb package publishing to S3

FedericoCeratto:
WIP


.github/workflows/debops-ci
```diff
@@ -0,0 +1,563 @@
...
+        )
+        with open(tf.name, "w") as f:
+            f.write(html)
+
+        r, o = runc(f"{put} {tf.name} {baseuri}/index.html")
+
+    def lock(self, conf, baseuri):
+        """Rename semaphore file"""
+        print(f"Locking {baseuri} ...")
+        cmd = f"s3cmd mv --no-progress {baseuri}/.debrepos3.nolock {baseuri}/.debrepos3.lock"
```
hellais, line 430: Be careful that the mv operation is not atomic on s3 and is basically equivalent to doing a copy and delete. It might still be enough to prevent race conditions on the CI which has high latency, but it's probably worth noting this caveat in the code to prevent surprises down the line.

hellais (approved):
Left a minor comment, but otherwise looks good to me.

bassosimone:
Thank you, @FedericoCeratto! I will look into merging this package after 3.10.0-beta is out. (In 3.10.0-beta we fix a bunch of Android-related issues and, once that is done, we can prepare 3.10.0's `cli` release.)

.github/workflows/debops-ci
FedericoCeratto: The copy + delete process should implement something similar to a 2-step lock: during the interval when both files exist other attempt at copy should fail.

.github/workflows/debops-ci
```diff
@@ -0,0 +1,563 @@
+#!/usr/bin/env python3
+
+# Copyright 2020-2021 Federico Ceratto <federico@openobservatory.org>
+#                                      <federico@debian.org>
+# Released under GPLv3 license
```
bassosimone, line 5:
```suggestion
```

Let's keep the code under BSD 3-clause and the copyright to OONI. All the rest of the repository is released under these rules and there is no need that I can see to use different rules for this file specifically.

bassosimone (approved):
Thanks, I am going to merge this work soon!

bassosimone:
Not merged yet because it seems we still have some issues with the bucket.

PR #317, opened 2021-04-28
fix(engine): simplify logic around the submitter

bassosimone:
Useful to implement https://github.com/ooni/probe/issues/950


DecFox:
I am going to close this since we now support re-uploading with the multiplatform

PR #319, opened 2021-04-29
Release v3.10.0-beta.1

bassosimone:
* fix the way in which we generate the iOS build
* bump version number to v3.10.0-beta.1


bassosimone:
We'll deal with some tests not being green later.

PR #331, opened 2021-05-06
feat(gha): build debian package using ./make

bassosimone:
Part of https://github.com/ooni/probe/issues/1466. We're building both `arm64` and `amd64`. We are still not publishing `arm64` packages, which is what is asked in the original issue, but we're really close to doing that.


.github/workflows/linux.yml
```diff
@@ -1,6 +1,7 @@
 # linux builds ooniprobe for Linux
 name: linux
 on:
+  pull_request:
```
bassosimone, line 4:
```suggestion
```

PR #337, opened 2021-05-06
feat(gha/linux): upload deb packages for arm64

bassosimone:
We're still working on https://github.com/ooni/probe/issues/1466. The idea here is to teach the GH action for Linux to publish the debian package for arm64. When this is done, we can cleanup legacy build scripts and GH actions, because there is no remaining use case for them: we now build everything using the `./make` tool.


bassosimone:
In https://github.com/ooni/probe-cli/pull/337/checks?check_run_id=2524891027, we successfully published a Debian package for `arm64`. So, I am going to restrict the branches where we build for debian and do other cleanups.

PR #345, opened 2021-05-13
Release 3.10.0 beta.3

bassosimone:
See https://github.com/ooni/probe/issues/1468


bassosimone:
We oughta have an all green condition. This is mostly it: we have the usual transient failures on `fbmessenger` here. Plus, an issue of mismatch between the client definitions of the API and the testing API where @FedericoCeratto is doing development, so it's fine to ignore it for now. Finally, another (possibly transient?) issue in the tor test. All the rest is good.

PR #355, opened 2021-06-04
cleanup(platform): we don't need CGO anymore

bassosimone:
Since go1.16, we have the `ios` port. So we can easily
disambiguate between ios and darwin.

This means we don't need to rely on CGO to correctly guess
whether we are on ios or darwin anymore.

So, now miniooni does not depend on a C compiler even
when you are not cross compiling.


internal/platform/platform.go
```diff
@@ -24,28 +24,20 @@ import "runtime"
...
 func Name() string {
-	if name := cgoname(); name != "unknown" {
-		return name
-	}
-	return puregoname(runtime.GOOS, runtime.GOARCH)
+	return name(runtime.GOOS)
 }
 
-func puregoname(goos, goarch string) string {
+// name is a utility function for imlementing Name.
```
bassosimone, line 30:
```suggestion
// name is a utility function for implementing Name.
```

PR #362, opened 2021-06-08
refactoring: prepare for further netx reorganization

bassosimone:
I am currently working on https://github.com/ooni/probe-engine/issues/725 with an eye on https://github.com/ooni/probe/issues/1507. While approaching these issues, it seems that the `netx` codebase could use some consolidation (_coagula_) and some splitting (_solve_).

The general idea of these changes is that I want to arrive to the situation where we have (1) a `New` factory method for each package under `netx` for which it makes sense (e.g., `dialer.New`, `tlsdialer.New`, `httptransport.New`), (2) a separate `Config` structure per package (e.g., `dialer.Config`) rather than having all the possible config into the same structure (3) part of the `urlgetter` code (and namely the low-level part) moved into the `netx` package. (See https://github.com/ooni/probe/issues/1591.)

There is too much bureaucracy around writing a new experiment. Much of this bureaucracy will go (it seems) if we do what I have said above. After that, we will end up that you run tests by using the top-level `netx` package. (In any case, I am not of course 100% sure about all the changes that will come after, but this comment seems enough to set a direction.)

These are the changes in this pull request:

* refactor(netx): extract tlsdialer from dialer

We want these two packages to be separate. Dialer was doing too much before. Separating TLS dialing code into a separate package allows us to have a `tlsdialer.New` factory that clearly conveys the meaning. Also, this would allow us to much more clearly separate configuration and simplify reasoning on what each factory does.

* refactor(engine): move `tlsx` package inside `netx` and merge the `gocertifi` package inside it

It would be tempting to merge everything inside `tlsdialer` but the reality is that also the `quicdialer` package needs to use the same code, therefore, it seems more tidy to actually have some tls extensions in `netx`.


bassosimone:
This PR is still in progress because I realized about the need for a `tlsx` package only when writing the description of the PR and I would like to do that before declaring the pull request ready.

bassosimone:
I am going to merge this PR with a merge commit for clarity. I have verified each individual commit still passes the `-short -race` tests locally. This should be enough to ensure we don't break bisection in the future.

PR #377, opened 2021-06-14
cli: error wrapper refactoring

kelmenhorst:
Open issues/considerations, before merging:
- [ ] Many places in the code (mainly, but not only in the tests) use the OONI error string that was generated by the `MaybeBuild` func at `ErrWrapper` initialization, to verify results or make decisions. After removing the `ErrWrapper`, I replaced all checks with a call to `archival.NewFailure()`, to give me the OONI error. This is definitely not ideal, since now, all of those packets depend on `archival`. My idea would be to classify the OONI string whenever we wrap an error and store it as a field in `dialer.ErrDial` etc. (This however would go against our considerations that `archival` should be responsible for the classification.)
- [ ] Split up the classification code in `archival.ToFailureString`, so that it is more operation specific. For example, we cannot expect QUIC errors to occur during a HTTPS (over TCP) transaction.
- [ ] We should probably also wrap errors occurrring during TopLevel operations. I am inclined to giving each experiment it's own `ErrTopLevel`.
- [ ] Merge quic-go version changes (PR #376, once it is merged)  from master in here


kelmenhorst:
We will close this PR and replace it with one according to the updated issue [#1505](https://github.com/ooni/probe/issues/1505).

kelmenhorst:
I will delete branch `issue/1591` because it contains discarded code. The error refactoring that we intended to do with this PR and branch was instead implemented and [merged](https://github.com/ooni/probe-cli/pull/386) using branch `issue/1505`.

PR #379, opened 2021-06-15
fix(all): introduce and use iox.ReadAllContext

bassosimone:
This improvement over the ioutil.ReadAll utility returns early
if the context expires. This enables us to unblock stuck code in
case there's censorship confounding the TCP stack.

See https://github.com/ooni/probe/issues/1417.

Compared to the functionality postulated in the above mentioned
issue, I choose to be more generic and separate limiting the
maximum body size (not implemented here) from using the context
to return early when reading a body (or any other reader).

After implementing iox.ReadAllContext, I made sure we always
use it everywhere in the tree instead of ioutil.ReadAll.

This includes many parts of the codebase where in theory we don't
need iox.ReadAllContext. Though, changing all the places makes
checking whether we're not using ioutil.ReadAll where we should
not be using it easy: `git grep` should return no lines.

PR #380, opened 2021-06-15
fix(all): introduce and use iox.CopyContext

bassosimone:
This PR is part of https://github.com/ooni/probe/issues/1417.

In https://github.com/ooni/probe-cli/pull/379 we introduced a context
aware wrapper for io.ReadAll (formerly ioutil.ReadAll).

Here we introduce a context aware wrapper for io.Copy.


internal/engine/experiment/ndt7/ndt7.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	testName    = "ndt"
-	testVersion = "0.8.0"
+	testVersion = "0.9.0"
```
bassosimone, line 22: Let's bump the version since we've changed the way in which we read messages to honour the context.

internal/humanize/humanize.go
```diff
@@ -7,7 +7,7 @@ import "fmt"
 // specially tailored for printing download speeds.
 func SI(value float64, unit string) string {
 	value, prefix := reduce(value)
-	return fmt.Sprintf("%3.0f %s%s", value, prefix, unit)
+	return fmt.Sprintf("%6.2f %s%s", value, prefix, unit)
```
bassosimone, line 10: I don't like to see just `5 Gbit/s` w/o some extra digits

PR #386, opened 2021-06-16
cli: error classification refactoring

kelmenhorst:
Referring to issue [#1505](https://github.com/ooni/probe/issues/1505).
Modifies the `errorx` classifier (`errorx.toFailureString`) to make it more Windows-compatible and less dependent on error strings.
Introduces local error classification for QUIC (`quicdialer`) and Resolve (`resolver`) errors.

Before merging:
- [ ] determine if there are other possibilities for local classifiers

Closes https://github.com/ooni/probe/issues/1526


internal/engine/netx/errorx/errorx.go
```diff
@@ -244,50 +280,83 @@ func toFailureString(err error) string {
...
+			return FailureConnectionRefused
+		}
+		// checkout alert.go in the qtls library
+		// (the alert constants are not exported, so this is not robust to changes in qtls)
+		errCode := uint8(transportError.ErrorCode)
+		if isCertificateError(errCode) {
 			return FailureSSLInvalidCertificate
 		}
+		if errCode == TLSDecryptError || errCode == TLSHandshakeFailure {
+			return FailureSSLHandshake
```
bassosimone, line 323: Do we perhaps want to treat these two as separate errors? What is the rationale of coalescing them?

internal/engine/netx/errorx/errorx.go
```diff
@@ -320,3 +389,11 @@ func toOperationString(err error, operation string) string {
 	}
 	return operation
 }
+
+func isCertificateError(alert uint8) bool {
+	return (alert == TLSBadCertificate ||
```
bassosimone, line 394: Where does this logic come from? Is it based on empirical experience or is it mirroring some existing code?

internal/engine/netx/errorx/errorx.go
bassosimone: (I see this is a new error that we were not previously reporting, right? So a good question would probably be if we could not take advantage of existing errors? Is this a new error condition? If so, could we take advantage of this error condition and, instead, perhaps _also_ use this error for non-quic handshakes?)

internal/engine/netx/quicdialer/errorwrapper_test.go
```diff
@@ -43,6 +43,28 @@ func errorWrapperCheckErr(t *testing.T, err error, op string) {
...
+	if err == nil {
+		t.Fatal("expected an error here")
+	}
+	if sess != nil {
+		t.Fatal("expected nil sess here")
+	}
+	if err.Error() != errorx.FailureSSLInvalidCertificate {
+		t.Fatal("unexpected failure")
+	}
+}
```
bassosimone, line 67: Side note: the fact that we need to write tests for this functionality we're adding in `errorx` in _another_ package does seem to indicate that perhaps part of the classifier should live here instead. (There will _always_ be some cross dependency since `errorx` contains the errno classifier but it seems to me moving the QUIC specific classifier here would reduce the coupling, and hence enhance maintainability.)

Would you mind adding a comment in `errorx.go` mentioning that perhaps we want to do that in the future?

internal/engine/netx/errorx/errorx.go
```diff
@@ -244,50 +280,83 @@ func toFailureString(err error) string {
 	if strings.HasSuffix(s, "TLS handshake timeout") {
 		return FailureGenericTimeoutError
 	}
```
bassosimone, line 282: @kelmenhorst I suppose the above three lines are TLS specific, right? Maybe add a comment mentioning that we should move them or, alternatively, just move them right away?

bassosimone (commented):
Thank you @kelmenhorst for spearheading this work! The `errorx` package needed some attention to improve its implementation standards as well as to more correctly map onto errors on Windows. With this work, you have done that, and this is great!

I think it's also very useful to start splitting the code and have operation-specific errors separated from general errno-like errors. This is good for future maintainability!

Before merging, there are a bunch of small nits and code refactoring suggestions that I have highlighted and I would like you to apply to this branch.

Also, there were a couple of places where I was not 100% sure about what I was reading and hence needed some clarifying answers.

Thanks! ☺️

internal/engine/netx/errorx/errorx.go
```diff
@@ -196,42 +240,34 @@ func toFailureString(err error) string {
...
 
-	if errors.Is(err, ErrDNSBogon) {
-		return FailureDNSBogonError // not in MK
+	// filter out system errors
+	var errno syscall.Errno
+	if errors.As(err, &errno) {
+		// checkout https://pkg.go.dev/golang.org/x/sys/windows and https://pkg.go.dev/golang.org/x/sys/unix
+		switch errno {
+		case ECANCELED:
+			return FailureInterrupted
```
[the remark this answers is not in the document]
bassosimone: Update: I double-checked. We're using `FailureInterrupted` both for `ECANCELED` and for `context.Canceled`.

internal/engine/netx/errorx/errorx.go
```diff
@@ -196,42 +240,34 @@ func toFailureString(err error) string {
...
+		case ECONNRESET:
+			return FailureConnectionReset
+		case ECONNREFUSED:
+			return FailureConnectionRefused
+		case EHOSTUNREACH:
+			return FailureHostUnreachable
+		case ETIMEDOUT:
+			return FailureGenericTimeoutError
```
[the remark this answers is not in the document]
bassosimone: Update: I double-checked and this is good!

internal/engine/netx/errorx/errorx.go
```diff
@@ -196,42 +240,34 @@ func toFailureString(err error) string {
...
-	}
-	var x509CertificateInvalidError x509.CertificateInvalidError
-	if errors.As(err, &x509CertificateInvalidError) {
-		// Test case: https://expired.badssl.com/
-		return FailureSSLInvalidCertificate
-	}
-
 	s := err.Error()
-	if strings.HasSuffix(s, "operation was canceled") {
+	if strings.Contains(s, "operation was canceled") {
```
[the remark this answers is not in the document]
bassosimone: Could you perhaps add a comment mentioning the need to use `errors.Is` more in this code?

internal/engine/netx/errorx/errorx.go
```diff
@@ -6,8 +6,10 @@ import (
 	"crypto/x509"
```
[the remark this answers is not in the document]
bassosimone: > [...] at the same time think doing this change now would make the diff less readable than it currently is.

Motivation: if we have a regression in the future and we need to bisect, it's ~important that the scope of the change is understandable when we try to figure out what happens. Hence, the desired to not rewrite too much stuff at a time.

internal/engine/netx/errorx/errorx.go
```diff
@@ -49,6 +57,36 @@ const (
...
+	TLSCertificateExpired = 45
+
+	// Some unspecified issue arose inprocessing the certificate, rendering it unacceptable.
+	TLSCertificateUnknown = 46
+
+	// Certificate was not accepted because the CA certificate could not be located or could not be matched with a known trust anchor.
+	TLSUnknownCA = 48
+
+	// Handshake (not record layer) cryptographic operation failed.
+	TLSDecryptError = 50
```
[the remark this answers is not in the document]
bassosimone: (FWIW, I double-checked the codes with not only qtls but also the RFC and it looks qtls numbers are correct for all the alter codes we have here, _except_ decrypt error, which is `51` also in the RFC.)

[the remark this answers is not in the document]
kelmenhorst: Oh sorry, yes, that is a typo. 
I didn't incorporate `DecodeError` since we don't have a TLS record layer error type yet, not for QUIC and not for TLS over TCP. `DecryptError` describes a failure during the handshake.

internal/engine/netx/errorx/errorx.go
kelmenhorst: the rationale was that both errors are caused by a failed or corrupted parameter negotiation during the TLS handshake

internal/engine/netx/errorx/errorx.go
kelmenhorst: This is based on my understanding of the error condition described here: https://datatracker.ietf.org/doc/html/rfc8446#section-6.2

internal/engine/netx/errorx/errorx.go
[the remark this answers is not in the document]
bassosimone: Thank you!

internal/engine/netx/errorx/errorx.go
bassosimone: I would add a comment saying that, then!

internal/engine/netx/quicdialer/errorwrapper_test.go
kelmenhorst: I'm not sure if I agree that it will be better to move them. But I added the comment so we can keep that in mind :)

bassosimone: Thanks! Why do you disagree with moving them?

internal/engine/netx/errorx/errorx.go
```diff
@@ -244,50 +280,83 @@ func toFailureString(err error) string {
 	if strings.HasSuffix(s, "TLS handshake timeout") {
 		return FailureGenericTimeoutError
 	}
-	if strings.HasSuffix(s, "no such host") {
+	if strings.HasSuffix(err.Error(), "no such host") {
```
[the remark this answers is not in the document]
kelmenhorst: I have no idea :D

internal/engine/netx/quicdialer/saver_test.go
```diff
@@ -85,7 +85,7 @@ func TestHandshakeSaverSuccess(t *testing.T) {
 
 func TestHandshakeSaverHostNameError(t *testing.T) {
 	nextprotos := []string{"h3"}
-	servername := "wrong.host.badssl.com"
+	servername := "example.com"
```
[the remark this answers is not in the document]
kelmenhorst: `wrong.host.badssl.com` does not support QUIC. In this test, the host name error is triggered by targeting a Google IP while using a different domain name as SNI. Using `wrong.host.badssl.com` suggests that this service is the reason why the host name error occurs which is not the case.

internal/engine/netx/errorx/errorx.go
bassosimone: could you add a comment mentioning that please?

internal/engine/netx/errorx/errorx.go
```diff
@@ -244,50 +280,83 @@ func toFailureString(err error) string {
...
-	// from qtls rather than strings for TLS errors below.
-	//
-	// TODO(kelmenhorst): make sure we have tests for all errors. Also,
-	// how to ensure we are robust to changes in other libs?
-	//
-	// special QUIC errors
-	matched, err := regexp.MatchString(`.*x509: certificate is valid for.*not.*`, s)
-	if matched {
-		return FailureSSLInvalidHostname
+func ClassifyQUICFailure(err error) string {
```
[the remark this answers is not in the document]
kelmenhorst: you mean other than `errorx.TestClassifyQUICFailure` ?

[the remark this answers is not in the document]
bassosimone: Yes, I am wondering whether there are ways to provoke these errors in a real scenario and make sure that the QUIC library indeed returns such errors so we are actually doing things correctly beyond unit testing.

internal/engine/netx/quicdialer/errorwrapper_test.go
kelmenhorst: Well, these classifiers are used to translate a given error to a string constant which is defined in `errorx`. So I think that if `errorx` introduces and defines those OONI error strings, it makes sense that `errorx` also handles the translation/mapping. It seems wrong (at least to me) to let the transport layer packages handle this mapping... :thinking:

internal/engine/netx/errorx/errorx.go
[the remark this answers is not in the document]
kelmenhorst: Yes, I see, I think this should be doable for at least a few errors. Where should those integration tests live?

internal/engine/netx/quicdialer/errorwrapper_test.go
kelmenhorst: I do see the point that it does not seem very elegant to use the `errorx` string as a check in `quicdialer/errorwrapper_test`. Saying that, I realize, that I aimed to do an integration test of a "real" QUIC failure, as you mentioned above, with this test.

bassosimone: I see what you mean, and that's a good point I also see myself!

The flip side is that you need to write tests in, say, `quicdialer` to ensure that the mapping that is defined in `errorx` is working correctly for real networking scenarios. At the same time, when you want to, say, add a new error for QUIC, you need to modify code in `quicdialer` and then `errorx` and then make sure that they are in sync, which kinda breaks the nice property that packages should be as self-contained as possible.

It may be that it's more convenient to have everything quic-errors related in `quicdialer` and to use `errorx` only as a support package to classify the errors that we don't otherwise know how to classify. Even this solution, though, may not be ideal considering that we would still probably have some functionality here and some functionality there.

Given it's not clear what's best, the best course of action is just documenting this as something we may wanna consider for the future, next time we will need to apply changes to this part of the tree.

kelmenhorst: yes :+1: I understand. It's difficult to clearly separate the functionalities here.

internal/engine/netx/errorx/errorx.go
```diff
@@ -241,6 +289,7 @@ func toFailureString(err error) string {
 	if strings.HasSuffix(s, "i/o timeout") {
 		return FailureGenericTimeoutError
 	}
+	// TODO(kelmenhorst,bassosimone): this can probably be (re-)moved
```
bassosimone, line 293:
```suggestion
	// TODO(kelmenhorst,bassosimone): this can probably be moved since it's TLS specific
```

internal/engine/netx/errorx/errorx.go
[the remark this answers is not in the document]
bassosimone: > Where should those integration tests live?

They cannot live in `errorx` unless we use external testing because of circular dependencies, so probably the most rational place where to put them is inside of the `quicdialer`, `dialer`, `tlsdialer`, etc. packages.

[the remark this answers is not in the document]
kelmenhorst: Yes. I did such a test in `TestErrorWrapperInvalidCertificate` (`quicdialer/errorwrapper_test.go`) but it would be good to add more.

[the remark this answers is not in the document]
bassosimone: yes!

bassosimone (approved):
The code in here looks good to me! I am going to perform QA and merge!

🐳

PR #389, opened 2021-06-18
feat: introduce services and rewrite tor

bassosimone:
This PR contains code that does two things:

1. introduces the concept of "services" for build block operations
2. re-writes the tor experiment to use such services

The concept of "services" is quite a U-turn from the original design of `netx`. The original design was that we were using code as idiomatic as possible and we were collecting measurements as a side effect.

It turns out that this design hits two problems:

1. it is not super easy to run parallel operations w/o adding complexity
2. we still have lots of complexity and magic, when we use the context and when we do error wrapping

The alternative design of the services is an attempt to solve these two problems as follows.

We can have better parallelism by defining the number of operations (e.g., connect, resolve, TLS handshake) that can go in parallel and by starving goroutines that cannot be serviced immediately.

We have much less complexity because in most cases we can collect data after a micro-operation and we know exactly which operation failed and we can immediately fill the measurements.

The code in here has still some problems and is mostly still a proof of concept. I'm committing this code to signal that the first stage of https://github.com/ooni/probe/issues/1596 has now been completed.

The follow-up tracking issue for continuing this work is: https://github.com/ooni/probe/issues/1688


bassosimone:
We just merged into master a small rewrite of tor that avoids using the legacy netx.

PR #405, opened 2021-06-25
refactor(quicdialer): separate saving from listening

bassosimone:
With this change, we will soon be able to move the creation of
a QUIC session inside of the netxlite package.

Part of https://github.com/ooni/probe/issues/1505.


bassosimone:
(I will write more tests as I proceed with the refactoring)

PR #413, opened 2021-06-30
refactor: introduce factory for stdlib http transport

bassosimone:
With this factory, we want to construct ourselves the TLS dialer
so that we can use a dialer wrapper that always sets timeouts when
reading, addressing https://github.com/ooni/probe/issues/1609.

As a result, we cannot immediately replace the i/e/netx factory
for creating a new HTTP transport, since the functions signatures
are not directly compatible.

Refactoring is part of https://github.com/ooni/probe/issues/1505.

### DIFF
--- a/internal/engine/netx/httptransport/http3transport.go
+++ b/internal/engine/netx/httptransport/http3transport.go
@@ -5,6 +5,10 @@ import (
 )
 
 // NewHTTP3Transport creates a new HTTP3Transport instance.
+//
+// Deprecation warning
+//
+// New code should use netxlite.NewHTTP3Transport instead.
 func NewHTTP3Transport(config Config) RoundTripper {
 	return netxlite.NewHTTP3Transport(config.QUICDialer, config.TLSConfig)
 }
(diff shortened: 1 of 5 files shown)


bassosimone:
The CI failure is like:

```
--- FAIL: TestOBFS4DialerWorks (15.02s)
Error:     obfs4_test.go:20: dial tcp 192.95.36.142:443: i/o timeout
--- FAIL: TestOBFS4DialerWorksWithContextExpiration (131.03s)
Error:     obfs4_test.go:136: the goroutine did not call close
FAIL
```

This seems to be an unrelated problem with obfs4 bridges.

bassosimone:
The bridge is indeed currently offline according to https://metrics.torproject.org/rs.html#details/D9C805C955CB124D188C0D44F271E9BE57DE2109

bassosimone:
Here's another possible bridge we can use: https://metrics.torproject.org/rs.html#details/FD96EFC317B7B3C3DBA25743945C9FF9496F63A3

PR #414, opened 2021-06-30
cli: utls for TLS parroting

kelmenhorst:
This is the ground work for [issue #1424](https://github.com/ooni/probe/issues/1424).
This PR introduces the use of [utls](https://gitlab.com/yawning/utls) for mimicking the TLS ClientHellos of popular browser implementations in order to decrease the risk of OONI clients being blocked based on their characteristic TLS fingerprint. The functionality is implemented by a UTLSHandshaker in `netxlite`.

When using utls, `net/http` fails to derive the right HTTP version from the ALPN (the issue is described [here](https://github.com/refraction-networking/utls/issues/16)). To solve this, the PR adds a `roundTripper` wrapper (package `httptransport`) to take care of this decision:

roundtripper


internal/netxlite/tlshandshaker.go
```diff
@@ -26,6 +28,10 @@ type TLSHandshakerConfigurable struct {
 	// Timeout is the OPTIONAL timeout imposed on the TLS handshake. If zero
 	// or negative, we will use default timeout of 10 seconds.
 	Timeout time.Duration
+
+	// UTLSHandshaker is the OPTIONAL Handshaker that uses utls for TLS parroting
+	// if this Handshaker is not set, we'll use net
+	UTLSHandshaker *UTLSHandshaker
```
bassosimone, line 34: Rather than adding an `UTLSHandshaker` type, could you instead provide an implementation of `NewConn` that returns a connection that implements `TLSConn` and uses `*utls.UConn` as the underlying conn?

bassosimone (changes requested):
I would like to proceed to merge this PR piecemeal. The first step I'd like to achieve is to add support for handshaking using `utls`. I would like you to focus just on changes in that part of the codebase for now, by trying to implement using `utls` via the `NewConn` factory rather than adding a new `UTLSHandshaker` type.

Once we have refactored this part of the PR, we'll work to only merge that part and then we'll continue the discussion regarding the rest of your PR.

My general sense is that the code you've written is consistent with the way in which we do things currently in `probe-cli`, _but_ it may be that there is a better way that could allow us to perform the _same_ measurements with less code. Because this PR addresses a crucial and fundamental issue in `probe-cli`, it may be worth it spending some extra brain cycles to figure out the extent to which we can simplify and improve the code design.

Thank you!

(To further clarify what I mean: I suggest working on this branch and applying the required `NewConn` change. Once that is okay, I will take care of merging this part of your contribution to master. Once that is done, I'd like to have a design conversation regarding a few design options that I see here, so that we can discuss how to finish the rest of the work.)

go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	github.com/sirupsen/logrus v1.7.0 // indirect
 	github.com/ziutek/mymysql v1.5.4 // indirect
 	gitlab.com/yawning/obfs4.git v0.0.0-20210511220700-e330d1b7024b
+	gitlab.com/yawning/utls.git v0.0.12-1 // indirect
```
bassosimone, line 42: I am surprised it's indicated as `indirect`! Would you mind running `go mod tidy` and checking whether this is enough to mark this dependency as a direct dependency?

internal/netxlite/tlshandshaker.go
```diff
@@ -105,4 +107,42 @@ func (h *TLSHandshakerLogger) Handshake(
 	return tlsconn, state, nil
 }
 
+// UTLSConn implements TLSConn and uses a utls UConn as its underlying connection
+type UTLSConn struct {
+	*utls.UConn
+}
+
+// this factory creates a NewConn function creating a utls connection with a specified ClientHelloID
```
bassosimone, line 115:
```suggestion
// NewConnUTLS creates a NewConn function creating a utls connection with a specified ClientHelloID
```

internal/netxlite/tlshandshaker.go
```diff
@@ -105,4 +107,42 @@ func (h *TLSHandshakerLogger) Handshake(
...
+// TODO(kelmenhorst) Should this be here or in netx?
```
bassosimone, line 116:
```suggestion
```

No, it's fine to keep it here, thank you!

internal/netxlite/tlshandshaker.go
```diff
@@ -105,4 +107,42 @@ func (h *TLSHandshakerLogger) Handshake(
...
+func NewConnUTLS(clientHello utls.ClientHelloID) func(conn net.Conn, config *tls.Config) TLSConn {
```
bassosimone, line 117: could `clientHello` be a pointer?

bassosimone (changes requested):
I'm requesting some small changes before we can merge the part in which extend netxlite so that later we can focus on the rest of the diff. Thank you!

bassosimone:
Superseded by https://github.com/ooni/probe-cli/pull/432

PR #431, opened 2021-08-06
cli: new web connectivity test helper

kelmenhorst:
This pull request adds a prototype of the new web connectivity test helper which supports testing multiple IP endpoints per domain and introduces HTTP/3 control measurements.

The issue for this PR can be found here: https://github.com/ooni/probe/issues/1729

The companion specification PR is here: https://github.com/ooni/spec/pull/219


internal/cmd/oohelperd/internal/nwcth/dns.go
```diff
@@ -0,0 +1,39 @@
...
+	Domain string
+}
+
+// NewResolver creates a new DNS resolver instance
+func NewResolver() netx.Resolver {
+	return netx.NewResolver(netx.Config{Logger: log.Log})
+}
+
+// DNSDo performs the DNS check.
+func DNSDo(ctx context.Context, config *DNSConfig) CtrlDNSResult {
```
bassosimone, line 31: Can you return a `*CtrlDNSResult` here? I'm trying to follow this pattern ~everywhere now.

internal/cmd/oohelperd/internal/nwcth/dns.go
```diff
@@ -0,0 +1,39 @@
...
+// the Web Connectivity test helper.
+type CtrlDNSResult = nwebconnectivity.ControlDNS
+
+// DNSConfig configures the DNS check.
+type DNSConfig struct {
+	Domain string
+}
+
+// NewResolver creates a new DNS resolver instance
+func NewResolver() netx.Resolver {
```
bassosimone, line 26: Unless there's a reason to export `NewResolver`, I think we can rename it `newResolver`.

internal/cmd/oohelperd/internal/nwcth/fake_test.go
```diff
@@ -0,0 +1,127 @@
+package nwcth
```
bassosimone, line 1: Please, check whether you can use `internal/netmocks` instead?

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+
+// HTTPConfig configures the HTTP check.
+type HTTPConfig struct {
+	Jar       *cookiejar.Jar
+	Headers   map[string][]string
+	Transport http.RoundTripper
+	URL       *url.URL
+}
+
+// HTTPDo performs the HTTP check.
```
bassosimone, line 23: Can you also document what is returned and why? Something like `CtrlHTTPRequest is [...] and NextLocationInfo is [...]`. Make sure you mention when they can be `nil` (if they can). In general, it would be good to mention when a function returns something and when it returns `nil`. (Absent an optional type, it is quite useful to have a way to represent the absence via `nil`.)

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+		Headers:    headers,
+	}, httpRedirect
+}
+
+func newRequest(ctx context.Context, URL *url.URL) (*http.Request, error) {
+	scheme := URL.Scheme
+	if strings.Contains(scheme, "h3") {
+		scheme = "https"
+	}
+	return http.NewRequestWithContext(ctx, "GET", scheme+"://"+URL.Hostname(), nil)
```
bassosimone, line 87: I think this function should be improved as follows:

1. I'd match for equality for the QUIC versions we support rather than using a `string.Contains`

2. the way in which you construct a new URL completely forgets about the URL path.

Regarding the latter point, please use this pattern:

1. create a new URL

2. initialize such an URL using the original URL as input

3. replace the scheme if needed

4. use the new URL serialized as a string to initialize the new request

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	data, err := iox.ReadAllContext(ctx, reader)
+	return &CtrlHTTPRequest{
+		BodyLength: int64(len(data)),
+		Failure:    newfailure(err),
+		StatusCode: int64(resp.StatusCode),
+		Headers:    headers,
+	}, httpRedirect
+}
+
+func newRequest(ctx context.Context, URL *url.URL) (*http.Request, error) {
```
bassosimone, line 82: Please add documentation to this function: I think it's useful to explain why we're using it.

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/ooni/probe-cli/v3/internal/iox"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
+// HTTPConfig configures the HTTP check.
+type HTTPConfig struct {
+	Jar       *cookiejar.Jar
+	Headers   map[string][]string
+	Transport http.RoundTripper
+	URL       *url.URL
```
bassosimone, line 20: You should document which fields are mandatory and which are optional. For example, the `Jar` seems to be optional and also `Headers` seem optional, while `Transport` seems mandatory. If something is optional, document what happens if it's not provided. I'd also like to see the generic type for `Jar` (i.e., the interface in `net/http`) being used as opposed to the concrete type.

Hint that may be useful (later) to implement testing: if, say, `Transport` is optional and you make it optional and you construct a suitable `Transport` if it's not set, you probably have minimized the scope of `Transport`, reduced what we called bureaucracy, and reduced the amount of code you need to write in the caller, isn't it?

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return &CtrlHTTPRequest{Failure: newfailure(err)}, nil
+	}
+	// The original test helper failed with extra headers while here
+	// we're implementing (for now?) a more liberal approach.
+	for k, vs := range config.Headers {
+		switch strings.ToLower(k) {
+		case "user-agent":
+		case "accept":
+		case "accept-language":
```
bassosimone, line 35: Would you mind updating the specification (which currently is a comment in the reference issue) to bump it's version and to mention that those headers are the only ones that are passed along? (Also, once we do that, the comment can go away, since now we're following the _new_ spec.)

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: (It may be tempting to use `RequestHeaders` and `RequestURL` for additional clarity. Sometimes I'm a bit more brief with naming, especially when there is inline docs such as

```Go
// Headers contains the request headers
Headers map[string][]string
```

because one can hover with the mouse and see what a field is. For this reason, it may be smart to provide a small piece of documentation to any field of any structure. Hovering can be very useful when navigating code.)

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	var redirectReq *http.Request
+
+	jar := config.Jar
+	if jar == nil {
+		jar, err = cookiejar.New(nil)
+		runtimex.PanicOnError(err, "cookiejar.New failed")
+	}
+	client := http.Client{
+		CheckRedirect: func(r *http.Request, reqs []*http.Request) error {
+			redirectReq = r
```
bassosimone, line 51: Are you sure `CheckRedirect` is called from the same goroutine in which we define `redirectReq`? Did you try running the code with `-race`? I am not sure we have enough guarantee for the docs to disregard the possibility that the `CheckRedirect` callback will run in another goroutine.

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+			}
+		}
+	}
+	// redirectReq is a clone of the initial request, with possibly modified headers
+	var redirectReq *http.Request
+
+	jar := config.Jar
+	if jar == nil {
+		jar, err = cookiejar.New(nil)
+		runtimex.PanicOnError(err, "cookiejar.New failed")
```
bassosimone, line 47: I have seen a bunch of cases where you use this pattern `x, _ := operation()` where you're sure that the operation cannot fail. I'd rather see you use `runtimex.PanicOnError` in all these cases! 🙏

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+			return http.ErrUseLastResponse
+		},
+		Jar:       jar,
+		Transport: config.Transport,
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return &CtrlHTTPRequest{Failure: newfailure(err)}, nil
+	}
+	var httpRedirect *NextLocationInfo = nil
```
bassosimone, line 61:
```suggestion
	var httpRedirect *NextLocationInfo
```

Idiomatic ^^^. Stuff is zero initialized in Go.

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	loc, _ := resp.Location()
+	if loc != nil && redirectReq != nil {
+		loc.Scheme = config.URL.Scheme
+		httpRedirect = &NextLocationInfo{jar: jar, location: loc.String(), httpRedirectReq: redirectReq}
```
bassosimone, line 65: *Design question*: it is not fully clear to me why we're saving both the redirect request and the location. Isn't the redirect request what we get given a set of cookies and the location URL anyway? Is there another reason why we need to also save the redirect request? Cannot we just return `ErrUseLastRequest`, disregard what would the redirect request be, and then use `Location` to see whether we're redirected? I am also failing to see where we are checking whether we _should_ follow the redirect. Say the server returns `404` plus `Location`. In such a case, will we follow the redirect or not?

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	defer resp.Body.Close()
+	headers := make(map[string]string)
```
bassosimone, line 68: *Design question* can we use a representation of headers that more true to reality (i.e., `http.Headers`)? Sadly, it's not correct to use a `map[string]string` for headers because duplicate headers are legal.

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	for k := range resp.Header {
+		headers[k] = resp.Header.Get(k)
+	}
+	reader := &io.LimitedReader{R: resp.Body, N: maxAcceptableBody}
+	data, err := iox.ReadAllContext(ctx, reader)
+	return &CtrlHTTPRequest{
```
bassosimone, line 74: Can this structure be called `HTTPMeasurement` like in the spec?

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: ❌ note that the specification recommends `map[string][]string`, which is ~`http.Headers`

internal/cmd/oohelperd/internal/nwcth/internal.go
```diff
@@ -0,0 +1,51 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/ooni/probe-cli/v3/internal/iox"
+	"github.com/ooni/probe-cli/v3/internal/version"
+)
+
+const maxAcceptableBody = 1 << 24
```
bassosimone, line 13: I think we need to clarify that this is the maximum acceptable body both for JSON and for webpages we fetch and we also need to justify why we have chosen this value.

Take a look at https://github.com/ooni/probe/issues/1727 for useful data to link to.

internal/cmd/oohelperd/internal/nwcth/internal.go
```diff
@@ -0,0 +1,51 @@
...
+		return
+	}
+	cresp, err := Measure(req.Context(), &creq)
+	if err != nil {
+		w.WriteHeader(400)
+		return
+	}
+	// We assume that the following call cannot fail because it's a
+	// clearly serializable data structure.
+	data, _ = json.Marshal(cresp)
```
bassosimone, line 48: I think here the code is better written using `runtimex.PanicOnError`.

bassosimone: (If you happen to copy this snippet from the current test helper, would you mind fixing the original as well?)

internal/cmd/oohelperd/internal/nwcth/internal.go
```diff
@@ -0,0 +1,51 @@
...
+		"oohelperd/%s ooniprobe-engine/%s", version.Version, version.Version,
+	))
+	if req.Method != "POST" {
+		w.WriteHeader(400)
+		return
+	}
+	if req.Header.Get("content-type") != "application/json" {
+		w.WriteHeader(400)
+		return
+	}
```
bassosimone, line 29: On second though, I think this is too strict. What if the client sets `application/json; charset=UTF-8`? I think we should be liberal in what we accept and just assume the body is JSON and parse it as such.

bassosimone: (If this is also a problem of the current test helper, could you also fix it there?)

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+	CtrlHTTPMeasurement = nwebconnectivity.ControlHTTP
+
+	CtrlH3Measurement = nwebconnectivity.ControlH3
+
+	CtrlTLSMeasurement = nwebconnectivity.ControlTLSHandshake
+
+	CtrlHTTPRequest = nwebconnectivity.ControlHTTPRequest
+)
+
+// NextLocationInfo contains the redirected location as well as the request object which forwards most headers of the initial request.
```
bassosimone, line 37: This line seems a bit too long. Would you mind wrapping it?

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+// This forwarded request is generated by the http.Client and
```
bassosimone, line 38: This sentence ends abruptly and seems truncated.

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+type NextLocationInfo struct {
+	jar             *cookiejar.Jar `json:"-"`
+	location        string         `json:"-"`
+	httpRedirectReq *http.Request  `json:"-"`
+}
+
+type MeasureURLResult struct {
```
bassosimone, line 45: If possible use the same naming of the spec (we can discuss changing names in the spec though).

Also, please, document public structures.

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+		Transport: transport,
+		URL:       URL,
+	})
+	// find out of the host also supports h3 support, which is announced in the Alt-Svc Header
+	h3Support := discoverH3Server(httpMeasurement.HTTPRequest, URL)
+	if h3Support != "" {
+		quicURL, _ := url.Parse(URL.String())
+		quicURL.Scheme = h3Support
+		result.h3Location = quicURL.String()
+	}
```
bassosimone, line 208: I think you could refactor this snippet of code so that the helper function gives you the correct H3 URL to use or an error. (This comment takes into account my previous comment that `discoverH3Server` does not parse many compliant `Alt-Svc` headers and should therefore be extended.)

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+			continue
+		}
+		redirected[req.HTTPRequest] = true
+		urlM, err := MeasureURL(ctx, req, cresp, redirected)
+		if err != nil {
+			return nil, err
+		}
+		cresp.URLMeasurements = append(cresp.URLMeasurements, urlM.CtrlURLMeasurement)
+		nextRequests = append(nextRequests, urlM.redirectedReqs...)
+	}
```
bassosimone, line 87: We need to chat about this logic in person: I am not sure I fully understand it.

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+
+	wg := new(sync.WaitGroup)
+	out := make(chan *MeasureEndpointResult, len(enpnts))
+	for _, endpoint := range enpnts {
+		wg.Add(1)
+		go MeasureEndpoint(ctx, creq, URL, endpoint, wg, out)
+	}
+	wg.Wait()
+	close(out)
```
bassosimone, line 129: It seems to me closing this channel has no effect, am I missing something?

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+		cresp.URLMeasurements = append(cresp.URLMeasurements, urlM.CtrlURLMeasurement)
+		nextRequests = append(nextRequests, urlM.redirectedReqs...)
+	}
+
+	return cresp, nil
+}
+
+// Measure performs the measurement described by the request and
+// returns the corresponding response or an error.
+func MeasureURL(ctx context.Context, creq *CtrlRequest, cresp *CtrlResponse, redirected map[string]bool) (*MeasureURLResult, error) {
```
bassosimone, line 94: The `redirected` variable does not seem to be used by `MeasureURL`, right? If so, then I think we don't need to have this specific parameter for this function.

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+			h3Reqs = append(h3Reqs, req)
+		}
+	}
+	return &MeasureURLResult{CtrlURLMeasurement: urlMeasurement, h3Reqs: h3Reqs, redirectedReqs: redirectedReqs}, nil
+}
+
+func MeasureEndpoint(ctx context.Context, creq *CtrlRequest, URL *url.URL, endpoint string, wg *sync.WaitGroup, out chan *MeasureEndpointResult) {
+	defer wg.Done()
+	endpointResult := &MeasureEndpointResult{}
+	measureFactory[URL.Scheme](ctx, creq, endpoint, wg, endpointResult)
```
bassosimone, line 150: What would happen here in the unlikely event in which `URL.Scheme` is not part of the `measureFactory` map?

bassosimone: I find very confusing this design where the return value is passed by pointer as the last argument. Would it be possible to change this `measureFactory` to return a `MeasureEndpointResult` instead?

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+		Endpoint: endpoint,
+	})
+	if conn == nil {
+		return
+	}
+	defer conn.Close()
+	var transport http.RoundTripper
+	switch URL.Scheme {
+	case "http":
+		transport = nwebconnectivity.GetSingleTransport(conn, nil)
```
bassosimone, line 181: Factories usually in Go are called `NewFooBar` not `GetFooBar`

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+	case "https":
+		var tlsconn *tls.Conn
+		cfg := &tls.Config{ServerName: URL.Hostname()}
```
bassosimone, line 184: I think we want to set ALPN here, right?

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+		return
+	}
+	transport := nwebconnectivity.GetSingleH3Transport(sess, tlscfg, qcfg)
+	h3Measurement.HTTPRequest, result.httpRedirect = HTTPDo(ctx, &HTTPConfig{
+		Jar:       creq.HTTPCookieJar,
+		Headers:   creq.HTTPRequestHeaders,
+		Transport: transport,
+		URL:       URL,
+	})
+	sess.CloseWithError(0, "")
```
bassosimone, line 239: Is this the idiomatic way to close an HTTP3 session? Is this the right error code? (Can't we ask the transport to do this thing for us instead?)

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+	result.CtrlEndpoint = &h3Measurement
+}
+
+func mergeEndpoints(addrs []string, clientAddrs []string) (out []string) {
```
bassosimone, line 243: please document

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+	for key := range unique {
+		out = append(out, key)
+	}
+	return out
+}
+
+// getEndpoints connects IP addresses with the port associated with the URL scheme
+func getEndpoints(addrs []string, URL *url.URL) []string {
+	out := []string{}
+	if URL.Scheme != "http" && URL.Scheme != "https" && URL.Scheme != "h3" && URL.Scheme != "h3-29" {
```
bassosimone, line 260: I think we need an extensible way to add/remove supported QUIC versions: we have a bunch of places where we check for the scheme and it may be more convenient to have a single place where to change that.

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+		panic("passed an unexpected scheme")
+	}
+	p := URL.Port()
+	for _, a := range addrs {
+		var port string
+		switch true {
```
bassosimone, line 266: Couldn't this just be `switch {`?

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,281 @@
...
+			port = "80"
+		case URL.Scheme == "https":
+			port = "443"
+		case URL.Scheme == "h3-29" || URL.Scheme == "h3":
+			port = "443"
+		}
+		endpoint := net.JoinHostPort(a, port)
+		out = append(out, endpoint)
+	}
+	return out
```
bassosimone, line 280: Did you test this function with internationalized input?

internal/cmd/oohelperd/internal/nwcth/tls.go
```diff
@@ -0,0 +1,23 @@
+package nwcth
+
+import (
+	"context"
+	"crypto/tls"
+	"net"
+)
+
+// QUICConfig configures the QUIC handshake check.
```
bassosimone, line 9:
```suggestion
// TLSConfig configures the TLS handshake check.
```

internal/cmd/oohelperd/oohelperd.go
```diff
@@ -10,19 +10,22 @@ import (
...
-	srvwg     = new(sync.WaitGroup)
+	dialer     netx.Dialer
+	endpoint   = flag.String("endpoint", ":8080", "Endpoint where to listen")
+	httpx      *http.Client
+	httpnx     *http.Client
+	quicdialer netx.QUICDialer
+	resolver   netx.Resolver
+	srvcancel  context.CancelFunc
+	srvctx     context.Context
+	srvwg      = new(sync.WaitGroup)
```
bassosimone, line 28: It's unclear to me whether the code you added here still matters. It seems it does not since the only other change below is a new `mux.Handle` line. Am I missing something?

internal/engine/experiment/nwebconnectivity/control.go
```diff
@@ -0,0 +1,124 @@
...
+	"net/http/cookiejar"
+	"net/url"
+
+	"github.com/ooni/probe-cli/v3/internal/engine/httpx"
+	"github.com/ooni/probe-cli/v3/internal/engine/model"
+	"github.com/ooni/probe-cli/v3/internal/errorsx"
+)
+
+// NextLocationInfo contains the redirected location as well as the request object which forwards most headers of the initial request.
+// This forwarded request is generated by the http.Client and
```
bassosimone, line 15: This comment seems truncated.

internal/engine/experiment/nwebconnectivity/control.go
```diff
@@ -0,0 +1,124 @@
...
+type NextLocationInfo struct {
```
bassosimone, line 16: Hey, isn't this structure duplicate?

internal/engine/experiment/nwebconnectivity/control.go
```diff
@@ -0,0 +1,124 @@
+package nwebconnectivity
```
bassosimone, line 1: *Code organization suggestion* since the nwebconnectivity will likely need to use the same primitives of its control implementation, and the control implementation itself is private, how about having most of the code inside of nwebconnectivity and exporting only the bare minimum required by the new control code?

internal/cmd/oohelperd/internal/nwcth/dns.go
```diff
@@ -0,0 +1,39 @@
+package nwcth
+
+import (
+	"context"
+	"net"
+
+	"github.com/apex/log"
+	"github.com/ooni/probe-cli/v3/internal/engine/experiment/nwebconnectivity"
+	"github.com/ooni/probe-cli/v3/internal/engine/netx"
```
bassosimone, line 9: *Dependency question*: I see there is a bunch of places where we still depend on `netx`. Is this dependency functionally required or could it be removed by using `netxlite` + `errorsx`?

bassosimone (changes requested):
Thank you for working to implement the new test helper specification! I think you did a great job so far! My review contains a bunch of design and code organization questions, flags a couple of places where the implementation quality could be improved, and provides other suggestions to improve the documentation and make the code more clear and obvious.

internal/cmd/oohelperd/internal/nwcth/http.go
kelmenhorst: I'm not sure about making Transport optional. I feel like it is more explicit to set the Transport in the call to HTTPDo, as it identifies whether it is a TCP or QUIC based transport. Otherwise, I guess an HTTPS (so TCP) Transport should be the default?

internal/cmd/oohelperd/internal/nwcth/http.go
kelmenhorst: I cannot edit the comment in the reference issue, it seems.

internal/cmd/oohelperd/internal/nwcth/http.go
kelmenhorst: Testing with `-race` says ok. I am wondering whether we even need to store this request. We conserve the cookie jar, so the cookies are taken care of. Other than that we are only passing along "user-agent", "accept", "accept-language" headers. I'm trying to understand whether they could be changed by redirects, but I don't think so.

internal/cmd/oohelperd/internal/nwcth/http.go
kelmenhorst: > I am also failing to see where we are checking whether we should follow the redirect.

This is handled by the `http.Client` (checkout the use of `redirectBehavior` in `http/client.go`). For a 404 + `Location`, the client never lands in the `CheckRedirect` callback.

kelmenhorst: I thought it would be smart to exploit the `http.Client` functionality rather than checking the redirect behavior again. But this requires us to know that `CheckRedirect` was visited.

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: oh, I didn't know, then this is a TODO(@bassosimone) :-)

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: > I am wondering whether we even need to store this request

Good question. My instinct is that probably we don't.

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: A-ha, that's a very nice plan! Though it may be non-obvious, and, in fact, I didn't understand the rationale of doing that. I suggest adding a comment clearly explaining your intent here. If all you need to extract from the callback is just an atomic value, maybe you can just create an `atomicx.Int64` and atomically increment it by 1 if you enter into `CheckRedirect`. Then you can atomically check if it's different from zero to decide whether you entered into the callback or not. (I'd rather have things with atomic semantics in this callback unless the documentation for it explicitly says that it will always be called from the same goroutine of the caller.)

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: Good point! This is not a case where having a default would make things simpler. Let's leave it as is, then!

internal/cmd/oohelperd/internal/nwcth/http.go
kelmenhorst: good idea!

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,110 @@
...
+	scheme := URL.Scheme
+	if strings.Contains(scheme, "h3") {
+		scheme = "https"
+	}
+	return http.NewRequestWithContext(ctx, "GET", scheme+"://"+URL.Hostname(), nil)
+}
+
+// discoverH3Server inspects the Alt-Svc Header of the HTTP (over TCP) response of the control measurement
+// to check whether the server announces to support h3
+func discoverH3Server(r *CtrlHTTPRequest, URL *url.URL) string {
```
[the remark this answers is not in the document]
kelmenhorst: See https://github.com/ooni/probe-cli/commit/81bbd53789efd8fb99ae33118538ca61d3d6da4b . I couldn't figure out a better way to do this than using `strings.Split` a bunch of times... do you know of a better parsing pattern?

internal/cmd/oohelperd/internal/nwcth/internal.go
kelmenhorst: I removed the check from both `internal.go` files.

internal/cmd/oohelperd/internal/nwcth/measure.go
kelmenhorst: A few lines underneath, the channel is iterated. `for m := range out {...}` 
I'm pretty sure that for the loop to terminate, the channel has to be closed.

internal/cmd/oohelperd/internal/nwcth/measure.go
kelmenhorst: > What would happen here in the unlikely event in which URL.Scheme is not part of the measureFactory map?

`MeasureURL` throws an error when the URL Scheme (checkout `getEndpoints()`) is not either "http", "https", "h3", or "h3-29". So I thought it would be safe to assume that this case is covered. Is that sufficient?

internal/cmd/oohelperd/internal/nwcth/measure.go
kelmenhorst: I introduced a map to manage the supported QUIC versions. It lives in `measure.go`

internal/cmd/oohelperd/internal/nwcth/measure.go
kelmenhorst: I tested the complete measurement with `яндекс.рф` and it works. Why is this function in particular critical for internationalized input?

internal/engine/experiment/nwebconnectivity/control.go
kelmenhorst: yes, and I deleted it for now, but we will later need the same thing for `nwebconnectivity` also :D

internal/engine/experiment/nwebconnectivity/control.go
kelmenhorst: what data structures are you referring to?

bassosimone: It seems to me that most of the building blocks are the same and the difference is quite little, but maybe that's not the case. I propose to do nothing for now. If we need to merge/move things, we'll notice.

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: done here: https://github.com/ooni/spec/pull/219/commits/e7ec1e683bd4f1b158a41e632d9f705cd42e2a35

internal/cmd/oohelperd/internal/nwcth/dns.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/ooni/probe-cli/v3/internal/engine/netx/archival"
+)
+
```
bassosimone, line 12:
```suggestion

// TODO(bassosimone,kelmenhorst): figure out if we can _avoid_ using netx here.

```

internal/cmd/oohelperd/internal/nwcth/dns.go
bassosimone: Added comment so we remember about this.

internal/cmd/oohelperd/internal/nwcth/dns.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// newResolver creates a new DNS resolver instance
+func newResolver() netx.Resolver {
+	return netx.NewResolver(netx.Config{Logger: log.Log})
+}
+
+// DNSDo performs the DNS check.
+func DNSDo(ctx context.Context, config *DNSConfig) *DNSMeasurement {
+	if net.ParseIP(config.Domain) != nil {
```
bassosimone, line 34:
```suggestion
	// TODO(bassosimone,kelmenhorst): the underlying resolver probably does that. Is there a
	// way for us to be confident that this is the case and remove this `if` here?
	if net.ParseIP(config.Domain) != nil {
```

internal/cmd/oohelperd/internal/nwcth/fake_test.go
```diff
@@ -0,0 +1,127 @@
...
+
```
bassosimone, line 2:
```suggestion

// TODO(bassosimone,kelmenhorst): can we use `internal/netxmocks` instead?

```

internal/cmd/oohelperd/internal/nwcth/fake_test.go
bassosimone: added comment to remind us about this

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,162 @@
...
+		case "accept-language":
+			for _, v := range vs {
+				req.Header.Add(k, v)
+			}
+		}
+	}
+
+	jar := config.Jar
+	if jar == nil {
+		jar, err = cookiejar.New(nil)
```
bassosimone, line 54:
```suggestion
		jar, err = cookiejar.New(nil) // should not fail
```

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/ooni/probe-cli/v3/internal/atomicx"
+	"github.com/ooni/probe-cli/v3/internal/iox"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
+// HTTPConfig configures the HTTP check.
+type HTTPConfig struct {
+	// Jar contains the optional cookiejar from the previous hop in a redirect chain.
+	Jar http.CookieJar
```
bassosimone, line 21:
```suggestion
	Jar http.CookieJar

```

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,162 @@
...
+	// Headers contains the optional HTTP request headers.
+	Headers map[string][]string
```
bassosimone, line 23:
```suggestion
	Headers map[string][]string

```

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,162 @@
...
+	// Transport contains the mandatory HTTP RoundTripper object.
+	Transport http.RoundTripper
```
bassosimone, line 25:
```suggestion
	Transport http.RoundTripper

```

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,165 @@
...
+		Transport: config.Transport,
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return &HTTPRequestMeasurement{Failure: newfailure(err)}, nil
+	}
+	var httpRedirect *NextLocationInfo
+	loc, err := resp.Location()
+	if shouldRedirect.Load() > 0 && err == nil {
+		loc.Scheme = config.URL.Scheme
```
bassosimone, line 81: We should probably explain why here we're overriding the scheme

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,165 @@
...
+	reader := &io.LimitedReader{R: resp.Body, N: maxAcceptableBody}
+	data, err := iox.ReadAllContext(ctx, reader)
+	return &HTTPRequestMeasurement{
+		BodyLength: int64(len(data)),
+		Failure:    newfailure(err),
+		StatusCode: int64(resp.StatusCode),
+		Headers:    headers,
+	}, httpRedirect
+}
+
```
bassosimone, line 98:
```suggestion

// TODO(bassosimone,kelmenhorst): stuffing the h3 protocol into the scheme, rather than using a
// separate data structure holding the h3 protocol and the new URL, leads to more complex/tricky code,
// so we should probably see whether we can avoid doing that.

```

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,169 @@
...
+		Transport: config.Transport,
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return &HTTPRequestMeasurement{Failure: newfailure(err)}, nil
+	}
+	var httpRedirect *NextLocationInfo
+	loc, err := resp.Location()
+	if shouldRedirect.Load() > 0 && err == nil {
+		loc.Scheme = config.URL.Scheme
```
bassosimone, line 81:
```suggestion
		// This line is here to fix the scheme when we're following h3 redirects. The original URL
		// scheme holds the h3 protocol we're using (h3 or h3-29). As mentioned below, we should
		// find a less tricky solution to this problem, so we can simplify the code.
		loc.Scheme = config.URL.Scheme
```

internal/cmd/oohelperd/internal/nwcth/http.go
bassosimone: Added a comment explaining why we're doing that

internal/cmd/oohelperd/internal/nwcth/http.go
```diff
@@ -0,0 +1,172 @@
...
+	quicURL.Scheme = h3Svc.proto
+	if h3Svc.authorityHost != "" {
+		quicURL.Host = h3Svc.authorityHost
+	}
+	return quicURL, nil
+}
+
+// parseAltSvc parses the Alt-Svc HTTP header for entries advertising the use of H3
+func parseAltSvc(r *HTTPRequestMeasurement, URL *url.URL) *altSvcH3 {
+	if r == nil {
```
bassosimone, line 150:
```suggestion
	// TODO(bassosimone,kelmenhorst): see if we can make this algorithm more robust.
	if r == nil {
```

internal/cmd/oohelperd/internal/nwcth/internal.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/ooni/probe-cli/v3/internal/iox"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+	"github.com/ooni/probe-cli/v3/internal/version"
+)
+
+const maxAcceptableBody = 1 << 24
```
bassosimone, line 14:
```suggestion
// maxAcceptableBody is _at the same time_ the maximum acceptable body for incoming
// API requests and the maximum acceptable body when fetching arbitrary URLs. See
// https://github.com/ooni/probe/issues/1727 for statistics regarding the test lists including
// the empirical cumulative distribution of the body size for test lists URLs.
const maxAcceptableBody = 1 << 24
```

internal/cmd/oohelperd/internal/nwcth/internal.go
```diff
@@ -0,0 +1,53 @@
+package nwcth
```
bassosimone, line 1:
```suggestion
// Package nwcth implements the new web connectivity test helper.
//
// See https://github.com/ooni/spec/blob/master/backends/th-007-nwcth.md
package nwcth
```

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	wg := new(sync.WaitGroup)
+	out := make(chan *MeasureEndpointResult, len(enpnts))
+	for _, endpoint := range enpnts {
+		wg.Add(1)
+		go MeasureEndpoint(ctx, creq, URL, endpoint, wg, out)
+	}
+	wg.Wait()
+	close(out)
```
bassosimone, line 148:
```suggestion
	close(out) // so iterating over it terminates (see below)
```

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,312 @@
...
+			req := &CtrlRequest{HTTPRequest: m.h3Location}
+			h3Reqs = append(h3Reqs, req)
+		}
+	}
+	return &MeasureURLResult{URLMeasurement: urlMeasurement, h3Reqs: h3Reqs, redirectedReqs: redirectedReqs}, nil
+}
+
+func MeasureEndpoint(ctx context.Context, creq *CtrlRequest, URL *url.URL, endpoint string, wg *sync.WaitGroup, out chan *MeasureEndpointResult) {
+	defer wg.Done()
+	// endpointResult := &MeasureEndpointResult{}
```
bassosimone, line 168:
```suggestion
```

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,312 @@
...
+		transport = nwebconnectivity.NewSingleTransport(tlsconn, cfg)
+	}
+	// perform the HTTP request: this provides us with the HTTP request result and info about HTTP redirection
+	httpMeasurement.HTTPRequest, result.httpRedirect = HTTPDo(ctx, &HTTPConfig{
+		Jar:       creq.HTTPCookieJar,
+		Headers:   creq.HTTPRequestHeaders,
+		Transport: transport,
+		URL:       URL,
+	})
+	// find out of the host also supports h3 support, which is announced in the Alt-Svc Header
```
bassosimone, line 227:
```suggestion
	// find out if the host also supports h3 support, which is announced in the Alt-Svc Header
```

internal/engine/experiment/nwebconnectivity/singledialer.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/apex/log"
+	"github.com/lucas-clemente/quic-go"
+	"github.com/lucas-clemente/quic-go/http3"
+	"github.com/ooni/probe-cli/v3/internal/netxlite"
+)
+
+var ErrNoConnReuse = errors.New("cannot reuse connection")
+
+// NewSingleH3Transport creates am http3.RoundTripper
```
bassosimone, line 19:
```suggestion
// NewSingleH3Transport creates an http3.RoundTripper
```

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,312 @@
...
+		if _, ok := redirected[req.HTTPRequest]; ok {
+			continue
+		}
+		redirected[req.HTTPRequest] = true
+		urlM, err := MeasureURL(ctx, req, cresp)
+		if err != nil {
+			continue
+		}
+		cresp.URLMeasurements = append(cresp.URLMeasurements, urlM.URLMeasurement)
+		nextRequests = append(nextRequests, urlM.redirectedReqs...)
```
bassosimone, line 105:
```suggestion
		nextRequests = append(nextRequests, urlM.redirectedReqs...)
		nextRequests = append(nextRequests, urlM.h3Reqs...)
```

internal/cmd/oohelperd/internal/nwcth/measure.go
```diff
@@ -0,0 +1,312 @@
...
+
+	redirected := make(map[string]bool, 100)
+
+	urlM, err := MeasureURL(ctx, creq, cresp)
+	if err != nil {
+		return nil, err
+	}
+	cresp.URLMeasurements = append(cresp.URLMeasurements, urlM.URLMeasurement)
+
+	n := 0
```
bassosimone, line 87:
```suggestion
	// TODO(bassosimone,kelmenhorst): document this algorithm and see if it can be further simplified
	n := 0
```

bassosimone:
Superseded by https://github.com/ooni/probe-cli/pull/432

PR #442, opened 2021-08-18
enable utls for websteps

kelmenhorst:
This diff enables `websteps` to use uTLS for TLS parroting. It integrates the `oohttp.StdlibTransport` wrapper which uses the `ooni/oohttp` fork. `oohttp` supports TLS-like connections like `utls.Conn`.
As a prototype, the testhelper and `websteps` code now uses the `utls.HelloChrome_Auto` fingerprint, i.e. the simulated TLS fingerprint of the Google Chrome browser.

It is a further contribution for my GSoC project.

Reference issue: https://github.com/ooni/probe/issues/1733


go.mod
```diff
@@ -28,6 +28,7 @@ require (
 	github.com/miekg/dns v1.1.42
 	github.com/mitchellh/go-wordwrap v1.0.1
 	github.com/montanaflynn/stats v0.6.6
+	github.com/ooni/oohttp v0.0.0-20210817213129-35949dcf526f // indirect
```
bassosimone, line 31: Why is this indirect? Do you mind running `go mod tidy` and see whether it fixes this line?

internal/engine/experiment/websteps/factory.go
```diff
@@ -64,16 +64,33 @@ func NewSingleH3Transport(qsess quic.EarlySession, tlscfg *tls.Config, qcfg *qui
...
+		Config:        tlsConfig,
+		Dialer:        dialer,
+		TLSHandshaker: handshaker,
+	}).DialTLSContext
+	return transport
+}
 
-	transport = &netxlite.HTTPTransportLogger{Logger: log.Log, HTTPTransport: transport.(*http.Transport)}
+// NewBaseTransport creates a new HTTP transport with the default dialer.
+func NewBaseTransport() (transport *oohttp.StdlibTransport) {
```
bassosimone, line 89: Can you make this function private?

bassosimone (changes requested):
I think there are just two small changes and then we can bless this branch!

Thank you!

internal/cmd/oohelperd/internal/websteps/factory.go
```diff
@@ -1,96 +0,0 @@
-package websteps
```
bassosimone, line 1: For the records: this file has been deleted because a similar functionality was in the websteps experiment

bassosimone (approved):
🐳

PR #441, opened 2021-08-18
Apply go1.17 patches introduced in the release/3.10.1 branch

bassosimone:
## Checklist

- [x] I have read the [contribution guidelines](https://github.com/ooni/probe-cli/blob/master/CONTRIBUTING.md)
- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1738
- [x] related ooni/spec pull request: N/A

Location of the issue tracker: https://github.com/ooni/probe

## Description

This PR contains patches ported from the release/3.10.1 branch. We're working on https://github.com/ooni/probe/issues/1738.

PR #448, opened 2021-08-23
fix: update to lucas-clemente/quic-go v0.23.0

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1739

## Description

This pull request updates to lucas-clemente/quic-go v0.23.0. Updating leads to several CI runtime errors (see, e.g., this failed run: https://github.com/ooni/probe-cli/pull/441#issuecomment-902815439). To avoid these errors, when we are listening for QUIC, we need to return a type that is _both_ a `quic.OOBCapablePacketConn` and `net.Conn`. Otherwise, the underlying `golang.org/x/net/ipv4.NewPacketConn` function is going to panic because we are using a connection that cannot be converted to a `net.Conn`. Therefore, this PR introduces a new `internal/quicx.UDPLikeConn` interface that implements both `quic.OOBCapablePacketConn` and `net.Conn`. See https://github.com/ooni/probe/issues/1739.

It is also worth mentioning that this problem arose in https://github.com/ooni/probe-cli/pull/441, which contained backports from the `stable` branch. Since authoring these changes, I realised it's quite pointless to apply more changes in the stable branch, so I dropped https://github.com/ooni/probe-cli/pull/441/commits/f0d05c54d1db901c3c885cde11199d9678fdeab3 from the patchset.


internal/quicx/quicx.go
```diff
@@ -0,0 +1,32 @@
+// Package quicx contains lucas-clemente/quic-go extensions.
+//
+// This code introduces the UDPLikeConn, whose documentation explain
+// why we need to introduce this new type. We could not put this
+// code inside an existing package because it's used (as of 20 Aug 2021)
+// by the netxlite package as well as by the netx package.
```
bassosimone, line 6:
```suggestion
// by the netxlite package as well as by the netxmocks package.
```

bassosimone:
When running  `urlgetter` using `miniooni` with `HTTP3Enabled=true`, I see this error:

```
2021/08/23 14:23:39 connection doesn't allow setting of receive buffer size. Not a *net.UDPConn?. See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```

It's important to note that we don't see any `read_from` entry as well.

From the above data we can deduce that:

1. our test suite with `-short` does not include enough checks to notice that `read_from` is missing
2. reading `quic-go` code, it seems ~obvious we need to implement `SetReadBuffer`

I am going to open an issue for the former problem and to fix the latter as part of this pull request.

bassosimone:
Even after https://github.com/ooni/probe-cli/pull/448/commits/d587f5a8bc075073d6014a1165cd55a0aec61130, we are still without the `read_from` information in the measurement results. So, it seems this is _another_ issue.

bassosimone:
Superseded by https://github.com/ooni/probe/issues/1754, for which a PR will come soon.

PR #518, opened 2021-09-28
feat(oonimkall): instrument code to understand CI issue

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1785

## Description

It seems ~difficult to reproduce the problem locally and I could not
see it after five runs of

```
go test -race -count 1 ./pkg/oonimkall/...
```

So, here's some diagnostic code that could help understanding the
reference issue https://github.com/ooni/probe/issues/1785.

Also, it seems the issue pops up much more frequently when running
CI anyway. So, I am going to leave this diff around and when it
appears again I have more context to fix the issue.

### DIFF
--- a/pkg/oonimkall/internal/tasks/runner_internal_test.go
+++ b/pkg/oonimkall/internal/tasks/runner_internal_test.go
@@ -54,7 +54,7 @@ func TestRunnerMaybeLookupLocationFailure(t *testing.T) {
 				}
 			case "status.queued", "status.started", "status.end":
 			default:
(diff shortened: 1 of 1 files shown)


bassosimone:
So, this is the error:

```
panic: unexpected key: log - {LogLevel:WARNING Message:sessionresolver: https://doh.powerdns.org/... ooniresolver: refused}
```

I will make the test an integration test for the time being, so we run it less frequently.

PR #520, opened 2021-09-28
fix: allow to build miniooni for windows

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1733

## Description

I need to run test on Windows and I just discovered that:

1. the `errno_unix.go` filename does not mean anything because
`unix` is not a valid platform, so we need a filename for
each platform that we care about;

2. on Windows we need to use WSA prefixed names;

3. `i/e/session_psiphon.go` was not building because of the
migration from `netxlite/iox` to `netxlite`.

This diff attempts to fix all three issues.

The reference issue is https://github.com/ooni/probe/issues/1733,
because I was working on such an issue.


Korb:
> You can run OONI Probe CLI on macOS, Debian/Ubuntu Linux, and on Raspberry Pis.
https://ooni.org/support/ooni-probe-cli

Do I understand correctly that at the moment there is no way to schedule OONI Probe to run on Windows 10 x64 without using a VM with automated post-launch testing?

PR #521, opened 2021-09-29
fix(netxlite): map additional GetAddrInfoW errors

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1467

## Description

On Windows, GetAddrInfoW is a syscall and the Go resolver does
not attempt to map errors beyond WSA_HOST_NOT_FOUND, which becomes
"no such host", which we map to "dns_nxdomain_error".

See https://github.com/golang/go/blob/go1.17.1/src/net/lookup_windows.go#L16.

To map more GetAddrInfoW errors, thus, we need to enhance our
error classifier to have system specific errors.

Then, we need to filter for the WSA errors that are most likely
to pop up and map them to OONI failures. Those are three:

- WSANO_DATA which we have from our own UDP resolver as well
and which we can map to `dns_no_answer`

- WSANO_RECOVERY which we don't have but existed for MK so
we will use `dns_non_recoverable_failure`, which was an MK error

- WSATRY_AGAIN which likewise we map to the error that MK
used to emit, so `dns_temporary_failure`


internal/netxlite/internal/generrno/main.go
```diff
@@ -57,7 +102,6 @@ func (es *ErrorSpec) IsSystemError() bool {
 	return es.errno != ""
 }
 
-// Specs contains all the error specs.
 var Specs = []*ErrorSpec{
```
bassosimone, line 105:
```suggestion
// Specs contains all the error specs.
var Specs = []*ErrorSpec{
```

PR #543, opened 2021-10-13
fix(oohelperd): reduce errors to what the old TH would emit

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1707

## Description

Reducing the errors is not done in a perfect way.

We have documented the most striking differences inside
https://github.com/ooni/probe/issues/1707#issuecomment-942283746 and
some attempts to improve the situation further inside
https://github.com/ooni/probe/issues/1707#issuecomment-942341255.

A better strategy for the future would be to introduce more
specific timeout errors, such as dns_timeout_error, etc.

More testing may be needed to further validate and compare the
old and the new TH, but this requires Jafar improvements to
more precisely simulate more complex censorship.

### DIFF
--- a/internal/cmd/oohelperd/internal/webconnectivity/dns.go
+++ b/internal/cmd/oohelperd/internal/webconnectivity/dns.go
@@ -48,7 +48,7 @@ func dnsMapFailure(failure *string) *string {
 		switch *failure {
 		case netxlite.FailureDNSNXDOMAINError:
 			// We have a name for this string because dnsanalysis.go is
-			// alreayd checking for this specific error string.
+			// already checking for this specific error string.
 			s := webconnectivity.DNSNameError
 			return &s
 		case netxlite.FailureDNSNoAnswer,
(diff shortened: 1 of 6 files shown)


internal/cmd/oohelperd/internal/webconnectivity/dns.go
```diff
@@ -31,5 +32,35 @@ func DNSDo(ctx context.Context, config *DNSConfig) {
...
+// See https://github.com/ooni/backend/blob/6ec4fda5b18/oonib/testhelpers/http_helpers.py#L430
+func dnsMapFailure(failure *string) *string {
+	switch failure {
+	case nil:
+		return nil
+	default:
+		switch *failure {
+		case netxlite.FailureDNSNXDOMAINError:
+			// We have a name for this string because dnsanalysis.go is
+			// alreayd checking for this specific error string.
```
bassosimone, line 51:
```suggestion
			// already checking for this specific error string.
```

PR #563, opened 2021-11-02
refactor(netxlite/filtering): use strings for blocking policies

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1803#issuecomment-957323297

## Description

This change will simplify follow-up work done as part of
https://github.com/ooni/probe/issues/1803#issuecomment-957323297 to
implement a comprehensive self-censoring solution.

While there, rename the "proxy" action to "pass" because what we
are effectively doing is passing traffic to the network (that's a
minor change but it seems a better analogy).

### DIFF
--- a/internal/netxlite/filtering/dns.go
+++ b/internal/netxlite/filtering/dns.go
@@ -17,7 +17,7 @@ import (
 type DNSAction string
 
 const (
-	// DNSActionPass pass the traffic to the upstream server.
+	// DNSActionPass passes the traffic to the upstream server.
 	DNSActionPass = DNSAction("pass")
 
 	// DNSActionNXDOMAIN replies with NXDOMAIN.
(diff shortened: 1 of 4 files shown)


internal/netxlite/filtering/dns.go
```diff
@@ -13,27 +13,27 @@ import (
...
 
-// DNSAction is the action that this proxy should take.
-type DNSAction int
+// DNSAction is a DNS filtering action that this proxy should take.
+type DNSAction string
 
 const (
-	// DNSActionProxy proxies the traffic to the upstream server.
-	DNSActionProxy = DNSAction(iota)
+	// DNSActionPass pass the traffic to the upstream server.
```
bassosimone, line 20:
```suggestion
	// DNSActionPass passes the traffic to the upstream server.
```

PR #566, opened 2021-11-02
feat(filtering): add transparent proxy with censorship policies

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1803#issuecomment-957323297

## Description

This PR implements the core concept described at
https://github.com/ooni/probe/issues/1803#issuecomment-957323297


bassosimone:
Okay, good: as I was expecting, matching strings is more complex on Windows. So, while I remove the `err == nil` blocks I most likely also need to ensure that we convert the errors to the OONI errors and check against that.

PR #583, opened 2021-11-05
feat(QA): add test cases for websteps vs webconnectivity

bassosimone:
## Checklist

- [ ] reference issue for this pull request: https://github.com/ooni/probe/issues/1805
- [ ] related ooni/spec pull request: N/A

## Description

This pull request introduces a set of Node.js scripts for performing A/B comparison of websteps and webconnectivity as described in https://github.com/ooni/probe/issues/1805. Rather than using Jafar, I ended up using `miniooni`'s `--censor` command line flag introduced in [v3.12.0-alpha.1](https://github.com/ooni/probe-cli/releases/tag/v3.12.0-alpha.1). The main reason for doing so is that it's simpler to run tests without requiring root access and Linux (for example, I did not develop part of this diff using Linux). Additionally, I choose to use Node.js rather than extending the existing Python framework for QA, because I found Node.js easier when working with JSON data.


bassosimone:
I can flash merge this PR because the code I've committed is not yet attached to any test.

PR #600, opened 2021-11-16
chore: set version to 3.11.0-beta.2

bassosimone:


bassosimone:
The failure is caused by the ams-pg-test API disk being full. Merging anyway since this PR just bumped the version number.

PR #609, opened 2021-11-22
ci/cd: publish binaries onto a release when we create a tag

bassosimone:
## Checklist

- [x] reference issues for this pull request: https://github.com/ooni/probe/issues/1879, https://github.com/ooni/probe/issues/1875

## Description

This pull request changes `mk` and github workflows to build and publish binaries on tag. We also update the documentation to explain this new branching model. Basically, we have release branches where we produce binary packages and we add extra code, on tag, to publish such packages inside a release.

We discussed removing most secrets from builds in this repository and having a different tool/repository that takes in input also secrets for doing follow-up actions after publishing. As a consequence, this pull request also removes all pieces of code that require secrets. The next step is to reinstate this code in this new repository/tool.

The existing code in `mk` also implemented caching. This feature was useful when doing local builds because it reduced the time required to obtain binary releases. With builds running as part of GitHub actions, we don't need caching because we spawn parallel machines to build binaries. Therefore, let us also remove caching, which makes the code simpler. (Caching in itself is hard and in https://github.com/ooni/probe/issues/1875 I noted that, for example, caching of the `ooni/go` repository was leading to some unwanted behaviour when changing the branch. Without caching, this behaviour is gone and we always generally use fresh information to produce builds.) Of course, this means that local builds are now slower, but I do not think this is a problem _because_ we want to use GitHub actions for building in the common case.

Reference issues: https://github.com/ooni/probe/issues/1879 and https://github.com/ooni/probe/issues/1875.

The final aspect to mention to conclude this description is an implementation one:

```
          gh release create -p $tag --target $GITHUB_SHA || true
```

The code above uses `|| true` because there could already be a release. So, basically, it means that, if a release does not already exist, then we're going to create one. Otherwise, it does not matter because there's already a release.


mk
```diff
@@ -61,17 +55,6 @@ GIT_CLONE_DIR = $(HOME)/.ooniprobe-build/src
...
-#help: * GOLANG_DOCKER_GOCACHE : where to store golang's build cache to
-#help:                           speed up subsequent Docker builds.
-GOLANG_DOCKER_GOCACHE = $(HOME)/.ooniprobe-build/docker/gocache
-
-#help:
-#help: * GOLANG_DOCKER_GOPATH  : GOPATH directory used by builds running
-#help:                           inside docker to significantly speed
-#help:                           up subsequent Docker based builds.
-GOLANG_DOCKER_GOPATH := $(HOME)/.ooniprobe-build/docker/gopath
-
```
bassosimone, line 74: Removing caching since it has no impact on cloud builds speed and "caching is hard"

mk
```diff
@@ -46,12 +46,6 @@ ANDROID_INSTALL_EXTRA = 'build-tools;29.0.3' 'platforms;android-31'
 #help: * ANDROID_NDK_VERSION   : Android NDK version.
 ANDROID_NDK_VERSION = 23.1.7779620
 
-#help:
-#help: * DEBIAN_TILDE_VERSION  : if non-empty, this should be "[0-9]+" and
-#help:                           will be appended to the package version using
-#help:                           a tilde, thus producing, e.g., "1.0~1234".
-DEBIAN_TILDE_VERSION =
-
```
bassosimone, line 54: Debian builds will be done in a separate repository

mk
```diff
@@ -83,13 +66,6 @@ GOLANG_EXTRA_FLAGS =
 #help: * GOLANG_VERSION_NUMBER : the expected version number for golang.
 GOLANG_VERSION_NUMBER = 1.17.3
 
-#help:
-#help: * GPG_USER              : allows overriding the default GPG user used
-#help:                           to sign binary releases, e.g.:
-#help:
-#help:                               ./mk GPG_USER=john@doe.com ooniprobe/windows
-GPG_USER = simone@openobservatory.org
-
```
bassosimone, line 92: We'll be signing in a separate repository

mk
```diff
@@ -671,17 +489,15 @@ OONIPRIVATE_REPO = git@github.com:ooni/probe-private
 
 # $(OONIPRIVATE) clones the private repository in $(GIT_CLONE_DIR)
 $(OONIPRIVATE): search/for/git $(GIT_CLONE_DIR)
-	test -d $(OONIPRIVATE) || $(MAKE) -f mk __really_clone_private_repo
-
-__really_clone_private_repo:
+	test -d $(OONIPRIVATE) && rm -rf $(OONIPRIVATE)
```
bassosimone, line 492: This should be changed to become unconditional removal. We don't exercise this code path in cloud builds but it's nonetheless more robust to use `rm -rf` if this is what we want (and it is).

bassosimone (commented):
We're almost there. I flagged some more required changes before we can merge.

PR #623, opened 2021-12-01
fix(oonimkall): make logger used by tasks unit testable

bassosimone:
Reference issue: https://github.com/ooni/probe/issues/1903


bassosimone:
The test failure is absolutely unrelated (it's in another package), so I'm going to ignore it.

PR #539, opened 2021-10-12
fix: provide good default input for dnscheck and stunreachability

hellais:
## Checklist

- [x] I have read the [contribution guidelines](https://github.com/ooni/probe-cli/blob/master/CONTRIBUTING.md)
- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1814
- [x] companion ooni/spec PR: https://github.com/ooni/spec/pull/227

## Description

This diff aims to solve the problem that dnscheck and stunreachability are not easily runnable from `ooniprobe`. We started with a small diff, and then we needed to hammer other pieces of the codebase to make it WAI.

This diff will need to be backported to the release/3.11 branch. (Also, this diff will work better when using the release/3.11 branch because https://github.com/ooni/probe/issues/1816 and https://github.com/ooni/probe/issues/1808 make the `master` branch not always ideal, but we started on `master` and I'd say we merge on `master` and then backport.)

PR #630, opened 2021-12-03
refactor(stunreachability): input required and must be an URL

bassosimone:
Here we're refactoring stunreachability to not provide internally a
default input and to take in input an URL rather than a string.

The related ooni/spec change is https://github.com/ooni/spec/pull/227.

This diff has been extracted from https://github.com/ooni/probe-cli/pull/539.

Because the original diff was large, I'm splitting it in a set of
more easily manageable diffs.

The reference issue is https://github.com/ooni/probe/issues/1814, which
is complex enough to require us to proceed incrementally.

This diff WILL need to be backported to release/3.11.


internal/engine/experiment/stunreachability/stunreachability_test.go
```diff
@@ -89,22 +119,23 @@ func TestRunCustomInput(t *testing.T) {
...
+	measurement.Input = model.MeasurementTarget(defaultInput)
 	err := measurer.Run(
 		ctx,
 		&mockable.Session{},
 		measurement,
 		model.NewPrinterCallbacks(log.Log),
 	)
-	if err.Error() != "interrupted" {
-		t.Fatal("not the error we expected")
+	if !errors.Is(err, context.Canceled) {
```
bassosimone, line 131: equivalent error but more idiomatic way of checking

PR #446, opened 2021-08-20
feat: relicense the repository under GNU GPLv3

bassosimone:
## Checklist

- [x] I have read the [contribution guidelines](https://github.com/ooni/probe-cli/blob/master/CONTRIBUTING.md)
- [x] reference issue for this pull request: N/A (this is planned maintenance)
- [x] related ooni/spec pull request: N/A

Location of the issue tracker: https://github.com/ooni/probe

## Description

We're going to integrate some pieces of code under the GNU GPLv3
soon, therefore we also need to relicense this repository.

PR #641, opened 2021-12-15
fix(Readme.md): add explicit SPDX identifier

bassosimone:
See https://github.com/ooni/probe-cli/pull/446#issuecomment-995226707


chenrui333 (approved):
Thanks for doing this!

PR #652, opened 2022-01-05
refactor(tor): rewrite using measurex

bassosimone:
This diff rewrites the tor experiment to use measurex "easy" API.

To this end, we need to introduce an "easy" measurex API, which basically
performs easy measurements returning two pieces of data:

1. the resulting measurement, which is already using the OONI
archival data format and is always non-nil

2. a failure (i.e., the pointer to an error string), which
is nil on success and points to a string on failure

With this change, we should now be able to completely dispose of
the original netx API, which was only used by tor.

Reference issue: https://github.com/ooni/probe/issues/1688.

### DIFF
--- a/internal/engine/experiment/tor/tor.go
+++ b/internal/engine/experiment/tor/tor.go
@@ -13,10 +13,8 @@ import (
 	"time"
 
 	"github.com/ooni/probe-cli/v3/internal/atomicx"
-	"github.com/ooni/probe-cli/v3/internal/engine/httpheader"
-	"github.com/ooni/probe-cli/v3/internal/engine/legacy/netxlogger"
-	"github.com/ooni/probe-cli/v3/internal/engine/legacy/oonidatamodel"
-	"github.com/ooni/probe-cli/v3/internal/engine/legacy/oonitemplates"
+	"github.com/ooni/probe-cli/v3/internal/engine/netx/archival"
+	"github.com/ooni/probe-cli/v3/internal/measurex"
 	"github.com/ooni/probe-cli/v3/internal/model"
 	"github.com/ooni/probe-cli/v3/internal/netxlite"
 	"github.com/ooni/probe-cli/v3/internal/runtimex"
@@ -31,7 +29,7 @@ const (
 	testName = "tor"
 
 	// testVersion is the version of this experiment
-	testVersion = "0.3.0"
+	testVersion = "0.4.0"
 )
 
 // Config contains the experiment config.
@@ -44,18 +42,18 @@ type Summary struct {
 
 // TargetResults contains the results of measuring a target.
 type TargetResults struct {
-	Agent          string                          `json:"agent"`
-	Failure        *string                         `json:"failure"`
-	NetworkEvents  oonidatamodel.NetworkEventsList `json:"network_events"`
-	Queries        oonidatamodel.DNSQueriesList    `json:"queries"`
-	Requests       oonidatamodel.RequestList       `json:"requests"`
-	Summary        map[string]Summary              `json:"summary"`
-	TargetAddress  string                          `json:"target_address"`
-	TargetName     string                          `json:"target_name,omitempty"`
-	TargetProtocol string                          `json:"target_protocol"`
-	TargetSource   string                          `json:"target_source,omitempty"`
-	TCPConnect     oonidatamodel.TCPConnectList    `json:"tcp_connect"`
-	TLSHandshakes  oonidatamodel.TLSHandshakesList `json:"tls_handshakes"`
+	Agent          string                                    `json:"agent"`
+	Failure        *string                                   `json:"failure"`
+	NetworkEvents  []*measurex.ArchivalNetworkEvent          `json:"network_events"`
+	Queries        []*measurex.ArchivalDNSLookupEvent        `json:"queries"`
+	Requests       []*measurex.ArchivalHTTPRoundTripEvent    `json:"requests"`
+	Summary        map[string]Summary                        `json:"summary"`
+	TargetAddress  string                                    `json:"target_address"`
+	TargetName     string                                    `json:"target_name,omitempty"`
+	TargetProtocol string                                    `json:"target_protocol"`
+	TargetSource   string                                    `json:"target_source,omitempty"`
+	TCPConnect     []*measurex.ArchivalTCPConnect            `json:"tcp_connect"`
+	TLSHandshakes  []*measurex.ArchivalQUICTLSHandshakeEvent `json:"tls_handshakes"`
 
 	// Only for testing. We don't care about this field otherwise. We
 	// cannot make this private because otherwise the IP address sanitizer
@@ -64,11 +62,11 @@ type TargetResults struct {
 }
 
 func registerExtensions(m *model.Measurement) {
-	oonidatamodel.ExtHTTP.AddTo(m)
-	oonidatamodel.ExtNetevents.AddTo(m)
-	oonidatamodel.ExtDNS.AddTo(m)
-	oonidatamodel.ExtTCPConnect.AddTo(m)
-	oonidatamodel.ExtTLSHandshake.AddTo(m)
+	archival.ExtHTTP.AddTo(m)
+	archival.ExtNetevents.AddTo(m)
+	archival.ExtDNS.AddTo(m)
+	archival.ExtTCPConnect.AddTo(m)
+	archival.ExtTLSHandshake.AddTo(m)
 }
 
 // fillSummary fills the Summary field used by the UI.
@@ -178,10 +176,6 @@ func (m *Measurer) Run(
 	if err != nil {
 		return err
 	}
-	ctx, cancel := context.WithTimeout(
-		ctx, 15*time.Second*time.Duration(len(targets)),
-	)
-	defer cancel()
 	registerExtensions(measurement)
 	m.measureTargets(ctx, sess, measurement, callbacks, targets)
 	return nil
@@ -251,7 +245,7 @@ func (m *Measurer) measureTargets(
 type resultsCollector struct {
 	callbacks       model.ExperimentCallbacks
 	completed       *atomicx.Int64
-	flexibleConnect func(context.Context, keytarget) (oonitemplates.Results, error)
+	flexibleConnect func(context.Context, keytarget) (*measurex.ArchivalMeasurement, *string)
 	measurement     *model.Measurement
 	mu              sync.Mutex
 	sess            model.ExperimentSession
@@ -293,15 +287,16 @@ func maybeSanitize(input TargetResults, kt keytarget) TargetResults {
 func (rc *resultsCollector) measureSingleTarget(
 	ctx context.Context, kt keytarget, total int,
 ) {
-	tk, err := rc.flexibleConnect(ctx, kt)
+	tk, failure := rc.flexibleConnect(ctx, kt)
+	runtimex.PanicIfNil(tk, "measurex should guarantee non-nil here")
 	tr := TargetResults{
 		Agent:         "redirect",
-		Failure:       setFailure(err),
-		NetworkEvents: oonidatamodel.NewNetworkEventsList(tk),
-		Queries:       oonidatamodel.NewDNSQueriesList(tk),
-		Requests:      oonidatamodel.NewRequestList(tk),
-		TCPConnect:    oonidatamodel.NewTCPConnectList(tk),
-		TLSHandshakes: oonidatamodel.NewTLSHandshakesList(tk),
+		Failure:       failure,
+		NetworkEvents: tk.NetworkEvents,
+		Queries:       tk.Queries,
+		Requests:      tk.Requests,
+		TCPConnect:    tk.TCPConnect,
+		TLSHandshakes: tk.TLSHandshakes,
 	}
 	tr.fillSummary()
 	tr = maybeSanitize(tr, kt)
@@ -319,7 +314,7 @@ func (rc *resultsCollector) measureSingleTarget(
 	}
 	rc.callbacks.OnProgress(percentage, fmt.Sprintf(
 		"tor: access %s/%s: %s", kt.maybeTargetAddress(), kt.target.Protocol,
-		errString(err),
+		failureString(failure),
 	))
 }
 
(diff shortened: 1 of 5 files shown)


internal/engine/experiment/tor/tor.go
```diff
@@ -64,11 +62,11 @@ type TargetResults struct {
...
-	oonidatamodel.ExtHTTP.AddTo(m)
-	oonidatamodel.ExtNetevents.AddTo(m)
-	oonidatamodel.ExtDNS.AddTo(m)
-	oonidatamodel.ExtTCPConnect.AddTo(m)
-	oonidatamodel.ExtTLSHandshake.AddTo(m)
+	archival.ExtHTTP.AddTo(m)
+	archival.ExtNetevents.AddTo(m)
+	archival.ExtDNS.AddTo(m)
+	archival.ExtTCPConnect.AddTo(m)
+	archival.ExtTLSHandshake.AddTo(m)
```
bassosimone, line 69: This functionality of filling the extensions is also implemented in the `archival` package.

internal/engine/experiment/tor/tor.go
```diff
@@ -178,10 +176,6 @@ func (m *Measurer) Run(
 	if err != nil {
 		return err
 	}
-	ctx, cancel := context.WithTimeout(
-		ctx, 15*time.Second*time.Duration(len(targets)),
-	)
-	defer cancel()
```
bassosimone, line 184: We don't need an overall timeout for the whole measurement here because each "easy" API call has its own internal timeout, hence it's more effective to let each timeout do its job. The measurement could run for longer but it could also be more precise because it could measure _more_ targets.

internal/engine/experiment/tor/tor.go
```diff
@@ -251,7 +245,7 @@ func (m *Measurer) measureTargets(
 type resultsCollector struct {
 	callbacks       model.ExperimentCallbacks
 	completed       *atomicx.Int64
-	flexibleConnect func(context.Context, keytarget) (oonitemplates.Results, error)
+	flexibleConnect func(context.Context, keytarget) (*measurex.ArchivalMeasurement, *string)
```
bassosimone, line 248: This function is the "engine" implementing the measurement. Interestingly, I only needed to change its return value and implementation to adapt the `tor` test (is this a good sign?).

internal/engine/experiment/tor/tor.go
```diff
@@ -293,15 +287,16 @@ func maybeSanitize(input TargetResults, kt keytarget) TargetResults {
 func (rc *resultsCollector) measureSingleTarget(
 	ctx context.Context, kt keytarget, total int,
 ) {
-	tk, err := rc.flexibleConnect(ctx, kt)
+	tk, failure := rc.flexibleConnect(ctx, kt)
+	runtimex.PanicIfNil(tk, "measurex should guarantee non-nil here")
```
bassosimone, line 291: Here is where I call `flexibleConnect`. Because the "easy" API provides a strong guarantee that it does not return a `nil` structure, I've added an assertion to catch the error. (Having worked a bit with type-aware languages like Rust and Kotlin recently, the treatment of `nil` in Go makes me a bit sad 🤷.)

internal/engine/experiment/tor/tor.go
```diff
@@ -293,15 +287,16 @@ func maybeSanitize(input TargetResults, kt keytarget) TargetResults {
...
-		Queries:       oonidatamodel.NewDNSQueriesList(tk),
-		Requests:      oonidatamodel.NewRequestList(tk),
-		TCPConnect:    oonidatamodel.NewTCPConnectList(tk),
-		TLSHandshakes: oonidatamodel.NewTLSHandshakesList(tk),
+		Failure:       failure,
+		NetworkEvents: tk.NetworkEvents,
+		Queries:       tk.Queries,
+		Requests:      tk.Requests,
+		TCPConnect:    tk.TCPConnect,
+		TLSHandshakes: tk.TLSHandshakes,
```
bassosimone, line 299: Here I am just adapting the filling of `TargetResult`.

internal/engine/experiment/tor/tor.go
```diff
@@ -319,7 +314,7 @@ func (rc *resultsCollector) measureSingleTarget(
 	}
 	rc.callbacks.OnProgress(percentage, fmt.Sprintf(
 		"tor: access %s/%s: %s", kt.maybeTargetAddress(), kt.target.Protocol,
-		errString(err),
+		failureString(failure),
```
bassosimone, line 317: The return value is not an `error` anymore but it's `*string` so I changed the name of the `func` handling it.

internal/engine/experiment/tor/tor.go
```diff
@@ -330,75 +325,59 @@ func maybeScrubbingLogger(input model.Logger, kt keytarget) model.Logger {
...
+		const timeout = 15 * time.Second
+		return mx.EasyHTTPGET(ctx, timeout, URL.String())
 	case "or_port", "or_port_dirauth":
-		r := oonitemplates.TLSConnect(ctx, oonitemplates.TLSConnectConfig{
-			Address:            kt.target.Address,
-			Beginning:          rc.measurement.MeasurementStartTimeSaved,
-			InsecureSkipVerify: true,
-			Handler:            netxlogger.NewHandler(logger),
-		})
-		tk, err = r.TestKeys, r.Error
```
bassosimone, line 364: The address is passed to the "easy" API. The logger and beginning are passed to `mx`. We create below an "easy" TLS config that skips TLS verification.

internal/engine/experiment/tor/tor.go
```diff
@@ -330,75 +325,59 @@ func maybeScrubbingLogger(input model.Logger, kt keytarget) model.Logger {
...
+		return mx.EasyTLSConnectAndHandshake(ctx, kt.target.Address, tlsConfig)
 	case "obfs4":
-		r := oonitemplates.OBFS4Connect(ctx, oonitemplates.OBFS4ConnectConfig{
-			Address:      kt.target.Address,
-			Beginning:    rc.measurement.MeasurementStartTimeSaved,
-			Handler:      netxlogger.NewHandler(logger),
-			Params:       kt.target.Params,
-			StateBaseDir: rc.sess.TempDir(),
-		})
-		tk, err = r.TestKeys, r.Error
```
bassosimone, line 373: The address is still passed below. The beginning and the logger are set into `mx`. The state directory and the params go directly into the "easy" API call, which also has an explicit timeout.

internal/engine/experiment/tor/tor.go
```diff
@@ -330,75 +325,59 @@ func maybeScrubbingLogger(input model.Logger, kt keytarget) model.Logger {
...
+//
+// - tk is the measurement, which is always non nil because
+// the measurex "easy" API provides this guarantee;
+//
+// - failure is nil or an OONI failure string.
+func (rc *resultsCollector) defaultFlexibleConnect(ctx context.Context,
+	kt keytarget) (tk *measurex.ArchivalMeasurement, failure *string) {
+	mx := measurex.NewMeasurerWithDefaultSettings()
+	mx.Begin = rc.measurement.MeasurementStartTimeSaved
+	mx.Logger = maybeScrubbingLogger(rc.sess.Logger(), kt)
```
bassosimone, line 347: The logger configured into the `mx` is a `maybeScrubbingLogger` like it was before.

internal/engine/experiment/tor/tor_test.go
```diff
@@ -26,7 +25,7 @@ func TestNewExperimentMeasurer(t *testing.T) {
 	if measurer.ExperimentName() != "tor" {
 		t.Fatal("unexpected name")
 	}
-	if measurer.ExperimentVersion() != "0.3.0" {
+	if measurer.ExperimentVersion() != "0.4.0" {
```
bassosimone, line 28: Because we changed the minor version.

internal/engine/experiment/tor/tor_test.go
```diff
@@ -258,8 +257,8 @@ func TestResultsCollectorMeasureSingleTargetGood(t *testing.T) {
 		new(model.Measurement),
 		model.NewPrinterCallbacks(log.Log),
 	)
-	rc.flexibleConnect = func(context.Context, keytarget) (oonitemplates.Results, error) {
-		return oonitemplates.Results{}, nil
+	rc.flexibleConnect = func(context.Context, keytarget) (*measurex.ArchivalMeasurement, *string) {
+		return &measurex.ArchivalMeasurement{}, nil
```
bassosimone, line 261: Here I needed to adapt `flexibleConnect` (remember: it always returns a non-nil measurement)

internal/engine/experiment/tor/tor_test.go
```diff
@@ -292,8 +291,9 @@ func TestResultsCollectorMeasureSingleTargetWithFailure(t *testing.T) {
 		new(model.Measurement),
 		model.NewPrinterCallbacks(log.Log),
 	)
-	rc.flexibleConnect = func(context.Context, keytarget) (oonitemplates.Results, error) {
-		return oonitemplates.Results{}, errors.New("mocked error")
+	rc.flexibleConnect = func(context.Context, keytarget) (*measurex.ArchivalMeasurement, *string) {
+		failure := "mocked error"
+		return &measurex.ArchivalMeasurement{}, &failure
```
bassosimone, line 296: Here I needed to adapt `flexibleConnect` (remember: it always returns a non-nil measurement)

internal/engine/experiment/tor/tor_test.go
```diff
@@ -331,14 +331,14 @@ func TestDefautFlexibleConnectDirPort(t *testing.T) {
...
+	tk, failure := rc.defaultFlexibleConnect(ctx, wrapTestingTarget(staticTestingTargets[1]))
+	if failure == nil {
+		t.Fatal("expected a failure here")
 	}
-	if !strings.HasSuffix(err.Error(), "interrupted") {
+	if !strings.HasSuffix(*failure, "interrupted") {
 		t.Fatal("not the error we expected")
 	}
-	if tk.HTTPRequests == nil {
+	if tk.Requests == nil {
```
bassosimone, line 341: The above are (1) changing `err` to `failure` to avoid giving wrong names to variables and (2) changes in the names used by `measurex` to represent some (otherwise equivalent) lists of values.

internal/engine/experiment/tor/tor_test.go
```diff
@@ -353,18 +353,18 @@ func TestDefautFlexibleConnectOrPort(t *testing.T) {
...
+	tk, failure := rc.defaultFlexibleConnect(ctx, wrapTestingTarget(staticTestingTargets[2]))
+	if failure == nil {
+		t.Fatal("expected a failure here")
 	}
-	if err.Error() != "interrupted" {
+	if *failure != "interrupted" {
 		t.Fatal("not the error we expected")
 	}
-	if tk.Connects == nil {
+	if tk.TCPConnect == nil {
```
bassosimone, line 363: The above are (1) changing `err` to `failure` to avoid giving wrong names to variables and (2) changes in the names used by `measurex` to represent some (otherwise equivalent) lists of values.

internal/engine/experiment/tor/tor_test.go
```diff
@@ -378,18 +378,18 @@ func TestDefautFlexibleConnectOBFS4(t *testing.T) {
...
+	tk, failure := rc.defaultFlexibleConnect(ctx, wrapTestingTarget(staticTestingTargets[0]))
+	if failure == nil {
+		t.Fatal("expected a failure here")
 	}
-	if err.Error() != "interrupted" {
+	if *failure != "interrupted" {
 		t.Fatal("not the error we expected")
 	}
-	if tk.Connects == nil {
+	if tk.TCPConnect == nil {
```
bassosimone, line 388: The above are (1) changing `err` to `failure` to avoid giving wrong names to variables and (2) changes in the names used by `measurex` to represent some (otherwise equivalent) lists of values.

internal/engine/experiment/tor/tor_test.go
```diff
@@ -378,18 +378,18 @@ func TestDefautFlexibleConnectOBFS4(t *testing.T) {
...
 		t.Fatal("expected connects data here")
 	}
-	if tk.NetworkEvents == nil {
-		t.Fatal("expected network events data here")
+	if tk.NetworkEvents != nil {
+		t.Fatal("expected no network events data here")
```
bassosimone, line 392: Here's the main change since before: `measurex` only includes the TCPConnect event into the list of TCP connect attempts and _does not_ also include it inside the network events. (I think this change will not have any significant impact on the data processing pipeline, since it ignores network events 🤷.)

internal/measurex/easy.go
```diff
@@ -0,0 +1,266 @@
+package measurex
```
bassosimone, line 1: We need to write tests for the whole of `measurex` before the next release. We need that because we want to ship a tested version of `websteps`. Hence I'm deferring writing tests for now.

internal/measurex/easy.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// EasyTLSConfig helps you to generate a *tls.Config.
+type EasyTLSConfig struct {
+	config *tls.Config
+}
+
+// NewEasyTLSConfig creates a new EasyTLSConfig instance.
+func NewEasyTLSConfig() *EasyTLSConfig {
+	return &EasyTLSConfig{config: &tls.Config{}}
```
bassosimone, line 61: TODO(@bassosimone): I should configure the `RootCAs` here

internal/measurex/easy.go
```diff
@@ -0,0 +1,266 @@
...
+// field will never be a nil pointer);
+//
+// - failure is either nil or a pointer to a OONI failure.
+//
+// Note:
+//
+// - we use the Measurer's TCPConnectTimeout and TLSHandshakeTimeout.
+func (mx *Measurer) EasyTLSConnectAndHandshake(ctx context.Context, endpoint string,
+	tlsConfig *EasyTLSConfig) (meas *ArchivalMeasurement, failure *string) {
+	db := &MeasurementDB{}
```
bassosimone, line 121: TODO(@bassosimone): here I should mention we have a default timeout.

internal/runtimex/runtimex_test.go
```diff
@@ -71,3 +71,25 @@ func TestPanicIfTrue(t *testing.T) {
...
+	badfunc := func(in interface{}, message string) (out error) {
+		defer func() {
+			out = errors.New(recover().(string))
+		}()
+		runtimex.PanicIfNil(in, message)
+		return
+	}
+
+	t.Run("value is not nil", func(t *testing.T) {
+		runtimex.PanicIfNil(false, "antani failed")
```
bassosimone, line 85: TODO(@bassosimone): maybe this `message` parameter is not the best possible

bassosimone (commented):
Explained the changes and left some TODOs for me to address.

PR #656, opened 2022-01-07
cleanup: remove unnecessary legacy interfaces

bassosimone:
This diff addresses another point of https://github.com/ooni/probe/issues/1956:

> - [ ] observe that we're still using a bunch of private interfaces for common interfaces such as the `Dialer`, so we can get rid of these private interfaces and always use the ones in `model`, which allows us to remove a bunch of legacy wrappers

Additional cleanups may still be possible. The more I cleanup, the more I see
there's extra legacy code we can dispose of (which seems good?).


internal/engine/experiment/hhfm/hhfm.go
```diff
@@ -322,7 +321,7 @@ func (d Dialer) DialContext(ctx context.Context, network, address string) (net.C
 	if child == nil {
 		// TODO(bassosimone): figure out why using dialer.New here
 		// causes the experiment to fail with eof_error
-		child = &net.Dialer{Timeout: 15 * time.Second}
+		child = netxlite.DefaultDialer
```
bassosimone, line 324: This is probably good but (1) we should probably construct a netxlite dialer and (2) address the comment or, alternatively, use a `SimpleDialer` and continue using `net.Dialer`.

bassosimone (commented):
Highlight non-obvious hunks in the diff

PR #659, opened 2022-01-10
refactor: move i/netx/archival structs to i/model

bassosimone:
We recently started moving core data structures inside of the
internal/model package as detailed in https://github.com/ooni/probe/issues/1885.

The chief reason to do that is to have a set of fundamental
shared data types to help us rationalize the codebase.

This specific diff moves internal/netx/archival's core data types
inside the internal/model package. While there, it also refactors the
existing tests to improve their quality. Additionally, we also added
an extra test to ensure `ArchivalHTTPBody` is an alias for
`ArchivalMaybeBinaryData`, which is required to ensure the
custom JSON serialization process works for it.

We're doing that because both internal/netx/archival and
internal/measurex define their own archival data structures.

We developed measurex using its own structures because it
allowed to iterate more quickly. Now that we have sketched
out measurex, the time has come to consolidate.

My overall aim is to spend a few more hours this week on
engineering measurex. This work is preliminary work before
we finish up both measurex and websteps.

We described this cleanup in https://github.com/ooni/probe/issues/1957.


internal/model/archival.go
```diff
@@ -0,0 +1,311 @@
...
+// to define a custom JSON encoder that allows us to choose the proper
+// representation depending on whether the Value field is valid UTF-8 or not.
+//
+// See https://github.com/ooni/spec/blob/master/data-formats/df-001-httpt.md#maybebinarydata
+type ArchivalMaybeBinaryData struct {
+	Value string
+}
+
+// MarshalJSON marshals a string-like to JSON following the OONI spec that
+// says that UTF-8 content is represened as string and non-UTF-8 content is
```
bassosimone, line 72:
```suggestion
// says that UTF-8 content is represented as string and non-UTF-8 content is
```

PR #661, opened 2022-01-12
fix(netxlite): robust {ReadAll,Copy}Context with wrapped io.EOF

bassosimone:
Closes https://github.com/ooni/probe/issues/1965


bassosimone:
So, commit https://github.com/ooni/probe-cli/pull/661/commits/e94cfff0dcd6b83068ceecef6ff8af38da4b48fe is the first step of the plan, that is ensuring that we introduce a new test case that fails and shows we've not fully solved the problem yet.

bassosimone:
So, commit https://github.com/ooni/probe-cli/pull/661/commits/f687f29e2f987b6100e4c7030be38f13ba16acb8 has introduced precise failing unit test cases showing that the issue is exactly inside `io.ReadAll` and `io.Copy`.

bassosimone:
Commit https://github.com/ooni/probe-cli/pull/661/commits/8323b56f2496b009a77b878c395ecc9877698ab9 fails to build because we introduced new checks to ensure we're not using `io.ReadAll` or `io.Copy` anywhere in the codebase. We will now introduce a bypass for the only file allowed to use them (`internal/netxlite/iox.go`). This file needs to call them to create safe wrappers.

PR #662, opened 2022-01-12
Refactor the list measurements function to make use of nested queries

hellais:
With a dataset of 1489 test, the ooniprobe list command went from
taking 17.27s to run, to requiring 0.17s or a 100x speed boost

This fixes: https://github.com/ooni/probe/issues/1966

### DIFF
--- a/cmd/ooniprobe/internal/cli/list/list.go
+++ b/cmd/ooniprobe/internal/cli/list/list.go
@@ -1,6 +1,8 @@
 package list
 
 import (
+	"strings"
+
 	"github.com/alecthomas/kingpin"
 	"github.com/apex/log"
 	"github.com/ooni/probe-cli/v3/cmd/ooniprobe/internal/cli/root"
@@ -92,14 +94,17 @@ func init() {
 			netCount := make(map[uint]int)
 			output.SectionTitle("Results")
 			for idx, result := range doneResults {
-				totalCount, anmlyCount, err := database.GetMeasurementCounts(probeCLI.DB(), result.Result.ID)
-				if err != nil {
-					log.WithError(err).Error("failed to list measurement counts")
+				testKeys := "{}"
+
+				// We only care to expose in the testKeys the value of the ndt test result
+				if result.TestGroupName == "performance" {
+					for _, e := range strings.Split(result.TestKeys, "|") {
+						if strings.Contains(e, "download") {
+							testKeys = e
+						}
+					}
 				}
-				testKeys, err := database.GetResultTestKeys(probeCLI.DB(), result.Result.ID)
-				if err != nil {
-					log.WithError(err).Error("failed to get testKeys")
-				}
+
 				output.ResultItem(output.ResultItemData{
 					ID:                      result.Result.ID,
 					Index:                   idx,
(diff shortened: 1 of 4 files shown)


cmd/ooniprobe/internal/database/actions.go
```diff
@@ -103,76 +103,55 @@ func GetMeasurementJSON(sess sqlbuilder.Database, measurementID int64) (map[stri
...
+		db.Raw("results.network_id"),
+		db.Raw("results.result_is_viewed"),
+		db.Raw("results.result_runtime"),
+		db.Raw("results.result_is_done"),
+		db.Raw("results.result_is_uploaded"),
+		db.Raw("results.result_data_usage_up"),
+		db.Raw("results.result_data_usage_down"),
+		db.Raw("results.measurement_dir"),
+
+		db.Raw("COUNT(CASE WHEN measurements.is_anomaly = TRUE THEN 1 END) as anomaly_count"),
```
hellais, line 129: TODO: add tests to check if COUNT() is the right aggregate function or if instead we should be using SUM()

cmd/ooniprobe/internal/cli/list/list.go
```diff
@@ -92,14 +94,17 @@ func init() {
...
-				totalCount, anmlyCount, err := database.GetMeasurementCounts(probeCLI.DB(), result.Result.ID)
-				if err != nil {
-					log.WithError(err).Error("failed to list measurement counts")
-				}
-				testKeys, err := database.GetResultTestKeys(probeCLI.DB(), result.Result.ID)
-				if err != nil {
-					log.WithError(err).Error("failed to get testKeys")
+				testKeys := "{}"
+
+				// We only care to expose in the testKeys the value of the ndt test result
```
bassosimone, line 99: I would suggest to document the way in which we're merging the test keys using `|` and why we're doing that in a more thorough way. Because this is potentially technical debt, it's quite important that we ensure that we understand what are its boundaries, should we change that in the future. To this end I suggest to:

1. expand the comment above to explain what is going on (like why we're collapsing the keys)
2. make sure we mention the same in the place where we are merging the test keys

cmd/ooniprobe/internal/database/actions.go
```diff
@@ -103,76 +103,55 @@ func GetMeasurementJSON(sess sqlbuilder.Database, measurementID int64) (map[stri
...
+		db.Raw("COUNT() as total_count"),
+		db.Raw("group_concat(test_keys, '|') as test_keys"),
```
bassosimone, line 131: Per my previous comment, this line should be marked to refer to the way in which we're merging test keys

bassosimone (changes requested):
The overall diff looks good to me. I think we cannot merge until we figure out whether `COUNT` or `SUM` is the required construct to use. It would also be nice to have a bit more through testing. (For example, it would be nice to check whether the anomaly count is indeed correct, which dissipates any doubt regarding `COUNT` versus `SUM`.)

cmd/ooniprobe/internal/database/actions.go
hellais: I re-read the documentation of the `COUNT()` construct and it states that it will return the number of times the value is non-NULL. This means we are only counting the instances in which `is_anomaly = TRUE`, because in those other cases the `CASE` construct will return NULL.

I have also added a test for this.

cmd/ooniprobe/internal/cli/list/list.go
hellais: I have pushed a commit that enriches these comments.

cmd/ooniprobe/internal/database/actions.go
hellais: I have pushed a commit that enriches these comments.

hellais:
Thanks for the review. I have addressed your comments in the last two commits.

bassosimone (approved):
🐳 thank you!

PR #674, opened 2022-01-25
feat: add a simple dnsping experiment

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1987
- [x] related ooni/spec pull request: https://github.com/ooni/spec/pull/238

## Description

This experiment will ping the URLs you pass it in input. Each URL
is like udp://<host>:<port> and represents an UDP resolver.

It will run 10 repetitions by default. The default domain we're
trying to resolve is edge-chat.instagram.com.

You can change both settings by passing `-O` with miniooni.

The reference issues are https://github.com/ooni/probe/issues/1497 and https://github.com/ooni/probe/issues/1987.

Here's an example of usage:

```
./miniooni -i udp://8.8.8.8:53 -i udp://1.1.1.1:53 dnsping
```

Here's another example where we change the domain we're pinging:

```
./miniooni -ORepetitions=3 -i udp://8.8.8.8:53 -i udp://1.1.1.1:53 dnsping
```

Here's another example where we change the domain we resolve:

```
./miniooni -ODomain=x.org -i udp://8.8.8.8:53 -i udp://1.1.1.1:53 dnsping
```


internal/engine/experiment/dnsping/dnsping.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if parsed.Scheme != "udp" {
+		return errors.New("we only support udp://<host>:<port> for now")
+	}
+	tk := new(TestKeys)
+	measurement.TestKeys = tk
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	for i := int64(0); i < m.config.repetitions(); i++ {
+		mx := measurex.NewMeasurerWithDefaultSettings()
```
bassosimone, line 91: Here I think I should use the same `mx` for the whole session

PR #675, opened 2022-01-26
fix(reduceErrors): return error when given an empty list

bassosimone:
See https://github.com/ooni/probe/issues/1985 for context.

While there, ensure nextlite has 100% of coverage.


bassosimone:
(So, I know that the ams-pg-test server is down for maintenance, which means I expect `coverage` to fail.)

bassosimone:
So, as predicted `coverage` failed. But this change only modifies unit tests and these unittests are in `netxlite`, which completed successfully. Also, the same tests are passing also locally. Hence, Imma merge!

PR #677, opened 2022-01-27
feat: add an experimental quicping experiment

kelmenhorst:
This experiment pings a QUIC-able host. It can be used to measure QUIC availability independently from TLS.
This is the reference issue: https://github.com/ooni/probe/issues/1994

### A QUIC PING is:
- a QUIC Initial packet with a size of 1200 bytes (minimum datagram size defined in the [RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#initial-size)),
- with a random payload (i.e. no TLS ClientHello),
- with the version string 0xbabababa which forces Version Negotiation at the server.

QUIC-able hosts respond to the QUIC PING with a Version Negotiation packet.

The input is a domain name or an IP address. The default port used by quicping is 443, as this is the port used by HTTP/3. The port can be modified with the `-O Port=` option.
The default number of repetitions is 10, it can be changed with `-O Repetitions=`.

### Usage:
```
./miniooni -i google.com quicping
./miniooni -i 142.250.181.206 quicping
./miniooni -i 142.250.181.206 -OPort=443 quicping
./miniooni -i 142.250.181.206 -ORepetitions=2 quicping

```

### DIFF
--- a/internal/engine/experiment/quicping/crypto.go
+++ b/internal/engine/experiment/quicping/crypto.go
@@ -108,7 +108,7 @@ func encryptHeader(raw, hdr, clientSecret []byte) []byte {
 // https://www.rfc-editor.org/rfc/rfc9001.html#name-packet-protection
 //
 // encryptPayload encrypts the payload of the packet.
-func encryptPayload(payload, destConnID ConnectionID, clientSecret []byte) []byte {
+func encryptPayload(payload, destConnID connectionID, clientSecret []byte) []byte {
 	myKey, myIV := computeInitialKeyAndIV(clientSecret)
 	encrypter := aeadAESGCMTLS13(myKey, myIV)
 
--- a/internal/engine/experiment/quicping/crypto.go
+++ b/internal/engine/experiment/quicping/crypto.go
@@ -1,0 +1,167 @@
+package quicping
+
+import (
+	"crypto"
+	"crypto/aes"
+	"crypto/cipher"
+	"encoding/binary"
+	"fmt"
+
+	"golang.org/x/crypto/hkdf"
+)
+
+// https://github.com/marten-seemann/qtls-go1-15/blob/0d137e9e3594d8e9c864519eff97b323321e5e74/cipher_suites.go#L281
+type aead interface {
+	cipher.AEAD
+
+	// explicitNonceLen returns the number of bytes of explicit nonce
+	// included in each record. This is eight for older AEADs and
+	// zero for modern ones.
+	explicitNonceLen() int
+}
+
+const (
+	aeadNonceLength   = 12
+	noncePrefixLength = 4
+)
+
+// https://github.com/marten-seemann/qtls-go1-15/blob/0d137e9e3594d8e9c864519eff97b323321e5e74/cipher_suites.go#L375
+func aeadAESGCMTLS13(key, nonceMask []byte) aead {
+	if len(nonceMask) != aeadNonceLength {
+		panic("tls: internal error: wrong nonce length")
+	}
+	aes, err := aes.NewCipher(key)
+	if err != nil {
+		panic(err)
+	}
+	aead, err := cipher.NewGCM(aes)
+	if err != nil {
+		panic(err)
+	}
+	ret := &xorNonceAEAD{aead: aead}
+	copy(ret.nonceMask[:], nonceMask)
+	return ret
+}
+
+// computeInitialKeyAndIV derives the packet protection key and Initialization Vector (IV)
+// from the initial secret.
+// https://www.rfc-editor.org/rfc/rfc9001.html#protection-keys
+func computeInitialKeyAndIV(secret []byte) (key, iv []byte) {
+	key = hkdfExpandLabel(crypto.SHA256, secret, []byte{}, "quic key", 16)
+	iv = hkdfExpandLabel(crypto.SHA256, secret, []byte{}, "quic iv", 12)
+	return
+}
+
+// computeHP derives the header protection key from the initial secret.
+// https://www.rfc-editor.org/rfc/rfc9001.html#protection-keys
+func computeHP(secret []byte) (hp []byte) {
+	hp = hkdfExpandLabel(crypto.SHA256, secret, []byte{}, "quic hp", 16)
+	return
+}
+
+// computeSecrets computes the initial secrets based on the destination connection ID.
+// https://www.rfc-editor.org/rfc/rfc9001.html#name-initial-secrets
+func computeSecrets(destConnID []byte) (clientSecret, serverSecret []byte) {
+	initialSalt := []byte{0x38, 0x76, 0x2c, 0xf7, 0xf5, 0x59, 0x34, 0xb3, 0x4d, 0x17, 0x9a, 0xe6, 0xa4, 0xc8, 0x0c, 0xad, 0xcc, 0xbb, 0x7f, 0x0a}
+	// initial_secret = HKDF-Extract(initial_salt,client_dst_connection_id)
+	initialSecret := hkdf.Extract(crypto.SHA256.New, destConnID, initialSalt)
+	// client_initial_secret = HKDF-Expand-Label(initial_secret, "client in", "", 32) = c00cf151ca5be075ed0ebfb5c80323c42d6b7db67881289af4008f1f6c357aea
+	clientSecret = hkdfExpandLabel(crypto.SHA256, initialSecret, []byte{}, "client in", crypto.SHA256.Size())
+	serverSecret = hkdfExpandLabel(crypto.SHA256, initialSecret, []byte{}, "server in", crypto.SHA256.Size())
+	return
+}
+
+// encryptHeader applies header protection to the packet bytes (raw).
+// https://www.rfc-editor.org/rfc/rfc9001.html#name-client-initial
+// https://www.rfc-editor.org/rfc/rfc9001.html#name-header-protection
+func encryptHeader(raw, hdr, clientSecret []byte) []byte {
+	hp := computeHP(clientSecret)
+	block, err := aes.NewCipher(hp)
+	if err != nil {
+		panic(fmt.Sprintf("error creating new AES cipher: %s", err))
+	}
+	hdroffset := 0
+	payloadOffset := len(hdr)
+	sample := raw[payloadOffset : payloadOffset+16]
+
+	mask := make([]byte, block.BlockSize())
+	if len(sample) != len(mask) {
+		panic("invalid sample size")
+	}
+	block.Encrypt(mask, sample)
+
+	pnOffset := len(hdr) - 4
+	pnBytes := raw[pnOffset:payloadOffset]
+	raw[hdroffset] ^= mask[0] & 0xf
+	for i := range pnBytes {
+		pnBytes[i] ^= mask[i+1]
+	}
+	return raw
+}
+
+// encryptPayload encrypts the payload of the packet.
+// https://www.rfc-editor.org/rfc/rfc9001.html#name-packet-protection
+func encryptPayload(payload, destConnID ConnectionID, clientSecret []byte) []byte {
+	myKey, myIV := computeInitialKeyAndIV(clientSecret)
+	encrypter := aeadAESGCMTLS13(myKey, myIV)
+
+	nonceBuf := make([]byte, encrypter.NonceSize())
+	var pn int64 = 2
+	binary.BigEndian.PutUint64(nonceBuf[len(nonceBuf)-8:], uint64(pn))
+
+	encrypted := encrypter.Seal(nil, nonceBuf, payload, nil)
+	return encrypted
+}
+
+// hkdfExpandLabel HKDF expands a label.
+// https://github.com/lucas-clemente/quic-go/blob/master/internal/handshake/hkdf.go
+func hkdfExpandLabel(hash crypto.Hash, secret, context []byte, label string, length int) []byte {
+	b := make([]byte, 3, 3+6+len(label)+1+len(context))
+	binary.BigEndian.PutUint16(b, uint16(length))
+	b[2] = uint8(6 + len(label))
+	b = append(b, []byte("tls13 ")...)
+	b = append(b, []byte(label)...)
+	b = b[:3+6+len(label)+1]
+	b[3+6+len(label)] = uint8(len(context))
+	b = append(b, context...)
+
+	out := make([]byte, length)
+	n, err := hkdf.Expand(hash.New, secret, b).Read(out)
+	if err != nil || n != length {
+		panic("quic: HKDF-Expand-Label invocation failed unexpectedly")
+	}
+	return out
+}
+
+// xoredNonceAEAD wraps an AEAD by XORing in a fixed pattern to the nonce before each call.
+// https://github.com/marten-seemann/qtls-go1-15/blob/0d137e9e3594d8e9c864519eff97b323321e5e74/cipher_suites.go#L319
+type xorNonceAEAD struct {
+	nonceMask [aeadNonceLength]byte
+	aead      cipher.AEAD
+}
+
+func (f *xorNonceAEAD) NonceSize() int        { return 8 } // 64-bit sequence number
+func (f *xorNonceAEAD) Overhead() int         { return f.aead.Overhead() }
+func (f *xorNonceAEAD) explicitNonceLen() int { return 0 }
+
+func (f *xorNonceAEAD) Seal(out, nonce, plaintext, additionalData []byte) []byte {
+	for i, b := range nonce {
+		f.nonceMask[4+i] ^= b
+	}
+	result := f.aead.Seal(out, f.nonceMask[:], plaintext, additionalData)
+	for i, b := range nonce {
+		f.nonceMask[4+i] ^= b
+	}
+	return result
+}
+
+func (f *xorNonceAEAD) Open(out, nonce, ciphertext, additionalData []byte) ([]byte, error) {
+	for i, b := range nonce {
+		f.nonceMask[4+i] ^= b
+	}
+	result, err := f.aead.Open(out, f.nonceMask[:], ciphertext, additionalData)
+	for i, b := range nonce {
+		f.nonceMask[4+i] ^= b
+	}
+	return result, err
+}
--- a/internal/engine/experiment/quicping/crypto.go
+++ b/internal/engine/experiment/quicping/crypto.go
@@ -7,6 +7,7 @@ import (
 	"encoding/binary"
 	"fmt"
 
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
 	"golang.org/x/crypto/hkdf"
 )
 
@@ -31,13 +32,9 @@ func aeadAESGCMTLS13(key, nonceMask []byte) aead {
 		panic("tls: internal error: wrong nonce length")
 	}
 	aes, err := aes.NewCipher(key)
-	if err != nil {
-		panic(err)
-	}
+	runtimex.PanicOnError(err, fmt.Sprintf("aes.NewCipher failed: %s", err))
 	aead, err := cipher.NewGCM(aes)
-	if err != nil {
-		panic(err)
-	}
+	runtimex.PanicOnError(err, fmt.Sprintf("cipher.NewGCM failed: %s", err))
 	ret := &xorNonceAEAD{aead: aead}
 	copy(ret.nonceMask[:], nonceMask)
 	return ret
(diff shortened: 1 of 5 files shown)


bassosimone (commented):
We did a bit of live review with Kathrin and here are some comments, more to come.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,319 @@
...
+// buildPacket constructs an Initial QUIC packet
+// and applies Initial protection.
+// https://www.rfc-editor.org/rfc/rfc9001.html#name-client-initial
+func buildPacket() ([]byte, ConnectionID, ConnectionID) {
+	destConnID, srcConnID := generateConnectionIDs()
+	// generate random payload
+	minPayloadSize := 1200 - 14 - (len(destConnID) + len(srcConnID))
+	randomPayload := make([]byte, minPayloadSize)
+	random.Seed(time.Now().UnixNano())
+	random.Read(randomPayload)
```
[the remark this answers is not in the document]
kelmenhorst: I decided to use `crypto/rand` rather than `math/rand` to generate random byte arrays.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,395 @@
...
+	for _, req := range sendInfoMap {
+		tk.Pings = append(tk.Pings, &SinglePing{
+			ConnIdDst:         req.dstID,
+			ConnIdSrc:         req.srcID,
+			Failure:           archival.NewFailure(timeoutErr),
+			Request:           &model.ArchivalMaybeBinaryData{Value: string(req.raw)},
+			Response:          nil,
+			SupportedVersions: nil,
+		})
+	}
```
bassosimone, line 243: So, with the changes I recommended above, the main loop becomes something like:

```Go
for {
  select {
  case ping := <-pingch:
    // store the ping info into the map (no mutex needed since the map is owned by Run)
  case pong := <-pongch:
    // figure out if we have a corresponding ping (again no mutex needed)
  case <-ctx.Done():
    break
  }
}
```

bassosimone (commented):
Thanks for exploring ways to separate the sender and the receiver. I have mostly focused this review in trying to explain how I would further refactor the code to use a pattern where we have independent goroutines cooperating through channels. In addition to that, I have also tried to provide some smaller refactoring hints. It seems my proposal to refactor the code does not provide a clear model of how to handle UDP errors, mostly because I am not sure which specific UDP errors could actually occur here. This last part is something we should probably chat about: I suspect I am missing some details that do not help me see which errors could happen and chatting is probably faster.

internal/engine/experiment/quicping/crypto.go
```diff
@@ -0,0 +1,180 @@
...
+	"crypto/aes"
+	"crypto/cipher"
+	"encoding/binary"
+	"fmt"
+
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+	"golang.org/x/crypto/hkdf"
+)
+
+// BSD 3-Clause "New" or "Revised" License
```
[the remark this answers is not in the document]
kelmenhorst: Ah, sorry, I misunderstood you there, when we talked about it earlier.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,396 @@
...
+	Repetitions int64 `ooni:"number of times to repeat the measurement"`
+
+	// Port is the port to test.
+	Port int64 `ooni:"port is the port to test"`
+
+	// Timeout is the number of milliseconds to wait for the ping response
+	Timeout int64 `ooni:"Timeout is the number of milliseconds to wait for the ping response"`
+
+	// NetworkLibrary is the underlying network library. Can be used for testing.
+	NetworkLibrary model.UnderlyingNetworkLibrary
```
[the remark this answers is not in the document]
kelmenhorst: That depends on the package of the test code, right? I have often seen the pattern in the code base where you name the test package of `package x`,  `package x_test` which makes it necessary for test fields to be public. I'm not sure I understood what the advantage of using a new package for testing is though, so maybe I'm missing something.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,396 @@
...
+	RequestTime       string                         `json:"request_time"`
+	Response          *model.ArchivalMaybeBinaryData `json:"response"`
+	ResponseTime      string                         `json:"response_time"`
+	SupportedVersions []uint32                       `json:"supported_versions"`
+}
+
+// Measurer performs the measurement.
+type Measurer struct {
+	config Config
+	mu     sync.Mutex
```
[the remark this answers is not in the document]
kelmenhorst: it's protecting the `sendInfoMap` used to remember sent ping requests

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,395 @@
...
+				})
+				continue
+			}
+			m.mu.Lock()
+			sendInfoMap[hex.EncodeToString(srcID)] = &sendInfo{dstID: dstID, srcID: srcID, raw: sent, sendTime: sendTime}
+			m.mu.Unlock()
+			<-ticker.C
+		}
+		wg.Done()
+	}()
```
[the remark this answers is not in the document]
kelmenhorst: Thanks for this explanation! I get what you mean and I agree.
The only thing that I am not too convinced about is the channel. I have been thinking about this quite a bit when refactoring and though I also started out using a channel to share ping information, I ended up using a map. My reasoning for this was that it is a lot easier to lookup the corresponding ping request. Especially with respect to the scenario with an injected or lost response.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,443 @@
...
+			packet, dstID, srcID := buildPacket()     // build QUIC Initial packet
+			_, err := pconn.WriteTo(packet, destAddr) // send Initial packet
+
+			sess.Logger().Infof("PING %s", destAddr)
+
+			// propagate send information, including errors
+			out <- requestInfo{raw: packet, t: sendTime, dstID: hex.EncodeToString(dstID), srcID: hex.EncodeToString(srcID), err: err}
+			i += 1
+
+		default:
```
bassosimone, line 157:
```suggestion
```

Having a `default` arm here strikes me as incorrect. It means you are not blocked in either of the two arms of the `select` and instead you are busy looping. I think you should double check the CPU utilization during the experiment and then remove this `default` arm of the `select` statement.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,443 @@
...
+				SupportedVersions: resp.versions,
+			})
+			if len(tk.Pings) >= int(rep) {
+				break L // break when we collected all ping responses
+			}
+
+		case <-ctx.Done():
+			break L
+
+		default:
```
bassosimone, line 287:
```suggestion
```

Having a default arm here strikes me as incorrect. It means you are not blocked in either of the two arms of the select and instead you are busy looping. I think you should double check the CPU utilization during the experiment and then remove this default arm of the select statement.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,443 @@
...
+		case resp := <-recvInfoChan: // a new response has been received
+			var (
+				req *requestInfo
+				ok  bool
+			)
+			// match response to request
+			if req, ok = sendMap[resp.dstID]; !ok {
+				continue // ignore any version negotiation response with an unknown destination ID
+			}
+			delete(sendMap, resp.dstID)
```
bassosimone, line 269: I suggest to change the structure of the experiment as follows:

1. you allow more than a single response per ping
2. you keep the `sendMap` until the end
3. you store every reply into the `sendMap` by appending to the relevant ping
4. you include the case you are ignoring in the `receiver` in the results by returning an error
5. when you are done, you walk through the send map and assign `timeout` to every response-less entry
6. finally you transform the send map into a list of pings

This refactoring would allow us to detect multiple replies and spurious replies.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,443 @@
...
+	// create UDP socket
+	pconn, err := m.config.networkLibrary().ListenUDP("udp", &net.UDPAddr{})
+	if err != nil {
+		return err
+	}
+	defer pconn.Close()
+
+	// set context and read timeouts
+	deadline := time.Duration(rep*5) * time.Second
+	pconn.SetReadDeadline(time.Now().Add(deadline))
```
bassosimone, line 233: I don't think it could happen in practice that we're blocked on `Write`. Though, to be on the safe side, I'd really  use `SetDeadline` here rather than `SetReadDeadline` so we know the `pconn` will eventually timeout.

(Maybe it's not needed but it's more robust to future refactorings.)

```suggestion
	pconn.SetDeadline(time.Now().Add(deadline))
```

bassosimone (changes requested):
Thanks for refactoring the code! I really like where this diff is going! 💯 

I have requested a few changes to improve robustness and data accuracy.

internal/engine/experiment/quicping/quicping.go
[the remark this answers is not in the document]
bassosimone: Yes, I used to use the pattern you refer to. That's mostly used on old code. In new code I'm mostly using internal testing (i.e., where there is no `package x_test` and tests list in `package x`). The advantage of external testing (i.e., `package x_test`) is that you are only testing the external accessible interface. From the point of view of accuracy, you're really only testing what can happen. However, sometimes this leads to more complicate APIs for the package itself. Therefore, I'd rather have internal testing where still we try to only test from the point of view of someone using the package, where this is possible. It's a compromise between purity and speed, I think.

internal/engine/experiment/quicping/quicping.go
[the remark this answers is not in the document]
bassosimone: I see. Then, I'd say it's important to remember to put a mutex protecting some resource in the same scope of the resource itself. For example, in the original code, the map was on the stack and the mutex should also have been on the stack. We don't want two related resources to have two potentially different lifetimes.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	_ "crypto/sha256"
+
+	"github.com/ooni/probe-cli/v3/internal/engine/netx/archival"
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/netxlite"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
+// A ConnectionID in QUIC
+type ConnectionID []byte
```
bassosimone, line 28:
```suggestion
// A ConnectionID in QUIC
type connectionID []byte
```

This type should be private

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	pconn model.UDPLikeConn,
+	destAddr *net.UDPAddr,
+	out chan<- requestInfo,
+	sess model.ExperimentSession,
+) {
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+
+	i := 0
+	for i < int(m.config.repetitions()) {
```
bassosimone, line 169:
```suggestion
	for i := int64(0); i < m.config.repetitions(); i++ {
```

Let's use the more compact for notation here and let's not downcast from `int64` to `int`.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+
+		case sendTime := <-ticker.C:
+			packet, dstID, srcID := buildPacket()     // build QUIC Initial packet
+			_, err := pconn.WriteTo(packet, destAddr) // send Initial packet
+
+			sess.Logger().Infof("PING %s", destAddr)
+
+			// propagate send information, including errors
+			out <- requestInfo{raw: packet, t: sendTime, dstID: hex.EncodeToString(dstID), srcID: hex.EncodeToString(srcID), err: err}
+			i += 1
```
bassosimone, line 182:
```suggestion
```

Line not needed if we use the `for` notation.

bassosimone: Also, I think here we should handle the `errors.Is(err, net.ErrClosed)` case by returning

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+				break
+			}
+			// store read failures and continue receiving
+			out <- responseInfo{t: respTime, err: err}
+			continue
+		}
+		resp := buffer[:n]
+
+		// dissect server response
+		supportedVersions, dst, err := m.DissectVersionNegotiation(resp)
```
bassosimone, line 212:
```suggestion
		supportedVersions, dst, err := m.dissectVersionNegotiation(resp)
```

This method should be private

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	for {
+		select {
+		case req := <-sendInfoChan: // a new ping was sent
+			if req.err != nil {
+				tk.Pings = append(tk.Pings, &SinglePing{
+					ConnIdDst:   req.dstID,
+					ConnIdSrc:   req.srcID,
+					Failure:     archival.NewFailure(req.err),
+					Request:     &model.ArchivalMaybeBinaryData{Value: string(req.raw)},
+					RequestTime: formatTime(req.t),
```
bassosimone, line 280: I think we should represent the time like we do for other nettests, i.e., as seconds elapsed since the `MeasurementStartTimeSaved` field of the measurement. This applies to both requests and responses: we should be using `time.Since(t0).Seconds()` every where we want to store a time. (In the previous snippet `t0` is the moment where a measurement was started (i.e., `MeasurementStartTimeSaved`.)

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+			break L
+		}
+	}
+	// transform ping requests into TestKeys.Pings
+	timeoutErr := errors.New("i/o timeout")
+	for _, ping := range pingMap {
+		if ping.request == nil { // this should not happen
+			return errors.New("internal error: ping.request is nil")
+		}
+		if len(ping.responses) == 0 {
```
bassosimone, line 314:
```suggestion
		if len(ping.responses) <= 0 {
```

I know `len` cannot return a negative value by design but I'd still rather to write robust code. Since `len` in principle could return an `int`, I'd like to see `<=` being used here.

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+		})
+	}
+	return nil
+}
+
+// DissectVersionNegotiation dissects the Version Negotiation response.
+// It returns the supported versions and the destination connection ID of the response,
+// The destination connection ID of the response has to coincide with the source connection ID of the request.
+// https://www.rfc-editor.org/rfc/rfc9000.html#name-version-negotiation-packet
+func (m *Measurer) DissectVersionNegotiation(i []byte) ([]uint32, ConnectionID, error) {
```
bassosimone, line 344: Please, move this function to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+}
+
+// GetSummaryKeys implements model.ExperimentMeasurer.GetSummaryKeys.
+func (m *Measurer) GetSummaryKeys(measurement *model.Measurement) (interface{}, error) {
+	return SummaryKeys{IsAnomaly: false}, nil
+}
+
+// buildHeader creates the unprotected QUIC header.
+// https://www.rfc-editor.org/rfc/rfc9000.html#name-initial-packet
+func buildHeader(destConnID, srcConnID ConnectionID, payloadLen int) []byte {
```
bassosimone, line 394: Please, move this function to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	binary.BigEndian.PutUint32(pn, uint32(2))
+	hdr = append(hdr, pn...) // packet number
+
+	return hdr
+}
+
+// buildPacket constructs an Initial QUIC packet
+// and applies Initial protection.
+// https://www.rfc-editor.org/rfc/rfc9001.html#name-client-initial
+func buildPacket() ([]byte, ConnectionID, ConnectionID) {
```
bassosimone, line 428: Please, move this function to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	encrypted := encryptPayload(randomPayload, destConnID, clientSecret)
+	hdr := buildHeader(destConnID, srcConnID, len(encrypted))
+	raw := append(hdr, encrypted...)
+
+	raw = encryptHeader(raw, hdr, clientSecret)
+	return raw, destConnID, srcConnID
+}
+
+// generateConnectionID generates a connection ID using cryptographic random
+func generateConnectionID(len int) ConnectionID {
```
bassosimone, line 445: Please, move this function to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	b := make([]byte, len)
+	_, err := rand.Read(b)
+	runtimex.PanicOnError(err, "rand.Read failed")
+	return ConnectionID(b)
+}
+
+// generateConnectionIDForInitial generates a connection ID for the Initial packet.
+// It uses a length randomly chosen between 8 and 18 bytes.
+func generateConnectionIDForInitial() ConnectionID {
```
bassosimone, line 454: Please, move this function to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	r := make([]byte, 1)
+	_, err := rand.Read(r)
+	runtimex.PanicOnError(err, "rand.Read failed")
+	len := minConnectionIDLenInitial + int(r[0])%(maxConnectionIDLen-minConnectionIDLenInitial+1)
+	return generateConnectionID(len)
+}
+
+// generateConnectionIDs generates a destination and source connection ID.
+func generateConnectionIDs() ([]byte, []byte) {
```
bassosimone, line 463: Please, move this function to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

internal/engine/experiment/quicping/quicping.go
```diff
@@ -0,0 +1,479 @@
...
+	destConnID := generateConnectionIDForInitial()
+	srcConnID := generateConnectionID(defaultConnectionIDLength)
+	return destConnID, srcConnID
+}
+
+// errUnexpectedResponse is thrown when the response from the server
```
bassosimone, line 469: Please, move this type to a separate file. Logically, this is a basic building block for the main implementation of the experiment, so I'd like it to be in a separate file. The way I see it, our stack here is like:

```
  experiment (quicping.go)
  --------
  separate file (maybe quic.go?)
  --------
  crypto (crypto.go)

```

bassosimone (changes requested):
Thanks! I'd like to request a few more changes, then for me we're good to go!

I am happy to see this nettest is about to be merged! 💯 🔝 🙏 🥳

internal/engine/experiment/quicping/quicping.go
bassosimone: *correction*: we should handle this case before propagating the error: I guess we don't want to propagate it?

internal/engine/experiment/quicping/quicping.go
kelmenhorst: maybe `dissect.go` ?

PR #683, opened 2022-02-04
feat(torsf): collect tor logs, select rendezvous method, count bytes

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2004, https://github.com/ooni/probe/issues/1686
- [x] if you changed anything related how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: https://github.com/ooni/spec/pull/231

## Description

This diff contains significant improvements over the previous
implementation of the torsf experiment.

We add support for configuring different rendezvous methods after
the convo at https://github.com/ooni/probe/issues/2004. In doing
that, I've tried to use a terminology that is consistent with the
names being actually used by tor developers.

In terms of what to do next, this diff basically instruments
torsf to always rendezvous using domain fronting. Yet, it's also
possible to change the rendezvous method from the command line,
when using miniooni, which allows to experiment a bit more. In the
same vein, by default we use a persistent tor datadir, but it's
also possible to use a temporary datadir using the cmdline.

Here's how a generic invocation of `torsf` looks like:

```bash
./miniooni -O DisablePersistentDatadir=true \
           -O RendezvousMethod=amp \
           -O DisableProgress=true \
           torsf
```

(The default is `DisablePersistentDatadir=false` and
`RendezvousMethod=domain_fronting`.)

With this implementation, we can start measuring whether snowflake
and tor together can boostrap, which seems the most important thing
to focus on at the beginning. Understanding why the bootstrap most
often does not converge with a temporary datadir on Android devices
remains instead an open problem for now. (I'll also update the
relevant issues or create new issues after commit this.)

We also address some methodology improvements that were proposed
in https://github.com/ooni/probe/issues/1686. Namely:

- we record the tor version;

- we include the bootstrap percentage because of the logs;

- we set the anomaly key correctly;

- we measure the bytes send and received (by `tor`).

What remains to be done is the possibility of including Snowflake
events into the measurement, which is not possible until the new
improvements at common/event in snowflake.git are included into a
tagged version of snowflake itself. (I'll make sure to mention
this aspect to @cohosh in https://github.com/ooni/probe/issues/2004.)


internal/cmd/ptxclient/ptxclient.go
```diff
@@ -1,43 +0,0 @@
-// Command ptxclient implements a pluggable transports CLI client. This command
-// is only meant for testing and is not production ready.
```
bassosimone, line 2: This command line client is no longer necessary.

internal/engine/experiment/torsf/torsf.go
```diff
@@ -5,23 +5,35 @@
...
 	"github.com/ooni/probe-cli/v3/internal/engine/netx/archival"
 	"github.com/ooni/probe-cli/v3/internal/model"
 	"github.com/ooni/probe-cli/v3/internal/ptx"
+	"github.com/ooni/probe-cli/v3/internal/scrubber"
 	"github.com/ooni/probe-cli/v3/internal/tunnel"
 )
 
 // testVersion is the tor experiment version.
-const testVersion = "0.1.1"
+const testVersion = "0.2.0"
```
bassosimone, line 25: This breaks unit tests. So, we need to adapt tests.

internal/engine/experiment/torsf/torsf.go
```diff
@@ -31,6 +43,15 @@ type TestKeys struct {
...
 	Failure *string `json:"failure"`
+
+	// PersistentDatadir indicates whether we're using a persistent tor datadir.
+	PersistentDatadir bool `json:"persistent_datadir"`
+
+	// RendezvousMethod contains the method used to perform the rendezvous.
+	RendezvousMethod string `json:"rendezvous_method"`
+
+	// TorLogs contains the bootstrap logs.
+	TorLogs []string `json:"tor_logs"`
```
bassosimone, line 54: We need to update the related ooni/spec to account for these new fields in the results.

internal/engine/experiment/torsf/torsf.go
```diff
@@ -73,22 +94,27 @@ func (m *Measurer) Run(
 	ctx context.Context, sess model.ExperimentSession,
 	measurement *model.Measurement, callbacks model.ExperimentCallbacks,
 ) error {
+	ptl, sfdialer, err := m.setup(sess.Logger())
+	if err != nil {
+		// we cannot setup the experiment
+		return err
+	}
+	defer ptl.Stop()
```
bassosimone, line 102: We recently said that sync tests should only return error when something fundamental happens and the test could not run. For this reason, I've refactored `torsf` so that first we see if all the preconditions are met. If they are not met, then we just return an error. Otherwise, we measure. This design is cleaner with respect to the changing requirements in terms of erroring out or returning nil.

internal/engine/experiment/torsf/torsf.go
```diff
@@ -101,47 +127,127 @@ func (m *Measurer) Run(
...
+//
+// Tor is know to be good software that does not break its output
+// unnecessarily and that does not include PII into its logs unless
+// explicitly asked to. This fact gives me confidence that we can
+// safely include this subset of the logs into the results.
+//
+// On this note, I think it's safe to include timestamps from the
+// logs into the output, since we have a timestamp for the whole
+// experiment already, so we don't leak much more by also including
+// the Tor proper timestamps into the results.
```
bassosimone, line 215: I think this is one of the most important parts to review.

internal/engine/session.go
```diff
@@ -204,6 +208,11 @@ func NewSession(ctx context.Context, config SessionConfig) (*Session, error) {
 	return sess, nil
 }
 
+// TunnelDir returns the persistent directory used by tunnels.
+func (s *Session) TunnelDir() string {
+	return s.tunnelDir
+}
+
```
bassosimone, line 215: I need to expose to experiments the directory used by persistent tunnels (e.g., the ones that up until this diff one could only leverage by using `miniooni --tunnel=TUNNEL ...`).

I am wondering whether we should create a lockfile here but I think `tor` probably already does that?

internal/ptx/snowflake.go
```diff
@@ -2,20 +2,130 @@ package ptx
...
+func NewSnowflakeRendezvousMethod(method string) (SnowflakeRendezvousMethod, error) {
+	switch method {
+	case "domain_fronting", "":
+		return NewSnowflakeRendezvousMethodDomainFronting(), nil
+	case "amp":
+		return NewSnowflakeRendezvousMethodAMP(), nil
+	default:
+		return nil, ErrSnowflakeNoSuchRendezvousMethod
+	}
+}
```
bassosimone, line 99: All of this is new code that needs to be tested. I chose to make the rendezvous strategy configurable because it seems to me we want to leverage this opportunity based on the convo at https://github.com/ooni/probe/issues/2004.

I am also wondering whether we want (maybe at a later stage?) to try different strategies each time we run torsf rather than just defaulting to domain_fronting + persistent data directory.

internal/ptx/snowflake.go
```diff
@@ -2,20 +2,130 @@ package ptx
...
 	newClientTransport func(config sflib.ClientConfig) (snowflakeTransport, error)
 }
 
+// NewSnowflakeDialer creates a SnowflakeDialer with default settings.
+func NewSnowflakeDialer() *SnowflakeDialer {
+	return &SnowflakeDialer{
+		RendezvousMethod:   NewSnowflakeRendezvousMethodDomainFronting(),
+		newClientTransport: nil,
+	}
+}
```
bassosimone, line 118: Minor but important aspect. I've changed the pattern with which we construct a `SnowflakeDialer` to require using a constructor _or_ explicitly setting a rendezvous method. (I could have made the rendezvous method optional, but TBH I am starting to think that using constructors in our code base reduces the amount of functions we need to write to deal with default cases, and this seems good in terms of easier testing - opinions?)

internal/tunnel/tunnel.go
```diff
@@ -62,8 +62,13 @@ type Tunnel interface {
 	// SOCKS5ProxyURL returns the SOCSK5 proxy URL.
 	SOCKS5ProxyURL() *url.URL
 
+	// LogFilePath returns the location of the logfile. If
+	// you read the log file before calling stop, you should
+	// be prepared for it to contain incomplete lines.
+	LogFilePath() (string, bool)
+
```
bassosimone, line 69: Tor was already saving a logfile. So, what we need here is to expose it. All tunnels but tor will not have a log file for now. But we'll probably also add more detailed logs for psiphon in the future.

bassosimone (commented):
First pass of self review. I've highlighted when we should improve with testing and have added comments for the reviewers.

bassosimone:
@hellais this pull request is not complete because it's still missing to adjust tests. However, it already contains ~finished code for methodological improvements with `torsf`. Thus, it would be great if you could (1) take a look at the diff with particular attention on the sections that I've marked as important and (2) test from the command line using `miniooni` (build with `go build -v ./internal/cmd/miniooni`, see the initial comment for usage info) and report back on the usefulness of the data structures that I am currently collecting. Thanks a lot! 🙏

internal/engine/experiment/torsf/torsf.go
```diff
@@ -101,47 +127,127 @@ func (m *Measurer) Run(
...
+	tk.BootstrapTime = tun.BootstrapTime().Seconds()
+}
+
+// readTorLogs attempts to read and include the tor logs into
+// the test keys if this operation is possible.
+//
+// This function aims to _only_ include:
+//
+// 1. information on the tor version by intercepting the line that
+// writes which tor version is opening a log file;
```
hellais, line 194: I tried running this test with:
```
./miniooni -ODisablePersistentDatadir=true \
           -ORendezvousMethod=amp \
           -ODisableProgress=true torsf
```

And didn't find in the `tor_logs` list the line referring to the specific tor version that was being used.

I wonder if we should rather obtain the tor_version from the control port and embed it as separate key in the data format (see: https://github.com/cretz/bine/blob/f9f678b84ccaed1c3f919406b44a7d95ddbd6d47/control/cmd_protocolinfo.go#L13).

internal/engine/experiment/torsf/torsf.go
```diff
@@ -101,47 +127,127 @@ func (m *Measurer) Run(
...
+	}
+	for _, bline := range bytes.Split(data, []byte("\n")) {
+		sline := string(bline) // avoid IP addresses in logs
+		if strings.HasSuffix(sline, "opening a new log file.") {
+			tk.TorLogs = append(tk.TorLogs, sline)
+		}
+		if !strings.Contains(sline, "[notice]") {
+			continue
+		}
+		sline = scrubber.Scrub(sline) // maybe unneeded paranoia to avoid including IPs
```
hellais, line 235: It looks like this scrubber is a bit too paranoid. When opening it up to log all messages, it's also matching the Tor version string which it thinks is an ipv4 address (ex. `0.4.6.7`).

Since we are already performing the stripping of the probe_ip at the JSON encoding level, shouldn't we be OK removing this?

internal/engine/session.go
hellais: What race condition are you concerned about? Multiple tor instances sharing the same data_dir?

internal/ptx/snowflake.go
hellais: I think this is a reasonable approach and I've seen ti quite commonly used in golang projects.

internal/engine/experiment/torsf/torsf.go
```diff
@@ -31,6 +43,15 @@ type TestKeys struct {
 
 	// Failure contains the failure string or nil.
 	Failure *string `json:"failure"`
+
+	// PersistentDatadir indicates whether we're using a persistent tor datadir.
+	PersistentDatadir bool `json:"persistent_datadir"`
```
hellais, line 48: Do we perhaps also want to include a timestamp (maybe rounded to the day) of when the datadir was last modified and if it's empty? I am thinking it would be helpful to know if your datadir actually has data in it or if the data is stale, hence needing to fetch descriptor information.

internal/engine/session.go
bassosimone: Yes, indeed! Now the data dir is `$OONI_HOME/tunnel/torsf` so two concurrent `torsf` tests would use it concurrently.

internal/engine/experiment/torsf/torsf.go
bassosimone: Okay, I removed the scrubber

internal/engine/experiment/torsf/torsf.go
```diff
@@ -101,47 +127,127 @@ func (m *Measurer) Run(
...
+		}
+		sline = scrubber.Scrub(sline) // maybe unneeded paranoia to avoid including IPs
+		if strings.Contains(sline, " Bootstrapped ") {
+			tk.TorLogs = append(tk.TorLogs, sline)
+		}
+		if strings.Contains(sline, " new bridge descriptor ") {
+			tk.TorLogs = append(tk.TorLogs, sline)
+		}
+	}
+}
```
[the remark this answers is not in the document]
bassosimone: We need the Snowflake team to tag a version that includes their new event API. I'm going to open an issue about this in their tracker once I'm done with this chunk of work.

internal/engine/experiment/torsf/torsf.go
bassosimone: I don't think it's necessary. The logs I'm including already clearly show whether the cache is cold or not. I think it is better to rely on parsing the logs than adding more complexity on the probe. Tor is known for stable output messages AFAICT, and therefore I would rely on that rather than adding more possible fragile code to check whether the cache is cold. (It may in principle be that the cache does not seem cold but actually it is cold: the time of the last modification is just a proxy and we can tell whether it was cold or not very reliably if we observe what tor had to do.)

internal/bytecounter/context.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// WithExperimentByteCounter assigns the experiment byte counter to the context.
+func WithExperimentByteCounter(ctx context.Context, counter *Counter) context.Context {
+	return context.WithValue(ctx, byteCounterExperimentKey{}, counter)
+}
+
+// WrapWithContextByteCounters wraps a conn with the byte counters
+// that have previosuly been configured into a context.
+func WrapWithContextByteCounters(ctx context.Context, conn net.Conn) net.Conn {
```
bassosimone, line 36: This function should be named `MaybeWrap...`

internal/engine/experiment/torsf/torsf.go
bassosimone: Good suggestion, thank you! I have implemented that!

internal/ptx/snowflake.go
bassosimone: Thanks!

internal/engine/experiment/torsf/torsf.go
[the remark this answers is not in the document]
bassosimone: I mean, I could also pin a commit instead of a stable version but I'd rather not do that 😬

bassosimone:
This PR is approaching readiness state but there's still need to address a couple of comments!

internal/engine/session.go
bassosimone: So, I tried to run two instances of `tor` sharing the same directory. Here's what I've got:

```
Feb 07 12:36:38.000 [warn] It looks like another Tor process is running with the same data directory.  Waiting 5 seconds to see if it goes away.
Feb 07 12:36:43.000 [err] No, it's still there.  Exiting.
Feb 07 12:36:43.000 [err] set_options: Bug: Acting on config options left us in a broken state. Dying. (on Tor 0.4.6.9 )
Feb 07 12:36:43.000 [err] Reading config failed--see warnings above.
```

I think we're covered here. ✔️

internal/engine/experiment/torsf/torsf.go
bassosimone: done! ✔️

internal/engine/experiment/torsf/torsf.go
bassosimone: I have documented into the related ooni/spec changes in https://github.com/ooni/spec/pull/231 the difference in terms of tor logs between a bootstrap with a stale cache and one with a valid cache. Because the different is quite obvious, I'd rather just leave the pipeline to post-process this information and decide about the state of the cache.

(It must also be said that, interestingly, when I was producing data for updating https://github.com/ooni/spec/pull/231 I saw a bootstrap with stale cache converging in 67 s and one with a good/fresh cache taking ~300 s.)

internal/engine/experiment/torsf/torsf.go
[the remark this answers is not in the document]
bassosimone: What I'll do is to open a related PR that pins to a development Snowflake version to collect Snowflake events and I'll update this PR and merge once support for that lands into a stable Snowflake release.

internal/ptx/snowflake_test.go
```diff
@@ -11,12 +11,107 @@ import (
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		if _, ok := meth.(*snowflakeRendezvousMethodAMP); !ok {
+			t.Fatal("unexpected method type")
+		}
+	})
+
+	t.Run("for another value", func(t *testing.T) {
+		meth, err := NewSnowflakeRendezvousMethod("amptani")
```
bassosimone, line 83: @hellais I hope you appreciate this portmanteau

internal/engine/experiment/torsf/torsf.go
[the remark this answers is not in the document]
bassosimone: So, here's the draft PR to collect snowflake events: https://github.com/ooni/probe-cli/pull/685. It seems something's wrong in my understanding, since I am not getting any events. Anyway, since this code has not been released as stable yet, I think collecting Snowflake events will be done on a separate PR.

internal/engine/experiment/torsf/testdata/tor.log
```diff
@@ -0,0 +1,31 @@
+Feb 04 15:04:29.000 [notice] Tor 0.4.6.9 opening new log file.
+Feb 04 15:04:29.360 [notice] We compiled with OpenSSL 101010cf: OpenSSL 1.1.1l  FIPS 24 Aug 2021 and we are running with OpenSSL 101010cf: 1.1.1l. These two versions should be binary compatible.
+Feb 04 15:04:29.363 [notice] Tor 0.4.6.9 running on Linux with Libevent 2.1.12-stable, OpenSSL 1.1.1l, Zlib 1.2.11, Liblzma 5.2.5, Libzstd 1.5.2 and Glibc 2.34 as libc.
+Feb 04 15:04:29.363 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://www.torproject.org/download/download#warning
+Feb 04 15:04:29.363 [warn] Tor was compiled with zstd 1.5.1, but is running with zstd 1.5.2. For safety, we'll avoid using advanced zstd functionality.
+Feb 04 15:04:29.363 [notice] Read configuration file "/home/sbs/.miniooni/tunnel/torsf/tor/torrc-2981077975".
```
hellais, line 6: I think we should add some sanitisation to the log parser so that we don't leak the username in the path in the measurement.

hellais: Ah nevermind, I see that we are already airing on the side of caution by ignoring anything that doesn't contain the string `Bootstrapped` or the `new bridge descriptor` line, so we should be good.

hellais:
In doing some testing of this, I noticed that if I run the `torsf` test without any arguments in miniooni I get the following error:
```
miniooni torsf
[      0.000222] <info> Current time: 2022-02-07 14:57:03 CET
[      0.002505] <info> miniooni home directory: $HOME/.miniooni
[      0.002755] <info> Looking up OONI backends; please be patient...
[      0.733684] <info> session: using probe services: {Address:https://ps1.ooni.io Type:https Front:}
[      0.733706] <info> Looking up your location; please be patient...
[      4.567626] <info> - country: IT
[      4.567656] <info> - network: Vodafone Italia S.p.A. (AS30722)
[      4.567666] <info> - resolver's IP: 172.253.12.129
[      4.567670] <info> - resolver's network: Google LLC (AS15169)
[      4.567683] <warn> cannot create experiment builder: map[error:no such experiment: torsf]
[      4.568241] <info> sessionresolver: failure rate: primary: 4/4; fallback: 0/4
[      4.568447] <info> whole session: recv   7 Mbyte, sent   5 kbyte
cannot create experiment builder%
```

I guess it's fine to require specifying correct arguments as miniooni is an experimental client, but if it's not too hard we might want to still support running the test with default arguments or improve the returned error (the fact I read `map[error:no such experiment: torsf]`, made me wonder if I was on the correct branch).

internal/engine/experiment/torsf/testdata/tor.log
bassosimone: To further clarify: this file is a test fixture containing all my logs to ensure we extract bootstrap logs.

PR #688, opened 2022-02-09
doc: clarify semantic versioning policy

bassosimone:
The oonimkall package is only public for technical reasons. We
cannot use `go mobile` on a private package. We consider oonimkall
our private interface to our mobile apps, thus we reserve the
right to change its API without bumping the major number.

We'll bump the major number in case of breaking changes in the
cmd/ooniprobe CLI interface, or in case of other major improvements
that significantly modify cmd/ooniprobe.


bassosimone:
This diff only changed documentation, so it's fine to merge it without waiting for the CI.

PR #695, opened 2022-02-13
fix(dnscheck):  log "ok" rather than "<nil>" on success

yeganathan18:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2020

## Description

I am working on the issue, by adding a convenience function that returns "ok" rather than "<nil>" values on success. Have to confirm the implementation of the function and tests for the same.

#### Screenshot:

Screenshot 2022-02-14 at 12 28 26 AM


bassosimone (changes requested):
Do you mind updating CONTRIBUTING.md stating that, when an experiment logs intermediate results, it should probably be using `ErrorToStringOrOK`?

I think it's great if we establish this as a new coding practice. Also, while this change is not strictly related to the PR, it feels like something nice to do while we're at it, [in the spirit of leaving the camp cleaner than we found it](https://matheus.ro/2017/12/11/clean-code-boy-scout-rule/).

Thanks!

bassosimone (approved):
Thanks a lot!

🐳

PR #701, opened 2022-02-23
cleanup: use ErrorToStringOrOK func in other tests that returns nil

yeganathan18:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2040

## Description

This PR aims to refactor and ensure "ok" is returned on success of an experiment rather than `<nil>` value.


bassosimone:
Thanks! The diff in here looks good already. I understand that there are more cases in which it seems a bit more tricky. So, I would recommend to first merge the code in here and then discuss how to handle the more tricky parts.

yeganathan18:
Thanks for the feedback, I will mark this PR as ready for review and let's discuss the tricky part!

bassosimone (approved):
🐳

PR #703, opened 2022-03-04
feat: add support for OpenBSD

Erethon:
## Checklist

## Description

This PR adds support for building and running probe-cli on an OpenBSD system. The first commit changes the existing generator files and the second one adds the generated files after running `go generate`.

I've tested this on an i386 OpenBSD 7.0 installation and seems to work fine. I can see my results on explorer.ooni.org with a platform of `openbsd`. Tests also pass, but I didn't run them with `-race` because it's not supported in OpenBSD go (neither i386 or amd64 afaik).


bassosimone:
Thanks a lot! Will include in next release!

bassosimone:
There's a minor failure in the tests here: https://github.com/ooni/probe-cli/runs/5463332835?check_suite_focus=true#step:4:171. It' related to flakiness in tests unrelated to this PR.

bassosimone (approved):
Thank you for your great work! <3

bassosimone:
Thank you, @Erethon! I have also backported your diff to https://github.com/ooni/probe-cli/releases/tag/v3.14.2, so that would be the first release with support for OpenBSD! 🥳 🥳 🥳

PR #706, opened 2022-03-19
fix(httpx): correctly combine paths

yeganathan18:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2010

## Description

This PR ensures that given a BaseURL `https://example.com/example` and an URLPath of `/foo/bar`, the code in httpx should access `https://example.com/example/foo/bar`. Also, the documentation for the same.

### DIFF
--- a/internal/httpx/httpx.go
+++ b/internal/httpx/httpx.go
@@ -144,8 +144,32 @@ func (c *apiClient) newRequestWithJSONBody(
 
 // joinURLPath joins the BaseURL and the URLPath, allows httpx to access the URL
 func (c *apiClient) joinURLPath(URLPath string) string {
-	accessURL := c.BaseURL + URLPath
-	return accessURL
+
+	if c.BaseURL != "" && URLPath != "" {
+		if c.BaseURL[len(c.BaseURL)-1] != '/' && URLPath[0] != '/' {
+			c.BaseURL += "/"
+			return c.BaseURL + URLPath
+		}
+
+		if c.BaseURL[len(c.BaseURL)-1] == '/' && URLPath[0] == '/' {
+			return c.BaseURL + URLPath[1:]
+		}
+
+		if c.BaseURL[len(c.BaseURL)-1] == '/' && URLPath[0] != '/' {
+			return c.BaseURL + URLPath
+		}
+
+		if c.BaseURL[len(c.BaseURL)-1] != '/' && URLPath[0] == '/' {
+			return c.BaseURL + URLPath
+		}
+		return c.BaseURL + URLPath
+	}
+
+	if c.BaseURL != "" && URLPath == "" {
+		return c.BaseURL
+	}
+
+	return ""
 }
 
 // newRequest creates a new request.
--- a/internal/httpx/httpx.go
+++ b/internal/httpx/httpx.go
@@ -74,9 +74,9 @@ const DefaultMaxBodySize = 1 << 22
 // baseURL path is appended with the resourcePath if it is
 // created using APIClientTemplate.
 type APIClient interface {
-	// GetJSON reads the JSON resource at combined path of baseURL with
-	// the resourcePath and unmarshals the results into output.
-	// The request is bounded by the lifetime of the
+	// GetJSON reads the JSON resource whose path is obtained concatenating
+	// the baseURL's path with `resourcePath` and unmarshals the results
+	// into `output`. The request is bounded by the lifetime of the
 	// context passed as argument. Returns the error that occurred.
 	GetJSON(ctx context.Context, resourcePath string, output interface{}) error
 
--- a/internal/httpx/httpx.go
+++ b/internal/httpx/httpx.go
@@ -143,25 +143,20 @@ func (c *apiClient) newRequestWithJSONBody(
 	return request, nil
 }
 
-// joinURLPath joins the BaseURL and the URLPath, allows httpx to access the URL
+// joinURLPath joins the BaseURL and the resource URL Path, allows httpx to access the URL
 func (c *apiClient) joinURLPath(origPath string, newPath string) string {
 
-	if origPath != "" && newPath != "" {
-		if !strings.HasSuffix(origPath, "/") && !strings.HasPrefix(newPath, "/") {
-			origPath += "/"
-		}
-
-		if strings.HasSuffix(origPath, "/") && strings.HasPrefix(newPath, "/") {
-			newPath = newPath[1:]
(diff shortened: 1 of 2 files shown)


internal/httpx/httpx.go
```diff
@@ -142,14 +142,43 @@ func (c *apiClient) newRequestWithJSONBody(
...
+
 // newRequest creates a new request.
 func (c *apiClient) newRequest(ctx context.Context, method, resourcePath string,
 	query url.Values, body io.Reader) (*http.Request, error) {
-	URL, err := url.Parse(c.BaseURL)
+	URL, err := url.Parse(c.joinURLPath(resourcePath))
 	if err != nil {
 		return nil, err
 	}
-	URL.Path = resourcePath
```
bassosimone, line 152: I realize this is a matter of taste, but I have this preference. I'd like `joinURLPath` to take two URL paths and join them. I see this as slightly better than the current implementation. Hence my suggestion:

1. keep the previous code for parsing the URL, i.e., `URL, err := url.Parse(c.BaseURL)`
2. modify line 152 to be like

```Go
URL.Path = c.joinURLPath(URL.Path, resourcePath)
```

3. then the implementation of `joinURLPath` could be like

```Go
if !strings.HasSuffix(origPath, "/") {
  origPath += "/"
}
if strings.HasPrefix(newPath, "/") {
  newPath = newPath[1:]
}
return origPath + newPath
```

4. luckily you've written comprehensive tests, so we should ~quickly see whether this proposed code is still working as intended or not.

5. you should also update the tests as mentioned below

6. please, update the documentation to ensure that we mention this behavior

internal/httpx/httpx_test.go
```diff
@@ -93,6 +93,64 @@ func newAPIClient() *apiClient {
 	}
 }
 
+func TestJoinURLPath(t *testing.T) {
```
bassosimone, line 96: This is very good! What would be excellent is to modify this test to be testing `newRequest`. Then you could set the `BaseURL` and the `resourcePath` like you're doing here and test whether the returned `http.Request`'s URL contains the desired path. I think this makes the test slightly more complex but give us one additional guarantee, i.e., that not only `joinURLPath` is correct but also `newRequest` uses the URL path.

bassosimone (changes requested):
Thank you for working on this! I particularly like that you wrote an extensive test suite for this new functionality. I've provided suggestions on how to improve your diff.

internal/httpx/httpx.go
yeganathan18: Thanks for the code review, I was searching for built-in functions, but every google search ended up old school way 🤦🏻‍♂️

internal/httpx/httpx.go
```diff
@@ -142,14 +143,35 @@ func (c *apiClient) newRequestWithJSONBody(
...
+			newPath = newPath[1:]
+		}
+		return origPath + newPath
+	}
+
+	if origPath != "" && newPath == "" {
+		return origPath
+	}
+
+	return ""
```
bassosimone, line 164:
```suggestion
	if !strings.HasSuffix(origPath, "/")  {
		origPath += "/"
	}
	if strings.HasPrefix(newPath, "/") {
		newPath = newPath[1:]
	}
	return origPath + newPath
```

Would this (simpler) implementation work? Does it look correct?

(If we can achieve the same objective with less code, it's probably better in terms of future maintenance.)

bassosimone: Also, please, modify the documentation of `httpx` to ensure we document what happens when the baseURL has a path and also the resource URL has a path. Thank you!

internal/httpx/httpx.go
```diff
@@ -70,18 +71,22 @@ const DefaultMaxBodySize = 1 << 22
...
 // APIClient is a client configured to call a given API identified
 // by a given baseURL and using a given model.HTTPClient.
+// baseURL path is appended with the resourcePath if it is
+// created using APIClientTemplate.
 type APIClient interface {
-	// GetJSON reads the JSON resource at resourcePath and unmarshals the
-	// results into output. The request is bounded by the lifetime of the
+	// GetJSON reads the JSON resource at combined path of baseURL with
+	// the resourcePath and unmarshals the results into output.
+	// The request is bounded by the lifetime of the
```
bassosimone, line 79:
```suggestion
	// GetJSON reads the JSON resource whose path is obtained concatenating
	// the baseURL's path with `resourcePath` and unmarshals the results
	// into `output`. The request is bounded by the lifetime of the
```

internal/httpx/httpx.go
```diff
@@ -70,18 +71,22 @@ const DefaultMaxBodySize = 1 << 22
...
 
 	// GetJSONWithQuery is like GetJSON but also has a query.
 	GetJSONWithQuery(ctx context.Context, resourcePath string,
 		query url.Values, output interface{}) error
 
-	// PostJSON creates a JSON subresource of the resource at resourcePath
-	// using the JSON document at input and returning the result into the
+	// PostJSON creates a JSON subresource of the resource at
+	// combined path of baseURL with the resourcePath using the JSON
+	// document at input and returning the result into the
```
bassosimone, line 89:
```suggestion
	// PostJSON creates a JSON subresource of the resource whose
	// path is obtained concatenating the baseURL'spath with `resourcePath` using
	// the JSON document at `input` as value and returning the result into the
```

internal/httpx/httpx.go
```diff
@@ -70,18 +71,22 @@ const DefaultMaxBodySize = 1 << 22
 
 // APIClient is a client configured to call a given API identified
 // by a given baseURL and using a given model.HTTPClient.
+// baseURL path is appended with the resourcePath if it is
+// created using APIClientTemplate.
```
bassosimone, line 75:
```suggestion
//
// The resource path argument passed to APIClient methods is appended
// to the base URL's path for determining the full URL's path.
```

internal/httpx/httpx.go
```diff
@@ -142,14 +147,32 @@ func (c *apiClient) newRequestWithJSONBody(
 	return request, nil
 }
 
+// joinURLPath combines the path of resource URL with the baseURL,
+// and allows httpx to access the URL when it creates a new request.
```
bassosimone, line 151:
```suggestion
// joinURLPath appends the path of resource URL to the baseURL taking
// care of multiple forward slashes gracefully.
```

bassosimone (approved):
🐳

PR #707, opened 2022-03-26
fix(torsf): ensure tor-logs-filtering regexp is correct

ParitoshKabra:
## Checklist

- [x] reference issue for this pull request: [2060] (https://github.com/ooni/probe/issues/2060)

## Description
https://github.com/ooni/probe-cli/blob/85664f1e31de0844f444af7c690a57628ff79ff7/internal/engine/experiment/torsf/torsf.go#L200-L201

The regex fragment [a-zA-z]+ was matching strings with an underscore, like conn_pt and onehop_create. Look closely: the second z, which you would expect to be uppercase, is actually lowercase. The underscore character _ (ASCII 95) lies between A (ASCII 65) and z (ASCII 122), which is why the regex works for its intended purpose, despite containing an evident typo.

A small fix in regex to increase its readabliity and thus avoid extra characters like `[\]^_\``.

### DIFF
--- a/internal/engine/experiment/torsf/torsf.go
+++ b/internal/engine/experiment/torsf/torsf.go
@@ -196,7 +196,7 @@ func (m *Measurer) bootstrap(ctx context.Context, sess model.ExperimentSession,
 
 // torProgressRegexp helps to extract progress info from logs.
 //
-// See https://regex101.com/r/3YfIed/1.
+// See https://regex101.com/r/cer3lm/1.
 var torProgressRegexp = regexp.MustCompile(
 	`^[A-Za-z0-9.: ]+ \[notice\] Bootstrapped [0-9]+% \([A-Za-z_]+\): [A-Za-z0-9 ]+$`)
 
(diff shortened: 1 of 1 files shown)


ParitoshKabra:
> 

One doubt 😅 ? No matter whatever order we take, the `[... ]+` will always match any character from the set provided within? Then why is it that we need to enforce this order uppercase first and lowercase second? Is it just for code readability?

ParitoshKabra:
@bassosimone can this small change be merged? 😅

hellais (approved):
Good catch! Also funny that although there was this typo we happened to still match what we needed.

I suggest, before merging, you update the documentation link to: https://regex101.com/r/3YfIed/1 with the fixed regexp.

internal/engine/experiment/torsf/torsf.go
```diff
@@ -198,7 +198,7 @@ func (m *Measurer) bootstrap(ctx context.Context, sess model.ExperimentSession,
 //
 // See https://regex101.com/r/3YfIed/1.
```
bassosimone, line 199:
```suggestion
// See https://regex101.com/r/cer3lm/1.
```

bassosimone:
@ParitoshKabra:

> One doubt 😅 ? No matter whatever order we take, the [... ]+ will always match any character from the set provided within? Then why is it that we need to enforce this order uppercase first and lowercase second? Is it just for code readability?

Yes, it's for readability/consistency.

PR #708, opened 2022-04-12
Issue/2067

bassosimone:
## Checklist

## Description


bassosimone:
Okay, so the status of running tests [does not look completely good](https://github.com/ooni/probe-cli/runs/5987172335). Let's see what is happening and decide whether to merge or change tests.

The [first error](https://github.com/ooni/probe-cli/runs/5987172335?check_suite_focus=true#step:4:115) is the following:

```
# github.com/ooni/probe-cli/v3/internal/archival [github.com/ooni/probe-cli/v3/internal/archival.test]
Error: internal/archival/quic_test.go:206:11: cannot use qtls.ConnectionStateWith0RTT{…} (value of type "github.com/marten-seemann/qtls-go1-17".ConnectionStateWith0RTT) as type "github.com/marten-seemann/qtls-go1-18".ConnectionStateWith0RTT in struct literal
```

The specific code reads:

```Go
			MockConnectionState: func() quic.ConnectionState {
				return quic.ConnectionState{
					TLS: qtls.ConnectionStateWith0RTT{
						ConnectionState: state,
					},
				}
			},
```

The problem is basically that `qtls` is the following:

```Go
	"github.com/marten-seemann/qtls-go1-17" // it's annoying to depend on that
```

So, from the comment we can say that it's obvious that this test would break when using go1.18. I think it's fine to ignore this failure, since we're still building ooniprobe using go1.17. We'll fix this when we'll migrate to go1.18.

The [second error](https://github.com/ooni/probe-cli/runs/5987172335?check_suite_focus=true#step:4:199) is this:

```
--- FAIL: TestDialerSystem (0.00s)
    --- FAIL: TestDialerSystem/DialContext (0.00s)
        --- FAIL: TestDialerSystem/DialContext/with_canceled_context (0.00s)
Error:             dialer_test.go:67: unexpected err dial tcp: lookup dns.google: operation was canceled
FAIL
```

where the code reads:

```Go
	t.Run("DialContext", func(t *testing.T) {
		t.Run("with canceled context", func(t *testing.T) {
			d := &dialerSystem{}
			ctx, cancel := context.WithCancel(context.Background())
			cancel() // immediately!
			conn, err := d.DialContext(ctx, "tcp", "dns.google:443")
			if err == nil || err.Error() != "dial tcp: operation was canceled" {
				t.Fatal("unexpected err", err)
			}
			if conn != nil {
				t.Fatal("unexpected conn")
			}
		})
```

Maybe this is a test that we can try to fix for both versions of Go.

bassosimone:
Okay, it seems we're good to merge with some (explained) 🔥 🔥 🔥

PR #233, opened 2021-02-21
riseupvpn: reduce false positives

cyBerta:
part of  https://github.com/ooni/probe/issues/1354

PR #711, opened 2022-04-19
tls_handshakes: add endpoint addresses to handshake list

DecFox:
## Checklist

- [x] reference issue for this pull request: [Issue #2065](https://github.com/ooni/probe/issues/2065)
- [x] if you changed anything related how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: [PR #233](https://github.com/ooni/spec/pull/233)

## Description
- Added IP addresses to the tls_handshakes list
- Tested measurement outputs on webconnectivity, IM and riseupvpn experiments. 

We can track the tls_handshakes issue here. Although I have made some changes, we may come up with a better approach to handle this issue.

### DIFF
--- a/internal/engine/experiment/urlgetter/getter.go
+++ b/internal/engine/experiment/urlgetter/getter.go
@@ -77,7 +77,7 @@ func (g Getter) Get(ctx context.Context) (TestKeys, error) {
 		tk.TCPConnect, archival.NewTCPConnectList(g.Begin, events)...,
 	)
 	tk.TLSHandshakes = append(
-		tk.TLSHandshakes, archival.NewTLSHandshakesList(g.Begin, events)...,
+		tk.TLSHandshakes, archival.NewTLSHandshakesList(g.Begin, g.Target, events)...,
 	)
 	return tk, err
 }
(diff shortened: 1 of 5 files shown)


hellais (commented):
Thanks for putting this together!

I think it's going to be a bit more robust if the `address` is extracted directly when the event is created inside of `internal/engine/netx/tlsdialer/saver.go` by calling `conn.RemoteAddr().String()`.

You will then have it attached to the event model and will just have to pass it around by changing the archival models (`internal/model/archival.go` `ArchivalTLSOrQUICHandshakeResult` and `internal/engine/netx/archival/archival.go`)

DecFox:
Thanks! This seems to be much better. Made the required changes.

bassosimone (approved):
LGTM

bassosimone:
Ah, pushed https://github.com/ooni/probe-cli/pull/711/commits/4acd41f8609f81c375b75136ed6cff15059503bd on top of the existing PR. Maybe it's just paranoia, but I think the diff makes the code slightly more robust to future changes.

PR #714, opened 2022-04-29
fix(ooniprobe): use ooniprobe-cli-unattended for unattended runs

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2081
- [x] if you changed anything related how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: N/A

## Description

This diff changes the software name used by unattended runs for which
we did not override the default software name (`ooniprobe-cli`).

It will become `ooniprobe-cli-unattended`. This software name is in line
with the one we use for Android, iOS, and desktop unattended runs.

While working in this diff, I introduced string constants for the run
types and a string constant for the default software name.

See https://github.com/ooni/probe/issues/2081.


hellais (approved):
This looks good to me.

I think it's quite an improvement to use const types instead of using strings everywhere for distinguishing between the run types.

PR #718, opened 2022-05-10
feat: re-implement the vanilla_tor experiment

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/803
- [x] if you changed anything related how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: https://github.com/ooni/spec/pull/239

## Description

This diff re-implements the vanilla_tor experiment. This experiment was
part of the ooni/probe-legacy implementation.

The reference issue is https://github.com/ooni/probe/issues/803. We didn't
consider the possible improvements mentioned by the
https://github.com/ooni/probe/issues/803#issuecomment-598715694 comment,
which means we'll need to create a follow-up issue for them. We will
then decide whether, when, and how to implement those follow-up measurements
either into `vanilla_tor` or into the existing `tor` experiment.

This novel `vanilla_tor` implementation emits test_keys that are mostly
compatible with the original implementation, however:

1. the `timeout` is a `float64` rather than integer (but the default
timeout is an integer, so there are no JSON-visible changes);

2. the `tor_log` string is gone and replaced by the `tor_logs` list
of strings, which contains the same information;

3. the definition of `error` has been augmented to include the
case in which there is an unknown error;

4. the implementation of vanilla_tor mirrors closely the one of torsf
and we have taken steps to make the two implementations as comparable
as possible in terms of the generated JSON measurement.

The main reason why we replaced `tor_log` with `tor_logs` are:

1. that `torsf` already used that;

2. that reading the JSON is easier with this implementation compared to
an implementation where all logs are into the same string.

If one is processing the new data format using Python, then it will
not be difficult convert `tor_log` to `tor_logs`. In any case, because
we extract the most interesting fields (e.g., the percentage of the
bootstrap where tor fails), it seems that logs are probably more useful
as something you want to read in edge cases (I guess).

Also, because we want `torsf` and `vanilla_tor` to have similar JSONs,
we renamed `torsf`'s `default_timeout` to `timeout`. This change has little
to none real-world impact, because no stable version of OONI Probe has
ever shipped a `torsf` producing the `default_timeout` field.

Regarding the structure of this diff, we have:

1. factored code to parse tor logs into a separate package;

2. implemented `vanilla_tor` as a stripped down `torsf` and added further
changes to ensure compatibility with the previous `vanilla_tor`'s data format;

3. improved `torsf` to merge back the changes in `vanilla_tor`, so the two
data formats of the two experiments are as similar as possible.

We believe producing as similar as possible data formats helps anyone who's
reading measurements generated by both experiments.

We have retained/introduced `vanilla_tor`'s `error` field, which is not very
useful when one has a more precise failure but is still what `vanilla_tor`
used to emit, so it makes sense to also have this field.

In addition to changing the implementation, we also updated the specs.

As part of our future work, we may want to consider factoring the common code
of these two experiments into the same underlying support library.

### DIFF
--- a/internal/engine/allexperiments.go
+++ b/internal/engine/allexperiments.go
@@ -328,7 +328,7 @@ var experimentsByName = map[string]func(*Session) *ExperimentBuilder{
 		}
 	},
 
-	"vanillator": func(session *Session) *ExperimentBuilder {
+	"vanilla_tor": func(session *Session) *ExperimentBuilder {
 		return &ExperimentBuilder{
 			build: func(config interface{}) *Experiment {
 				return NewExperiment(session, vanillator.NewExperimentMeasurer(
--- a/internal/engine/allexperiments.go
+++ b/internal/engine/allexperiments.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/engine/experiment/tor"
 	"github.com/ooni/probe-cli/v3/internal/engine/experiment/torsf"
 	"github.com/ooni/probe-cli/v3/internal/engine/experiment/urlgetter"
+	"github.com/ooni/probe-cli/v3/internal/engine/experiment/vanillator"
 	"github.com/ooni/probe-cli/v3/internal/engine/experiment/webconnectivity"
 	"github.com/ooni/probe-cli/v3/internal/engine/experiment/webstepsx"
 	"github.com/ooni/probe-cli/v3/internal/engine/experiment/whatsapp"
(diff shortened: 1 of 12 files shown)


bassosimone:
Because https://github.com/ooni/probe-cli/pull/718/commits/9f8fe9b33e6e28db263c32efdcf5107b0d7bc0b4 only contains a change in a comment, I am going to merge as soon as the previous build succeeds.

PR #733, opened 2022-05-15
fix(netxlite): prefer composition over embedding

bassosimone:
This diff has been extracted and adapted from https://github.com/bassosimone/websteps-illustrated/commit/8848c8c516a40663c2718b1aff271884a116a147

The reason to prefer composition over embedding is that we want the
build to break if we add new methods to interfaces we define. If the build
does not break, we may forget about wrapping methods we should
actually be wrapping. I noticed this issue inside netxlite when I was working
on websteps-illustrated and I added support for NS and PTR queries.

See https://github.com/ooni/probe/issues/2096

While there, perform comprehensive netxlite code review
and apply minor changes and improve the docs.

### DIFF
--- a/internal/netxlite/bogon.go
+++ b/internal/netxlite/bogon.go
@@ -18,7 +18,7 @@ import (
 // function a non-IP address causes it to return true.
 func IsBogon(address string) bool {
 	ip := net.ParseIP(address)
-	return ip == nil || isPrivate(address, ip)
+	return ip == nil || isBogon(address, ip)
 }
 
 // IsLoopback returns whether an IP address is loopback. Passing to this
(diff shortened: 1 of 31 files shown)


bassosimone:
So, I have run tests three times and each time the 1.18.1 coverage measurement failed with:

```
--- FAIL: TestCheckReportIDWorkingAsIntended (0.53s)
    checkreportid_test.go:32: httpx: request failed: 502 Bad Gateway
2022/05/15 17:15:38  info New reportID:            
2022/05/15 17:15:38  info New reportID:            
FAIL
coverage: 97.8% of statements
```

This kind of failure seems to be completely unrelated with the changes I implemented.

So, I am going to merge anyway.

PR #755, opened 2022-05-24
cleanup: mark more integration tests as !short mode

bassosimone:
The objective is to make PR checks run much faster.

See https://github.com/ooni/probe/issues/2113 for context.

Regarding netxlite's tests:

Checking for every commit on master or on a release branch is
good enough and makes pull requests faster than one minute since
netxlite for windows is now 1m slower than coverage.

We're losing some coverage but coverage from integration tests
is not so good anyway, so I'm not super sad about this loss.

### DIFF
--- a/internal/engine/experiment/whatsapp/whatsapp_test.go
+++ b/internal/engine/experiment/whatsapp/whatsapp_test.go
@@ -28,6 +28,9 @@ func TestNewExperimentMeasurer(t *testing.T) {
 }
 
 func TestSuccess(t *testing.T) {
+	if testing.Short() {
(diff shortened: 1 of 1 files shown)


pkg/oonimkall/session_integration_test.go
```diff
@@ -47,29 +43,6 @@ func TestNewSessionWithInvalidStateDir(t *testing.T) {
 	}
 }
 
-func TestMaybeUpdateResourcesWithCancelledContext(t *testing.T) {
```
bassosimone, line 50: Here we were testing a deprecated function that always returns `nil`.

PR #758, opened 2022-05-25
cleanup: merge httpheader and httpfailure into model

bassosimone:
These two small packages could easily be merged into the model
package, since they're clearly model-like packages.

Part of https://github.com/ooni/probe/issues/2115


internal/engine/geolocate/cloudflare.go
```diff
@@ -21,7 +20,7 @@ func cloudflareIPLookup(
 		BaseURL:    "https://www.cloudflare.com",
 		HTTPClient: httpClient,
 		Logger:     logger,
-		UserAgent:  httpheader.CLIUserAgent(),
+		UserAgent:  model.HTTPHeaderUserAgent,
```
bassosimone, line 23: Here it was actually wrong to use the CLI user agent

PR #761, opened 2022-05-25
Spring cleanup: remove unused/unneded code

bassosimone:
See https://github.com/ooni/probe/issues/2116 for rationale


bassosimone:
Of course the coverage decreases, because we're removing packages and some of them were very well tested. I do not think we should be too sad about this fact, though. We want anyway to mostly report coverage based on unit tests, so it means we will need to pay a coverage hit until we start having more unit tests. I'll go ahead with the cleanup.

bassosimone:
(This is probably my most productive day 😬...)

PR #766, opened 2022-05-30
netxlite: do not call netgo the CGO_ENABLED=0 resolver

bassosimone:
In https://github.com/ooni/probe/issues/2029#issuecomment-1140805266, we
explained why calling it "netgo" would be incorrect.

In other words, we can get the platform's `getaddrinfo` as long as
we're not cross compiling. We do cross compile `ooniprobe`, actually
it's not even possible to cross compile it.

For increased accuracy, we should stop cross compiling `miniooni`
as well, so it would also directly use `getaddrinfo`.

This diff fixes at the same time ooni/probe-cli and ooni/spec
and we'll open two pull requests in parallel.

### DIFF
--- a/internal/model/netx.go
+++ b/internal/model/netx.go
@@ -201,10 +201,11 @@ type SimpleResolver interface {
 
 	// Network returns the resolver type. It should be one of:
 	//
-	// - netgo: means we're using golang's "netgo" UDP resolver, which
-	// reads /etc/resolv.conf and only works on Unix systems;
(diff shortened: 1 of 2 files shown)


bassosimone:
[...]
> For increased accuracy, we should stop cross compiling `miniooni` as well, so it would also directly use `getaddrinfo`.

I created https://github.com/ooni/probe/issues/2119 to track this follow up work.

PR #773, opened 2022-06-01
refactor(tracex): start applying recent code conventions

bassosimone:
The code that is now into the tracex package was written a long
time ago, so let's start to make it more in line with the coding
style of packages that were written more recently.

I didn't apply all the changes I'd like to apply in a single diff
and for now I am committing just this diff.

Broadly, what we need to do is:

1. improve documentation

2. ~always use pointer receivers (object receives have the issue
that they are not mutable by accident meaning that you can mutate
them but their state do not change after the call returns, which
is potentially a source of bugs in case you later refactor to use
a pointer receiver, so always use pointer receivers)

3. ~always avoid embedding (let's say we want to avoid embedding
for types we define and it's instead fine to embed types that are
defined in the stdlib: if later we add a new method, we will not
see a broken build and we'll probably forget to add the new method
to all wrappers -- conversely, if we're wrapping rather than
embedding, we'll see a broken build and act accordingly)

4. prefer unit tests and group tests by type being tested rather
than using a flat structure for tests

Reference issue: https://github.com/ooni/probe/issues/2121

### DIFF
--- a/internal/engine/experiment/urlgetter/configurer_test.go
+++ b/internal/engine/experiment/urlgetter/configurer_test.go
@@ -119,7 +119,7 @@ func TestConfigurerNewConfigurationResolverDNSOverHTTPSPowerdns(t *testing.T) {
 	if !ok {
 		t.Fatal("not the resolver we expected")
 	}
-	stxp, ok := sr.Txp.(tracex.SaverDNSTransport)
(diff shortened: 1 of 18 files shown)


bassosimone:
There's a coverage slippage that I'll compensate in a follow-up diff where I'll focus on unit testing.

PR #822, opened 2022-07-01
fix(workflows): elevate GITHUB_TOKEN permissions when needed

bassosimone:
I am not 100% sure I was able to fix all the cases in which we
need higher permissions than the strict default.

At least, I tried.

It may be reasonable to make an interim release to check whether I
successfully fixed all the cases.

Ref issue: https://github.com/ooni/probe/issues/2154

### DIFF
--- a/.github/workflows/alltests.yml
+++ b/.github/workflows/alltests.yml
@@ -1,9 +1,10 @@
-# alltests runs the whole test suite
+# Runs the whole test suite
 name: alltests
 on:
   push:
     branches:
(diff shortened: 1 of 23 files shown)


bassosimone:
So, https://github.com/ooni/probe-cli/pull/823 showed that the workflow changes are good. It's sad we still have many failures but it's fine to have them in the `master` branch. This fact just suggests more work when we will be creating the `release/3.16` branch for tracking the next minor release of `ooniprobe`.

PR #823, opened 2022-07-01
[backport] fix(workflows): elevate GITHUB_TOKEN permissions when needed

bassosimone:
This diff backports 937f6e29043dbdab65fe663963e209afa66bddac from
the `master` branch. Original commit message follows:

- - -

I am not 100% sure I was able to fix all the cases in which we
need higher permissions than the strict default.

At least, I tried.

It may be reasonable to make an interim release to check whether I
successfully fixed all the cases.

Ref issue: https://github.com/ooni/probe/issues/2154

- - -

Conflicts:
	.github/workflows/android.yml
	.github/workflows/gosec.yml
	.github/workflows/netxlite.yml

### DIFF
--- a/.github/workflows/alltests.yml
+++ b/.github/workflows/alltests.yml
@@ -1,9 +1,10 @@
-# alltests runs the whole test suite
+# Runs the whole test suite
 name: alltests
 on:
   push:
     branches:
       - "release/**"
+
 jobs:
   test:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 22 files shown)


bassosimone:
So, changes to the workflow seem to work as far as _building_ is concerned. We cannot test _publishing_ easily until we bless a release. Also, it seems that `qafbmessenger` fails, but this is expected (see https://github.com/ooni/probe/issues/2128).

PR #832, opened 2022-07-05
fix(oohelperd): use throw-away HTTPClient, Dialer, Resolver

bassosimone:
This diff modifies the implementation of oohelperd in the release/3.15
branch to always use throw-away HTTPClient, Dialer, and Resolver.

The rationale of this change is to ensure we're not hitting limits of the
HTTPClient regarding the max number of connections per host.

This issue is described at https://github.com/ooni/probe/issues/2182.

While there, it feels more correct to use throw-away Dialer and Resolver.

While there, ensure we use the 8.8.8.8:53 resolver for the dialer and
the HTTPClient, which previously we were not doing 🤦.

We will need a different patch for `master` because there we have done
lots of refactoring of code using netx and netxlite (but the diff is going
to be more obvious to understand, so...)

### DIFF
--- a/internal/cmd/oohelperd/oohelperd.go
+++ b/internal/cmd/oohelperd/oohelperd.go
@@ -16,13 +16,13 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/runtimex"
 )
 
+// TODO(bassosimone): we should refactor this package to be flat w/o the internal
+// package. We're now past the many-small-packages time.
+
 const maxAcceptableBody = 1 << 24
 
 var (
-	dialer    model.Dialer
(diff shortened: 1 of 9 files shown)


internal/cmd/oohelperd/oohelperd.go
```diff
@@ -16,27 +16,27 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/runtimex"
 )
 
+// TODO(bassosimone): we should refactor this package to be flat w/o the internal
+// package. We're now past the many-small-packages time.
+
```
bassosimone, line 21:
```suggestion
```

Done here https://github.com/ooni/probe-cli/pull/834 and just for master

PR #837, opened 2022-07-08
refactor(engine): set options from any value

bassosimone:
This diff refactors how we set options for experiments to accept
in input an any value or a map[string]any, depending on which method
we choose to actually set options.

There should be no functional change, except that now we're not
guessing the type and then attempting to set the value of the selected
field: now, instead, we match the provided type and the field's type
as part of the same function (i.e., SetOptionAny).

This diff is functional to https://github.com/ooni/probe/issues/2184,
because it will allow us to load options from a map[string]any,
which will be part of the OONI Run v2 JSON descriptor.

If we didn't apply this change, we would only have been to set options
from a map[string]string, which is good enough as a solution for the
CLI but is definitely clumsy when you have to write stuff like:

```JSON
{
  "options": {
    "HTTP3Enabled": "true"
  }
}
```

when you could instead more naturally write:

```JSON
{
  "options": {
    "HTTP3Enabled": true
  }
}
```

### DIFF
--- a/internal/engine/experimentbuilder.go
+++ b/internal/engine/experimentbuilder.go
@@ -85,10 +85,10 @@ var (
 	// to SetOptionaAny is not a valid string representation of a bool.
 	ErrInvalidStringRepresentationOfBool = errors.New("invalid string representation of bool")
 
-	// ErrCannotSetBoolOption means SetOptionAny could set a bool option.
+	// ErrCannotSetBoolOption means SetOptionAny couldn't set a bool option.
 	ErrCannotSetBoolOption = errors.New("cannot set bool option")
 
-	// ErrCannotSetStringOption means SetOptionAny could set a string option.
+	// ErrCannotSetStringOption means SetOptionAny couldn't set a string option.
 	ErrCannotSetStringOption = errors.New("cannot set string option")
 
 	// ErrUnsupportedOptionType means we don't support the type passed to
--- a/internal/engine/experimentbuilder.go
+++ b/internal/engine/experimentbuilder.go
@@ -117,6 +117,64 @@ func (b *ExperimentBuilder) Options() (map[string]OptionInfo, error) {
 	return result, nil
 }
 
+// setOptionBool sets a bool option.
+func (b *ExperimentBuilder) setOptionBool(field reflect.Value, value any) error {
+	switch v := value.(type) {
+	case bool:
+		field.SetBool(v)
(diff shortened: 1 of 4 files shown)


internal/engine/experimentbuilder.go
```diff
@@ -63,22 +62,50 @@ func (b *ExperimentBuilder) InputPolicy() InputPolicy {
...
+	ErrNoSuchField = errors.New("no such field")
+
+	// ErrCannotSetIntegerOption means SetOptionAny could set an integer option.
+	ErrCannotSetIntegerOption = errors.New("cannot set integer option")
+
+	// ErrInvalidStringRepresentationOfBool indicates the string you passed
+	// to SetOptionaAny is not a valid string representation of a bool.
+	ErrInvalidStringRepresentationOfBool = errors.New("invalid string representation of bool")
+
+	// ErrCannotSetBoolOption means SetOptionAny could set a bool option.
```
bassosimone, line 88:
```suggestion
	// ErrCannotSetBoolOption means SetOptionAny couldn't set a bool option.
```

internal/engine/experimentbuilder.go
```diff
@@ -63,22 +62,50 @@ func (b *ExperimentBuilder) InputPolicy() InputPolicy {
...
+	ErrCannotSetBoolOption = errors.New("cannot set bool option")
+
+	// ErrCannotSetStringOption means SetOptionAny could set a string option.
```
bassosimone, line 91:
```suggestion
	// ErrCannotSetStringOption means SetOptionAny couldn't set a string option.
```

internal/engine/experimentbuilder.go
```diff
@@ -63,22 +62,50 @@ func (b *ExperimentBuilder) InputPolicy() InputPolicy {
...
 }
 
+var (
+	// ErrConfigIsNotAStructPointer indicates we expected a pointer to struct.
+	ErrConfigIsNotAStructPointer = errors.New("config is not a struct pointer")
+
+	// ErrNoSuchField indicates there's no field with the given name.
+	ErrNoSuchField = errors.New("no such field")
+
+	// ErrCannotSetIntegerOption means SetOptionAny could set an integer option.
```
bassosimone, line 81:
```suggestion
	// ErrCannotSetIntegerOption means SetOptionAny couldn't set an integer option.
```

PR #846, opened 2022-07-22
feat: TRR2 resolver 

DecFox:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2185

## Description

This intoduces a TRR2 resolver to measurexlite allowing us to opportunistcally test for DoH and make use of it whenever we can. The resolver uses the system resolver (`getaddrinfo`) as a fallback when DoH fails

### DIFF
--- a/internal/measurexlite/resolver.go
+++ b/internal/measurexlite/resolver.go
@@ -1,0 +1,106 @@
+package measurexlite
+
+//
+// Trusted Recursive Resolver
+//
+
+import (
+	"context"
+	"net"
+
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/netxlite"
+)
+
+// NewTrustedRecursiveResolver2 returns a new trusted recursive resolver
+func NewTrustedRecursiveResolver2(logger model.Logger, URL string) model.Resolver {
+	const mozillaURL = "https://mozilla.cloudflare-dns.com/dns-query"
+	if URL == "" {
+		URL = mozillaURL
+	}
+	return &TrustedRecursiveResolver2{
+		Logger:                            logger,
+		URL:                               URL,
+		NewParallelDNSOverHTTPSResolverFn: nil,
+		ResolverSystem:                    netxlite.NewStdlibResolver(logger),
+	}
+}
+
+// TrustedRecursiveResolver2 emulates Firefox's TRR2 mode
+type TrustedRecursiveResolver2 struct {
+	Logger model.Logger
+	URL    string
+
+	// NewParallelDNSOverHTTPSResolverFn is OPTIONAL and can be used to overide
+	// calls to the netxlite.NewParallelDNSOverHTTPSResolver factory
+	NewParallelDNSOverHTTPSResolverFn func(model.Logger, string) model.Resolver
+
+	// ResolverSystem is MANADATORY and is used as a fallback if the DoH resolver fails
+	ResolverSystem model.Resolver
+}
+
+var _ model.Resolver = &TrustedRecursiveResolver2{}
+
+// Address implements model.Resolver.Address
+// we can't return a decisive address here since we don't know if LookupHost
+// succeeds with the DoH resolver or the system resolver
+func (trr *TrustedRecursiveResolver2) Address() string {
+	return ""
+}
+
+// Network implements model.Resolver.Network
+// we can't return a decisive network here since we don't know if LookupHost
+// succeeds with the DoH resolver or the system resolver
+func (trr *TrustedRecursiveResolver2) Network() string {
+	return ""
+}
+
+// CloseIdleConnections implements model.Resolver.CloseIdleConnections
+// we only check the fallback resolver since DoH resolver is generated on the fly
+func (trr TrustedRecursiveResolver2) CloseIdleConnections() {
+	trr.ResolverSystem.CloseIdleConnections()
+}
+
+// LookupHost implements model.Resolver.LookupHost
+func (trr *TrustedRecursiveResolver2) LookupHost(ctx context.Context, hostname string) ([]string, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
+	addrs, err := reso.LookupHost(ctx, hostname)
+	if err != nil {
+		addrs, err = trr.ResolverSystem.LookupHost(ctx, hostname)
+	}
+	return addrs, err
+}
+
+// LookupHTTPS implements model.Resolver.LookupHTTPS
+func (trr *TrustedRecursiveResolver2) LookupHTTPS(ctx context.Context, domain string) (*model.HTTPSSvc, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
+	out, err := reso.LookupHTTPS(ctx, domain)
+	if err != nil {
+		out, err = trr.ResolverSystem.LookupHTTPS(ctx, domain)
+	}
+	return out, err
+}
+
+// LookupNS implements model.Resolver.LookupNS
+func (trr *TrustedRecursiveResolver2) LookupNS(ctx context.Context, domain string) ([]*net.NS, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
+	out, err := reso.LookupNS(ctx, domain)
+	if err != nil {
+		out, err = trr.ResolverSystem.LookupNS(ctx, domain)
+	}
+	return out, err
+}
+
+// Url returns the configured DoH URL for the TRR2 resolver
+func (trr *TrustedRecursiveResolver2) Url() string {
+	return trr.URL
+}
--- a/internal/measurexlite/resolver_test.go
+++ b/internal/measurexlite/resolver_test.go
@@ -1,0 +1,163 @@
+package measurexlite
+
+import (
+	"context"
+	"errors"
+	"net"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/model/mocks"
+)
+
+func TestNewTrustedRecursiveResolver(t *testing.T) {
+	var called bool
+	newResolver := func(model.Logger, string) model.Resolver {
+		return &mocks.Resolver{
+			MockLookupHost: func(ctx context.Context, domain string) ([]string, error) {
+				return []string{"1.1.1.1"}, nil
+			},
+			MockLookupHTTPS: func(ctx context.Context, domain string) (*model.HTTPSSvc, error) {
+				return &model.HTTPSSvc{
+					IPv4: []string{"1.1.1.1"},
+				}, nil
+			},
+			MockLookupNS: func(ctx context.Context, domain string) ([]*net.NS, error) {
+				return []*net.NS{
+					{
+						Host: "1.1.1.1",
+					},
+				}, nil
+			},
+			MockCloseIdleConnections: func() {
+				called = true
+			},
+		}
+	}
+
+	t.Run("default URL", func(t *testing.T) {
+		expected := "https://mozilla.cloudflare-dns.com/dns-query"
+		resolver := NewTrustedRecursiveResolver2(&mocks.Logger{}, "")
+		resolvert := resolver.(*TrustedRecursiveResolver2)
+		if resolvert.Url() != expected {
+			t.Fatal("unexpected default url")
+		}
+	})
+
+	t.Run("custom URL", func(t *testing.T) {
+		expected := "https://dns.google/dns-query"
+		resolver := NewTrustedRecursiveResolver2(&mocks.Logger{}, "https://dns.google/dns-query")
+		resolvert := resolver.(*TrustedRecursiveResolver2)
+		if resolvert.Url() != expected {
+			t.Fatal("unexpected default url")
+		}
+	})
+
+	t.Run("with DoH resolver", func(t *testing.T) {
+		resolver := NewTrustedRecursiveResolver2(&mocks.Logger{}, "")
+		resolvert := resolver.(*TrustedRecursiveResolver2)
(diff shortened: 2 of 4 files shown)


internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"context"
+	"net"
+
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/netxlite"
+)
+
+// NewTrustedRecursiveResolver2 returns a new trusted recursive resolver
```
bassosimone, line 15:
```suggestion
// NewTrustedRecursiveResolver2 returns a new trusted recursive resolver.
```

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return &TrustedRecursiveResolver2{
+		Logger:                            logger,
+		URL:                               URL,
+		NewParallelDNSOverHTTPSResolverFn: nil,
+		ResolverSystem:                    netxlite.NewStdlibResolver(logger),
+	}
+}
+
+// TrustedRecursiveResolver2 emulates Firefox's TRR2 mode
```
bassosimone, line 29:
```suggestion
// TrustedRecursiveResolver2 emulates Firefox's TRR2 mode.
```

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+type TrustedRecursiveResolver2 struct {
+	Logger model.Logger
+	URL    string
```
bassosimone, line 32: Please, document `Logger` and `URL` (and mention they're `MANDATORY`)

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+	ResolverSystem model.Resolver
+}
+
+var _ model.Resolver = &TrustedRecursiveResolver2{}
+
+// Address implements model.Resolver.Address
+// we can't return a decisive address here since we don't know if LookupHost
+// succeeds with the DoH resolver or the system resolver
+func (trr *TrustedRecursiveResolver2) Address() string {
+	return ""
```
bassosimone, line 48: How about returning the URL here?

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Network implements model.Resolver.Network
+// we can't return a decisive network here since we don't know if LookupHost
+// succeeds with the DoH resolver or the system resolver
+func (trr *TrustedRecursiveResolver2) Network() string {
+	return ""
```
bassosimone, line 55: How about returning `"trr2"` here?

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// CloseIdleConnections implements model.Resolver.CloseIdleConnections
+// we only check the fallback resolver since DoH resolver is generated on the fly
```
bassosimone, line 59:
```suggestion
```

I think this is an implementation detail, so we can omit the comment

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+func (trr TrustedRecursiveResolver2) CloseIdleConnections() {
+	trr.ResolverSystem.CloseIdleConnections()
+}
+
+// LookupHost implements model.Resolver.LookupHost
+func (trr *TrustedRecursiveResolver2) LookupHost(ctx context.Context, hostname string) ([]string, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
```
bassosimone, line 69: I think this code should read like:

```Go
reso := trr.newParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
```

where `newParallelDNSOverHTTPSResolver` should read:

```Go
func (trr *TrustedRecursiveResolver2) newParallelDNSOverHTTPSResolver(logger model.Logger, URL string) model.Resolver {
  factory := trr.NewParallelDNSOverHTTPSResolverFn
  if factory == nil {
    factory = netxlite.NewParallelDNSOverHTTPSResolver
  }
  return factory(logger, URL)
}
```

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+	return addrs, err
+}
+
+// LookupHTTPS implements model.Resolver.LookupHTTPS
+func (trr *TrustedRecursiveResolver2) LookupHTTPS(ctx context.Context, domain string) (*model.HTTPSSvc, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
```
bassosimone, line 82: This code should be refactored as suggested for `LookupHost`

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+	return out, err
+}
+
+// LookupNS implements model.Resolver.LookupNS
+func (trr *TrustedRecursiveResolver2) LookupNS(ctx context.Context, domain string) ([]*net.NS, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
```
bassosimone, line 95: This code should be refactored as suggested for `LookupHost`

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+	out, err := reso.LookupNS(ctx, domain)
+	if err != nil {
+		out, err = trr.ResolverSystem.LookupNS(ctx, domain)
+	}
+	return out, err
+}
+
+// Url returns the configured DoH URL for the TRR2 resolver
```
bassosimone, line 103: At this point while reading the diff, I am failing to understand why we need this method given that `URL` is already a public attribute of the structure and given that, if we use a `model.Resolver`, we cannot access the `Url` method anyway.

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
+package measurexlite
+
+//
+// Trusted Recursive Resolver
+//
```
bassosimone, line 5: Either call the file `trr2.go` and keep this comment or call this file `resolver.go` and say it contains resolvers

internal/measurexlite/resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+	"net"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/model/mocks"
+)
+
+func TestNewTrustedRecursiveResolver(t *testing.T) {
+	var called bool
```
bassosimone, line 15: Please, put suitable copies of this variable in the narrowest possible scope to avoid mistakes caused by the variable in a subtest using a value previously computed by another subtest.

internal/measurexlite/resolver.go
bassosimone: You should document that `logger` is `MANDATORY` while `URL` is `OPTIONAL` and you should say what happens if you provide an empty `URL`. You should probably say, for future robustness, that we're going to choose a suitable URL (or you could also be specific, up to you).

internal/measurexlite/resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+		resolver := NewTrustedRecursiveResolver2(&mocks.Logger{}, "https://dns.google/dns-query")
+		resolvert := resolver.(*TrustedRecursiveResolver2)
+		if resolvert.Url() != expected {
+			t.Fatal("unexpected default url")
+		}
+	})
+
+	t.Run("with DoH resolver", func(t *testing.T) {
+		resolver := NewTrustedRecursiveResolver2(&mocks.Logger{}, "")
+		resolvert := resolver.(*TrustedRecursiveResolver2)
```
bassosimone, line 59:
```suggestion
		resolver := NewTrustedRecursiveResolver2(&mocks.Logger{}, "").(*TrustedRecursiveResolver2)
```

This should save you a variable.

internal/measurexlite/resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+		resolvert.NewParallelDNSOverHTTPSResolverFn = newResolver
+		ctx := context.Background()
+		t.Run("LookupHost", func(t *testing.T) {
```
bassosimone, line 62:
```suggestion

		t.Run("LookupHost", func(t *testing.T) {
```

Please always add spaces around `t.Run` invocations. This allows for easy navigation of blocks using vi's `{` and `}` commands.

internal/measurexlite/resolver_test.go
bassosimone: (I see how in the code you only use it once, but I'd like to recommend you probably just create the resolver you need each time rather than having too much shared state for tests. The issue with shared state is that it's not robust to refactoring, so I'd rather just initialize each time the specific calls I need and leave others as `nil`)

internal/engine/probeservices/measurementmeta_test.go
```diff
@@ -52,7 +52,7 @@ func TestGetMeasurementMetaWorkingAsIntended(t *testing.T) {
 	if mmeta.Input == nil || *mmeta.Input != config.Input {
 		t.Fatal("unexpected input value")
 	}
-	if mmeta.MeasurementStartTime.String() != "2020-12-09 05:22:25 +0000 UTC" {
+	if mmeta.MeasurementStartTime.String() != "2020-12-09 04:22:25 +0000 UTC" {
```
bassosimone, line 55: I don't fully understand why this change is necessary and the commit does not seem to explain in detail why this was behaving in a flaky way. Would you mind clarifying?

bassosimone (changes requested):
I like the work you did so far. As I mention to you in private, we probably need to figure out some design details around this functionality, which won't cause dramatic changes, so we can discuss them in parallel with the code review proper. I highlighted a bunch of places in which I think you should tweak/refactor your code for improved robustness.

Additionally, I've seen quite a large coverage drop (-0.5%) which most likely means that we have added a bit too much code that we're not testing. I would recommend for you to check which are the new functions that we're not testing and add tests.

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+	trr.ResolverSystem.CloseIdleConnections()
+}
+
+// LookupHost implements model.Resolver.LookupHost
+func (trr *TrustedRecursiveResolver2) LookupHost(ctx context.Context, hostname string) ([]string, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
+	addrs, err := reso.LookupHost(ctx, hostname)
```
bassosimone, line 70: I forgot to mention: if we want to implement TRR2 correctly, this lookup should be strictly time bounded (i.e., we should timeout the lookup after N seconds have elapsed)

internal/measurexlite/resolver.go
```diff
@@ -0,0 +1,106 @@
...
+
+// LookupHTTPS implements model.Resolver.LookupHTTPS
+func (trr *TrustedRecursiveResolver2) LookupHTTPS(ctx context.Context, domain string) (*model.HTTPSSvc, error) {
+	reso := trr.NewParallelDNSOverHTTPSResolverFn(trr.Logger, trr.URL)
+	if reso == nil {
+		reso = netxlite.NewParallelDNSOverHTTPSResolver(trr.Logger, trr.URL)
+	}
+	out, err := reso.LookupHTTPS(ctx, domain)
+	if err != nil {
+		out, err = trr.ResolverSystem.LookupHTTPS(ctx, domain)
```
bassosimone, line 85: Because the system resolver only implements `LookupHost`, we should consider implementing `model.SimpleResolver` rather than `model.Resolver`.

bassosimone (changes requested):
Additional issues that I discovered while writing a design document for this feature

internal/measurexlite/resolver.go
DecFox: You are right, this is more robust. Thanks!

internal/measurexlite/resolver.go
DecFox: Sorry, this was actually some experimental code I forgot to remove. Thanks for pointing this out. I have removed it in the new commit.

internal/measurexlite/resolver_test.go
DecFox: You are right. Thanks!

internal/engine/probeservices/measurementmeta_test.go
DecFox: There was a bug in the data pipeline (as I had informed in our 1:1). I have reverted this back since the bug has been fixed now: https://github.com/ooni/probe-cli/pull/846/commits/f71c4e29d5c413254ef6d37884abe2bf0a86bdfe

internal/measurexlite/resolver.go
DecFox: I have added a timeout to the DoH resolver here: https://github.com/ooni/probe-cli/pull/846/commits/286c78f1b0a2fadabd2119f89180cea43d8c8476#diff-4c58fee97f07a0f47542cb08705a81d5f43f76be1d6235538ceb3e4c51725e72R33,
this is currently configured to use Firefox's default timeout value of 1.5s

internal/measurexlite/resolver.go
DecFox: > I like the work you did so far. As I mention to you in private, we probably need to figure out some design details around this functionality, which won't cause dramatic changes, so we can discuss them in parallel with the code review proper. I highlighted a bunch of places in which I think you should tweak/refactor your code for improved robustness.
> 
> Additionally, I've seen quite a large coverage drop (-0.5%) which most likely means that we have added a bit too much code that we're not testing. I would recommend for you to check which are the new functions that we're not testing and add tests.

Thank you for the feedback. I have made some changes based on the review. I also added some more tests but the coverage still seems to be decreased, so we probably need to add more tests.

DecFox:
> I like the work you did so far. As I mention to you in private, we probably need to figure out some design details around this functionality, which won't cause dramatic changes, so we can discuss them in parallel with the code review proper. I highlighted a bunch of places in which I think you should tweak/refactor your code for improved robustness.
> 
> Additionally, I've seen quite a large coverage drop (-0.5%) which most likely means that we have added a bit too much code that we're not testing. I would recommend for you to check which are the new functions that we're not testing and add tests.

Thanks you for the feedback! I have made some changes based on the review. I have also added more tests, but the coverage still seems to be decreased. So we probably need to write some more.

PR #847, opened 2022-08-02
Feat/dnsscan

hellais:
Added an experiment for scanning open DNS resolvers to detect blocking.

Currently experimental and as such doesn't have a spec for it yet.

Here is a sample of the output of a test run:
```
{
  "annotations": {
    "architecture": "amd64",
    "engine_name": "ooniprobe-engine",
Add dnsscan to allexperiments
    "engine_version": "3.16.0-alpha",
    "platform": "macos"
  },
  "data_format_version": "0.2.0",
  "input": "ooni.org",
  "measurement_start_time": "2022-08-02 11:22:37",
  "options": [
    "Resolver=udp://9.9.9.9:53"
  ],
  "probe_asn": "AS30722",
  "probe_cc": "IT",
  "probe_ip": "127.0.0.1",
  "probe_network_name": "Vodafone Italia S.p.A.",
  "report_id": "",
  "resolver_asn": "AS30722",
  "resolver_ip": "91.81.132.35",
  "resolver_network_name": "Vodafone Italia S.p.A.",
  "software_name": "miniooni",
  "software_version": "3.16.0-alpha",
  "test_keys": {
    "queries": [
      {
        "answers": [
          {
            "asn": 16509,
            "as_org_name": "Amazon.com, Inc.",
            "answer_type": "A",
            "ipv4": "75.2.60.5",
            "ttl": null
          },
          {
            "asn": 16509,
            "as_org_name": "Amazon.com, Inc.",
            "answer_type": "A",
            "ipv4": "99.83.231.61",
            "ttl": null
          }
        ],
        "engine": "udp",
        "failure": null,
        "hostname": "ooni.org",
        "query_type": "A",
        "resolver_hostname": null,
        "resolver_port": null,
        "resolver_address": "9.9.9.9:53",
        "t": 0.132053
      },
      {
        "answers": null,
        "engine": "udp",
        "failure": "dns_no_answer",
        "hostname": "ooni.org",
        "query_type": "AAAA",
        "resolver_hostname": null,
        "resolver_port": null,
        "resolver_address": "9.9.9.9:53",
        "t": 0.102853
      }
    ],
    "target_resolver": "udp://9.9.9.9:53"
  },
  "test_name": "dnsscan",
  "test_runtime": 0.132149438,
  "test_start_time": "2022-08-02 11:22:37",
  "test_version": "0.1.0"
}
```

### DIFF
--- a/internal/engine/experiment/dnsscan/dnsscan.go
+++ b/internal/engine/experiment/dnsscan/dnsscan.go
@@ -1,0 +1,151 @@
+// Package dnsscan is the experimental dnsscan experiment.
+package dnsscan
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"time"
+
+	"github.com/miekg/dns"
+	"github.com/ooni/probe-cli/v3/internal/measurexlite"
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/netxlite"
+)
+
+const (
+	testName    = "dnsscan"
+	testVersion = "0.1.0"
+)
+
+// TestKeys contains the experiment results.
+type TestKeys struct {
+	Queries        []*model.ArchivalDNSLookupResult `json:"queries"`
+	TargetResolver string                           `json:"target_resolver"`
+}
+
+// Config contains the experiment configuration.
+type Config struct {
+	// Address of the DNS resolver to be used for testing
+	Resolver string `ooni:"address of the DNS resolver to use"`
+}
+
+func (c Config) resolver() string {
+	if c.Resolver != "" {
+		return c.Resolver
+	}
+	return "udp://8.8.8.8:53"
+}
+
+// Measurer performs the measurement.
+type Measurer struct {
+	config Config
+}
+
+// ExperimentName implements ExperimentMeasurer.ExperiExperimentName.
+func (m *Measurer) ExperimentName() string {
+	return testName
+}
+
+// ExperimentVersion implements ExperimentMeasurer.ExperimentVersion.
+func (m *Measurer) ExperimentVersion() string {
+	return testVersion
+}
+
+var (
+	// errNoInputProvided indicates you didn't provide any input
+	errNoInputProvided = errors.New("not input provided")
+
+	// errInputFormatInvalid means that the input format is not valid
+	errInputFormatInvalid = errors.New("input is not a domain or URL")
+
+	// errNoInputProvided indicates that the chosen DNS resolver is not currently supported
+	errResolverNotSupported = errors.New("resolver not supported")
+
+	// errInvalidResolver indicates that the chosen DNS resolver is invalid
+	errInvalidResolver = errors.New("resolver address is invalid")
+
+	// errMissingPort indicates that there is no port.
+	errMissingPort = errors.New("the URL must include a port")
+)
+
+// Run implements ExperimentMeasurer.Run.
+func (m *Measurer) Run(
+	ctx context.Context,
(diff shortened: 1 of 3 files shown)


hellais:
One doubt that this test makes me think of is that we are writing to the `resolver_*` top level keys several bits of information that are not accurate to this test.

If somebody is to interpret this test as they would any other one, they would mistakenly attribute the results of the `queries` object to those listed inside of these keys, while the ought to instead be looking at the `resolver_address` inside of the queries.

Perhaps it's worth adding some additional information to our spec to make it explicit and clear that top level `resolver_*` keys are only relevant to a `system` type `engine` and when you are faced with any other engine you should instead be looking at the `resolver_address`.
Speaking of which, in these cases it might be beneficial to enrich the data with metadata about the resolver, similarly to what we do with the `resolver_ip`.

internal/engine/experiment/dnsscan/dnsscan.go
```diff
@@ -0,0 +1,151 @@
...
+		return errResolverNotSupported
+	}
+	if parsedResolver.Port() == "" {
+		return errMissingPort
+	}
+	resolverIP := net.ParseIP(parsedResolver.Hostname())
+	if resolverIP == nil {
+		return fmt.Errorf("%w: invalid IP address", errInvalidResolver)
+	}
+	resolverAddress := fmt.Sprintf("%s:%s", resolverIP, parsedResolver.Port())
```
bassosimone, line 107:
```suggestion
	resolverAddress := parsedResolver.Host
```

bassosimone: or, `net.JoinHostPort(resolverIP, parsedResolver.Port())` but you already have this value, so it's better to just use it from the URL rather than reconstructing it.

internal/engine/experiment/dnsscan/dnsscan.go
```diff
@@ -0,0 +1,151 @@
...
+	// is an IPv6 address
+	if resolverIP.To4() == nil {
+		resolverAddress = fmt.Sprintf("[%s]:%s", resolverIP, parsedResolver.Port())
+	}
```
bassosimone, line 111: unnecessary if you rely on URL parsing as mentioned here: https://github.com/ooni/probe-cli/pull/847/files#r935489263

internal/engine/experiment/dnsscan/dnsscan.go
```diff
@@ -0,0 +1,151 @@
...
+	m.dnsRoundTrip(ctx, measurement.MeasurementStartTimeSaved, sess.Logger(), resolverAddress, domain, tk)
+	return nil // return nil so we always submit the measurement
+}
+
+// dnsRoundTrip performs a round trip and returns the results to the caller.
+func (m *Measurer) dnsRoundTrip(ctx context.Context, zeroTime time.Time,
+	logger model.Logger, address string, domain string, tk *TestKeys) {
+	fmt.Printf("Measuing %s %s", domain, address)
```
bassosimone, line 119:
```suggestion
	fmt.Printf("Measuring %s %s", domain, address)
```

internal/engine/experiment/dnsscan/dnsscan.go
```diff
@@ -0,0 +1,151 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+	trace := measurexlite.NewTrace(0, zeroTime)
+	ol := measurexlite.NewOperationLogger(logger, "DNSScan %s %s", address, domain)
+
```
bassosimone, line 124:
```suggestion
```

PR #849, opened 2022-08-04
feat: expose engine FFI API to a Dart consumer (prototype 1)

bassosimone:
We are currently exploring what would happen if we attempted to
rewrite the following user-facing tools using Flutter:

* ooniprobe-android
* ooniprobe-desktop
* ooniprobe-ios

The general idea is to have a common dart library used by these
prodiucts (let's call this hypotetical library probe-commons).

Based on the current state of our trees, we will then have:

* ooniprobe-android written in flutter that uses probe-commons that
implements a driver to interact with oonimkall using Java;

* ooniprobe-ios written in flutter that uses probe-commons that implements
a driver to interact with oonimkall using ObjectiveC;

* ooniprobe-desktop written in flutter that uses probe-commons that
implements a driver to use ooniprobe-cli like we do now.

Because our original goal was to reduce code duplication, we then started
wondering whether we could get rid of these drivers and instead find
a better way for probe-commons to communicate with the OONI engine.

This diff in particular explores a possible way in which probe-commons
could communicate with the engine, with the objective of evaluating
whether this is a good solution that could help us, or whether it would
be better to explore other possibilities.

The idea we're exploring here is whether the engine could export
an FFI API that we can in turn use via `dart:ffi`.

Exposing an FFI API is something we used to do with Measurement Kit,
so here we have a similar API, even though it's not equal. A bunch of
details changed to allow for a smoother FFI experience.

This part of the problem seems solved because we already knew how to
solve the problem and `dart:ffi` means to manual work for us.

A second aspect of the problem is that, when you do FFI, unless you
aim to automatically wrap a very complex API, you're depending on
defining which messages to exchange over the FFI interface.

Ideally, you want some canonical definition of those messages and
then you want to automatically generate the messages such that when
you change the ABI your build basically breaks.

There were three options here:

1. use JSON and generate Dart code from Go code we actually use;

2. use JSON and generate Dart and Go code from a spec that is otherwise
neither used by the Dart nor by the Go implementation;

3. forget about JSON and use protobuf (which looks like the best
binary serializatin format for our use case, given that Go, protobuf,
and Dart are all Google projects).

In this diff, I have chosen the second approach. It's possible that in
the future there would be pull requests exploring other approaches.

Once you have the FFI API and you autogenerate the ABI, it's then
possible to additionally auto-generate some extra code in Dart
to avoid maintaining too much boilerplate.

As a concluding remark, a completely different but still viable strategy
here would be to explore whether this unofficial flutter extension:

    https://github.com/go-flutter-desktop/go-flutter

could also be useful for solving the problem we're focusing on.

Reference issue: https://github.com/ooni/probe/issues/2197


bassosimone:
Closing in favour of https://github.com/ooni/probe-cli/pull/851

PR #850, opened 2022-08-05
feat: expose engine FFI API to a Dart consumer (prototype 2)

bassosimone:
This diff is a variation of https://github.com/ooni/probe-cli/pull/849
where we use protobuf instead of a custom JSON message format
managed by us. Therefore, this diff allows us to evaluate which
is the cost of inventing a custom JSON data format.

At the FFI level, the most relevant change is that we are not
passing along UTF-8 strings anymore, rather we're now using
binary buffers. I spent some (but not too much) trying to see
whether we could actually avoid unnecessary copies. I could
do that for events but could not do that for messages. I think
it's fair enough since events are more frequent and we're not
exchanging super large messages anyway.

Also, the whole Dart codebase is a bit leaner because in the
meanwhile I leveled up a bit :^).

Reference issue https://github.com/ooni/probe/issues/2197


bassosimone:
(For the records, I also experimented with `cgo.Handle`, but I feel safer with a more forgiving solution such as the table that we're currently using, as opposed to an implementation that panics more frequently.)

bassosimone:
Closing in favour of https://github.com/ooni/probe-cli/pull/851.

PR #851, opened 2022-08-05
feat: expose engine FFI API to a Dart consumer (prototype 3)

bassosimone:
The general idea here is to avoid returning -1, having a more
uniform API, and allowing for sync calls.

It's a bit more work on both sides but it's more regular, so maybe
this change is for the greater good?

This will be a third pull request because it varies from
https://github.com/ooni/probe-cli/pull/850 mostly in
the way in which we interface at the FFI level.

Reference issue: https://github.com/ooni/probe/issues/2197


dart/ooniengine/pubspec.lock
```diff
@@ -0,0 +1,145 @@
+# Generated by pub
```
ainghazal, line 1: perhaps you want to ignore this one too?

mk
```diff
@@ -339,6 +339,80 @@ GOMOBILE = $(shell go env GOPATH)/bin/gomobile
 OONIMKALL_V := $(shell date -u +%Y.%m.%d-%H%M%S)
 OONIMKALL_R := $(shell git describe --tags || echo '0.0.0-dev')
 
+#help:
```
ainghazal, line 342: leaving here a couple alternatives to make that I've got in my radar (you probably know them already):

* https://github.com/fugue/zim/
* https://magefile.org/

pkg/ooniengine/geoip.go
```diff
@@ -0,0 +1,60 @@
...
+
+// geoIPRunner is the geoip task runner.
+type geoIPRunner struct{}
+
+var _ taskRunner = &geoIPRunner{}
+
+// main implements taskRunner.main.
+func (r *geoIPRunner) main(ctx context.Context, emitter taskMaybeEmitter, args []byte) {
+	logger := newTaskLogger(emitter)
+	var config abi.GeoIPConfig
```
ainghazal, line 31: pretty obvious, but it seems like there's some boilerplate here in the session init that can be factored out :)

dart/ooniengine/pubspec.lock
bassosimone: From https://dart.dev/tools/pub/glossary#lockfile

> A file named pubspec.lock that specifies the concrete versions and other identifying information for every immediate and transitive dependency a package relies on.
>
> Unlike the pubspec, which only lists immediate dependencies and allows version ranges, the lock file comprehensively pins down the entire dependency graph to specific versions of packages. A lockfile ensures that you can recreate the exact configuration of packages used by an application.
>
> The lockfile is generated automatically for you by pub when you run [pub get](https://dart.dev/tools/pub/cmd/pub-get), [pub upgrade](https://dart.dev/tools/pub/cmd/pub-upgrade), or [pub downgrade](https://dart.dev/tools/pub/cmd/pub-downgrade). If your package is an application package, you will typically check this into source control. For library packages, you usually won’t.

Because this is mostly a library, I'd say we want to commit it.

PR #863, opened 2022-08-17
cleanup: chmod -x mk && mv mk Makefile

bassosimone:
While there, document what we need to document without mentioning too
many details about release builds, for which we have the Makefile.

See https://github.com/ooni/probe/issues/2218.


bassosimone:
We're still having issues with the API. I am going to merge nonetheless 🤟

PR #873, opened 2022-08-22
feat(dns): expose more low-level fields

bassosimone:
This pull request started as a draft to enable users to see CNAME answers. It contained several patches which we merged separately (see https://github.com/ooni/probe-cli/pull/873#issuecomment-1222406732 and https://github.com/ooni/probe-cli/compare/2301a30630577637e5ec9912b027584e5ab4eebc...60b7d1f87be4b088929048edfb950dc93146a01f for details on what has actually changed, which is based on patches originally part of this PR). In its final form, however, this PR only deals with exposing more low-level DNS fields to the archival data format.

Closes: https://github.com/ooni/probe/issues/2228

Related PR spec: https://github.com/ooni/spec/pull/256


bassosimone:
Here's what is currently missing:

- [ ] move trace based code from `resolvercore.go` to `dnsovergetaddrinfo.go`
- [ ] (probably) remove `addrs` from the arguments passed to DNS callbacks
- [ ] modify `measurexlite` to extract from the `DNSResponse`

When all of this is done, `dnsping` (and any other `measurexlite` based code) can see the `CNAME`.

PR #891, opened 2022-08-28
feat: port-filtering experiment

DecFox:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2005
- [x] if you changed anything related how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: https://github.com/ooni/spec/pull/260

## Description

This is a premature implementation of the port-filtering experiment.


hellais:
I took a look at the PR and did some light testing of it and it looks good to me.

Some high level comments which we should keep in mind before deploying this include:

1. We should consider how we are going to distinguish between port blocking and the blocking of the IP of the test helper. Ideas for how we might go about doing this include (possibly a combination of them):
- Using a much larger set of ports, potentially with some randomised ones too and consider the whole experiment as invalid if we notice that no port is reachable
- Using something like MLab so we have a larger pools of IPs available
- Using ICMP or UDP to test reachability with the test helper
2. We should extend the set of ports which we scan and possibly add support for parallelising it. According to nmap-services (https://github.com/nmap/nmap/blob/master/nmap-services) there are 70 ports which are open with a probability >= 2%. This is probably a reasonable cutoff point.
3. If we are to extend the list of ports, we might also want to add the ability to pass a port list as an input file (or a CLI flag). In doing so I wonder if the structure of having all the measurement in a single measurement file is going to scale well. That is to say if you do a port scan of 10k ports then you will end up with a very "fat" measurement file with all the port results in it, conversely we could generate one measurement entry per port and perhaps that would scale better. Input from @FedericoCeratto and @bassosimone on this point would be helpful.

internal/cmd/ooporthelper/main.go
```diff
@@ -0,0 +1,81 @@
...
+	<-ctx.Done()
+}
+
+func listenTCP(ctx context.Context, port string) {
+	defer srvWg.Done()
+	address := net.JoinHostPort("127.0.0.1", port)
+	listener, err := net.Listen("tcp", address)
+	if err != nil {
+		runtimex.PanicOnError(err, "net.Listen failed")
+	}
```
bassosimone, line 47:
```suggestion
	runtimex.PanicOnError(err, "net.Listen failed")
```

bassosimone: `runtimex.PanicOnError` is designed to avoid writing (and testing) an `if`

internal/cmd/ooporthelper/ports.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+//
+// List of ports we want to test for filtering
```
bassosimone, line 4:
```suggestion
// List of ports we want to use when testing
```

internal/engine/experiment/portfiltering/config.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// Config for the port-filtering experiment
+//
+
+import "time"
+
+// Config contains the experiment configuration.
+type Config struct {
+	// Delay is the delay between each repetition (in milliseconds).
+	Delay int64 `ooni:"number of milliseconds to wait before knocking each port"`
```
bassosimone, line 12:
```suggestion
	Delay int64 `ooni:"number of milliseconds to wait before testing each port"`
```

bassosimone (approved):
Thanks! This is a great starting point to begin experimenting!

🐳 💯 🔝 🙏

PR #893, opened 2022-08-28
refactor: spin geoipx off geolocate

bassosimone:
A bunch of packages (including oohelperd) just need the ability to
use MaxMind-like databases. They don't need the additional functionality
implemented by the geolocate package. Such a package, in fact, is
mostly (if not only) needed by the engine package.

Therefore, move code to query MaxMind-like databases to a separate
package, and avoid depending on geolocate in all the packages for
which it's sufficient to use geoipx.

Part of https://github.com/ooni/probe/issues/2240


internal/geoipx/geoipx.go
```diff
@@ -0,0 +1,49 @@
+// Package geoipx contains code to use MaxMind-like GeoIP libraries.
```
bassosimone, line 1:
```suggestion
// Package geoipx contains code to use the embedded MaxMind-like databases.
```

internal/geoipx/geoipx.go
```diff
@@ -0,0 +1,49 @@
...
+		return
+	}
+	asn = record.AutonomousSystemNumber
+	if record.AutonomousSystemOrganization != "" {
+		org = record.AutonomousSystemOrganization
+	}
+	return
+}
+
+func LookupCC(ip string) (cc string, err error) {
```
bassosimone, line 33:
```suggestion
// LookupCC maps [ip] to a country code.
func LookupCC(ip string) (cc string, err error) {
```

internal/geoipx/geoipx_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"testing"
+
+	"github.com/ooni/probe-cli/v3/internal/model"
+)
+
+const ipAddr = "8.8.8.8"
+
+func TestLookupASN(t *testing.T) {
+	t.Run("with IP address", func(t *testing.T) {
```
bassosimone, line 12:
```suggestion
	t.Run("with valid IP address", func(t *testing.T) {
```

PR #904, opened 2022-08-29
feat: build ooniprobe for armv6

bassosimone:
Part of https://github.com/ooni/probe/issues/1753.

While there, introduce a rule by which, if the branch is named `fullbuild` we run all possible builds. It helps to test all the builds without creating a release branch. Because release branches are protected, they cannot be deleted easily. On the contrary, the `fullbuild` branch can easily be disposed of.

### DIFF
--- a/.github/workflows/alltests.yml
+++ b/.github/workflows/alltests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - "release/**"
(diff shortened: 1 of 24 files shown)


bassosimone:
We still have a bunch of red builds, but the build that mattered (i.e., ooniprobe on linux/armv6) is green! 🚀

PR #915, opened 2022-08-31
feat(oonirun): improve tests

bassosimone:
See https://github.com/ooni/probe/issues/2184

While there, rename `runtimex.PanicIfFalse` to `runtimex.Assert` (it was about time...)

### DIFF
--- a/internal/model/experiment.go
+++ b/internal/model/experiment.go
@@ -281,3 +281,25 @@ type ExperimentOptionInfo struct {
 	// Type contains the type.
 	Type string
 }
+
+// ExperimentInputLoader loads inputs from local or remote sources.
+type ExperimentInputLoader interface {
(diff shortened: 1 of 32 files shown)


internal/model/experiment.go
```diff
@@ -281,3 +281,25 @@ type ExperimentOptionInfo struct {
...
+	// report ID field in case of success.
+	Submit(ctx context.Context, m *Measurement) error
+}
+
+// Saver saves a measurement on some persistent storage.
+type Saver interface {
+	SaveMeasurement(m *Measurement) error
+}
+
+// ExperimentInputProcessor processes inputs running the given experiment.
```
bassosimone, line 302:
```suggestion
// ExperimentInputProcessor processes inputs for an experiment.
```

PR #927, opened 2022-09-03
ghpublish: attempt to create a draft release

bassosimone:
If we don't publish the release using `gh release` and we default to
a pre-release, we're able to manually disable the pre-release flag
when needed without making this script more complex.

### DIFF
--- a/script/ghpublish.bash
+++ b/script/ghpublish.bash
@@ -18,8 +18,8 @@ fi
 
 set -x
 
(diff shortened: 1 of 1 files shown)


bassosimone:
Ah, this cannot work because the script assumes we're building a tag. So, we need to adopt another approach where we check whether we're a pre-release by parsing the structure of the tag _anyway_. 🤦

PR #938, opened 2022-09-07
QA: refactor, cleanup, fix, and simplify

bassosimone:
While working on https://github.com/ooni/probe/issues/2130, and specifically on the action item related to making sure all workflows are green, I was confronted with the complexity of the QA directory.

There's plenty of cleaning up and simplifying there. The original intent was to A/B test `miniooni` and `measurement_kit` to ensure they were behaving the same. We don't have this need anymore.

Rather, it seems the QA scripts have grown large and flaky, to the point that I am always tempted to ignore them. The underlying censorship engine, jafar, has also not been developed for quite some time.

So, the first step towards improve the QA infrastructure seems to be humble and acknowledge that we cannot realistically maintain these checks using jafar as a backend for so many experiments.

Let us focus on our most important experiment, Web Connectivity, and let us keep QA checks for it.

Additionally, let us simplify and cleanup QA as much as possible, though without introducing radical changes.

The end result is a QA for Web Connectivity that seems reasonable and runs in six minutes.


.github/workflows/qa.yml
```diff
@@ -1,14 +1,16 @@
-# Runs QA checks for the webconnectivity experiment
-name: "qawebconnectivity"
+# Runs quality assurance checks for the all experiments
```
bassosimone, line 1:
```suggestion
# Runs quality assurance checks
```

PR #957, opened 2022-09-12
feat(webconnectivity@v0.5): probe and TH can't connect => website down

bassosimone:
This diff introduces a special rule to avoid emitting null, null when all the connects failed in both the probe and the TH.

While there, recognize that the subset of null, null we're hunting actually deals with websites that are down, so change the internal naming to reflect that and make the code easier to read/understand.

See https://github.com/ooni/probe/issues/2299

### DIFF
--- a/internal/experiment/webconnectivity/testkeys.go
+++ b/internal/experiment/webconnectivity/testkeys.go
@@ -72,7 +72,7 @@ type TestKeys struct {
 	// ControlFailure contains the failure of the control experiment.
 	ControlFailure *string `json:"control_failure"`
 
-	// DNSFlags cdescribes specific DNS anomalies we observed.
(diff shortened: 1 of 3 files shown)


internal/experiment/webconnectivity/testkeys.go
```diff
@@ -72,7 +72,7 @@ type TestKeys struct {
 	// ControlFailure contains the failure of the control experiment.
 	ControlFailure *string `json:"control_failure"`
 
-	// DNSFlags contains DNS analysis flags.
+	// DNSFlags cdescribes specific DNS anomalies we observed.
```
bassosimone, line 75:
```suggestion
	// DNSFlags describes specific DNS anomalies we observed.
```

PR #958, opened 2022-09-13
feat(webconnectivity@v0.5): detect TLS misconfig for probe and TH

bassosimone:
See https://github.com/ooni/probe/issues/2300.

### DIFF
--- a/internal/experiment/webconnectivity/analysiscore.go
+++ b/internal/experiment/webconnectivity/analysiscore.go
@@ -230,7 +230,7 @@ func (tk *TestKeys) analysisWebsiteDownDetectTLSMisconfigured(logger model.Logge
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/experiment/webconnectivity/analysiscore.go
```diff
@@ -178,8 +187,60 @@ const (
...
+			// implementation of the test helper and Twisted names
+			//
+			// TODO(bassosimone): this is the obvious algorithm but maybe
+			// it's a bit too strict and there is a more lax version of
+			// the same algorithm that it's still acceptable?
+			return false
+		}
+	}
+
+	// only if we have had some TLS handshakes to connect
```
bassosimone, line 233:
```suggestion
	// only if we have had some TLS handshakes for both probe and TH
```

PR #959, opened 2022-09-13
feat(webconnectivity@v0.5): stream the response body

bassosimone:
Reference issue: https://github.com/ooni/probe/issues/2295

### DIFF
--- a/internal/netxlite/iox.go
+++ b/internal/netxlite/iox.go
@@ -10,7 +10,7 @@ import (
 	"io"
 )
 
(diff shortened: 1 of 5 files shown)


internal/netxlite/iox.go
```diff
@@ -10,6 +10,9 @@ import (
 	"io"
 )
 
+// TODO(bassosimone): integrate StreamAllContext from
```
bassosimone, line 13:
```suggestion
// TODO(bassosimone): consider integrating StreamAllContext from
```

PR #960, opened 2022-09-14
webconnectivity@v0.5: handle successful https chains

bassosimone:
This diff includes a rule to recover from the "measurement failed" state that kicks in when we have a chain of successful redirects from the client side leading to a webpage _and_ any URL in the chain uses HTTPS. See https://github.com/ooni/probe/issues/2307.

While there, fix `i/e/w/iox.go` to avoid triggering the `./script/nocopyreadall.bash` script.

### DIFF
--- a/internal/experiment/webconnectivity/analysiscore.go
+++ b/internal/experiment/webconnectivity/analysiscore.go
@@ -3,6 +3,7 @@ package webconnectivity
 import (
 	"fmt"
 	"net"
+	"net/url"
 
 	"github.com/ooni/probe-cli/v3/internal/model"
 )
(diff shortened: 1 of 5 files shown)


internal/experiment/webconnectivity/analysiscore.go
```diff
@@ -138,12 +139,20 @@ func (tk *TestKeys) analysisToplevel(logger model.Logger) {
...
+			"ACCESSIBLE: flags=%d accessible=%+v, blocking=%+v",
 			tk.BlockingFlags, tk.Accessible, tk.Blocking,
 		)
 
 	default:
-		if tk.analysisWebsiteDownDetectNoAddrs(logger) {
+		// NullNull remediation
+		//
+		// If we arrive here, the measurement has failed. However, there are a
+		// bunch of cases where we can still explain what happen by applying specific
```
bassosimone, line 150:
```suggestion
		// bunch of cases where we can still explain what happened by applying specific
```

internal/experiment/webconnectivity/analysiscore.go
```diff
@@ -180,24 +202,83 @@ func (tk *TestKeys) analysisToplevel(logger model.Logger) {
...
 
-// analysisWebsiteDownDetectTLSMisconfigured runs when .Blocking = nil and
+// analysisNullNullDetectSuccessfulHTTPS runs when .Blocking = nil and
+// .Accessible = nil to flag successul HTTPS measurements chains that
+// occurred regardless of whatever else could have gone wrong.
+//
+// We need all requests to be HTTPS because an HTTP request in the
+// chain breaks the ~reasonable assumption that our custom CA bundle
+// is enough to protect against MITM. Of course, when we use this
+// algorithm, we're not super well position to flag server-side blocking.
```
bassosimone, line 229:
```suggestion
// algorithm, we're not well positioned to flag server-side blocking.
```

internal/experiment/webconnectivity/analysiscore.go
```diff
@@ -180,24 +202,83 @@ func (tk *TestKeys) analysisToplevel(logger model.Logger) {
...
+//
+// Version 0.4 of the probe implemented a similar algorithm, which
+// however run before other checks. Version, 0.5 on the contrary, runs
```
bassosimone, line 232:
```suggestion
// however ran before other checks. Version, 0.5 on the contrary, runs
```

PR #972, opened 2022-10-03
fix(E2E): ensure miniooni.bash is WAI

bassosimone:
Closes https://github.com/ooni/probe/issues/2336

### DIFF
--- a/E2E/miniooni.bash
+++ b/E2E/miniooni.bash
@@ -9,18 +9,33 @@
 
 set -euxo pipefail
 
-backends=()
-backends+=( "https://api.ooni.io" )
-backends+=( "https://dvp6h0xblpcqp.cloudfront.net" )
-backends+=( "https://ams-pg-test.ooni.org" )
+rm -f E2E/o.jsonl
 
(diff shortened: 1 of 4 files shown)


bassosimone:
Before merging, I need to figure out the correct settings for cloudfront.

bassosimone:
> Before merging, I need to figure out the correct settings for cloudfront.

So, in https://github.com/ooni/probe-cli/pull/972/commits/2b23bb67373dfb5ff22ef7e6adb6d3975fe1361e we have chosen to test the two cloudfronts that would actually work. I have prepared an internal document where I suggest to my colleagues a bunch of improvements to improve the situation here in terms of complexity.

bassosimone:
After re-reading the diff, I am not satisfied. It's unclear whether we're testing the testing backend or the production frontend or the tunnels or all of this at the same time. We should make that clear.

bassosimone:
So, I've chosen to use the prod API. I think, so far, miniooni integration tests have been one of the best way we have to make sure we're not breaking production, therefore let's keep using them until something better emerges.

PR #978, opened 2022-10-31
Fix the signal root CA

hellais:
The previously added certificate was not for the trust anchor, but rather the leaf server certificate signed by the root CA.

Since signal updated to use the new CA on more endpoints, our validation was failing.

See https://github.com/ooni/probe/issues/2344,


bassosimone (approved):
🐳

PR #981, opened 2022-11-14
cli: add support for passing proxy command line parameter

hellais:
This implements the bare minimum needed to have circumvention proxy support in OONI Probe CLI.

The reference issue is https://github.com/ooni/probe/issues/1955. (Most of that issue is implemented, save for the fact that currently we do not have support for `http` and `https` proxies.)

While there, add to `Makefile` a rule for correctly building `ooniprobe` and `miniooni` for "this system" (i.e., the default `GOOS` and `GOARCH` on a system), because we needed this for testing this patch and we needed to figure out the commands instead.

### DIFF
--- a/CLI/go-build-generic
+++ b/CLI/go-build-generic
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [[ $# -ne 1 ]]; then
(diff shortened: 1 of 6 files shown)


bassosimone (approved):
🐳

PR #982, opened 2022-11-16
feat: introduce database type 

DecFox:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2352
- [ ] if you change code inside an experiment, make sure you bump its version number

## Description
This diff introduces a `Database` type struct which allows us to consolidate database behavior and abstract to an interface easily. It should also allows us to introduce a `NullDatabase` in the future.

### DIFF
--- a/cmd/ooniprobe/internal/ooni/ooni.go
+++ b/cmd/ooniprobe/internal/ooni/ooni.go
@@ -19,7 +19,6 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/legacy/assetsdir"
 	"github.com/ooni/probe-cli/v3/internal/model"
 	"github.com/pkg/errors"
-	"github.com/upper/db/v4"
 )
 
 // DefaultSoftwareName is the default software name.
@@ -33,7 +32,7 @@ var logger = log.WithFields(log.Fields{
 // ProbeCLI is the OONI Probe CLI context.
 type ProbeCLI interface {
 	Config() *config.Config
-	DB() db.Session
+	DB() *database.Database
 	IsBatch() bool
 	Home() string
 	TempDir() string
@@ -53,7 +52,7 @@ type ProbeEngine interface {
 // Probe contains the ooniprobe CLI context.
 type Probe struct {
 	config  *config.Config
-	db      db.Session
+	db      *database.Database
 	isBatch bool
 
 	home      string
@@ -86,7 +85,7 @@ func (p *Probe) Config() *config.Config {
 }
 
 // DB returns the database we're using
-func (p *Probe) DB() db.Session {
+func (p *Probe) DB() *database.Database {
 	return p.db
 }
 
@@ -180,7 +179,9 @@ func (p *Probe) Init(softwareName, softwareVersion, proxy string) error {
 
 	p.dbPath = utils.DBDir(p.home, "main")
 	log.Debugf("Connecting to database sqlite3://%s", p.dbPath)
-	db, err := database.Connect(p.dbPath)
+	db, err := database.NewDatabase(&database.DatabaseConfig{
+		DatabasePath: p.dbPath,
+	})
 	if err != nil {
 		return err
 	}
--- a/cmd/ooniprobe/internal/ooni/ooni.go
+++ b/cmd/ooniprobe/internal/ooni/ooni.go
@@ -179,9 +179,7 @@ func (p *Probe) Init(softwareName, softwareVersion, proxy string) error {
 
 	p.dbPath = utils.DBDir(p.home, "main")
 	log.Debugf("Connecting to database sqlite3://%s", p.dbPath)
-	db, err := database.NewDatabase(&database.DatabaseConfig{
-		DatabasePath: p.dbPath,
-	})
+	db, err := database.New(p.dbPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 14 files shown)


cmd/ooniprobe/internal/ooni/ooni.go
```diff
@@ -180,7 +179,9 @@ func (p *Probe) Init(softwareName, softwareVersion, proxy string) error {
 
 	p.dbPath = utils.DBDir(p.home, "main")
 	log.Debugf("Connecting to database sqlite3://%s", p.dbPath)
-	db, err := database.Connect(p.dbPath)
+	db, err := database.NewDatabase(&database.DatabaseConfig{
+		DatabasePath: p.dbPath,
+	})
```
bassosimone, line 184: I understand that your plan is that of adding to `DatabaseConfig` in the future a boolean field and later choose it to select which implementation to instantiate (the `ooniprobe` one or a `null` one). However, as things stand, I do not think we need to select which database to instantiate at runtime, rather we can decide that already at compile time: the ordinary database for `ooniprobe` and the null one for `miniooni`. As such, I do feel like a better pattern here would be the following:

```
func New(dbpath string) (*database.Database, error)
```

I am omitting `Database` here from `NewDatabase` because `database.New` is already readable while `database.NewDatabase` stutters a bit, as it repeats the word database.

internal/database/actions.go
```diff
@@ -17,10 +17,41 @@ import (
 	"github.com/upper/db/v4"
 )
 
+var (
+	errInvalidDatabasePath = "database: passed an empty database path"
```
bassosimone, line 21: Do we need to introduce this new error? (If the answer is yes, then it's more idiomatic to do

```Go
var errInvalidDatabasePath = errors.New(/* string */)
```

than what you are doing above.

internal/database/actions.go
```diff
@@ -17,10 +17,41 @@ import (
...
+// NewDatabase returns a new database instance
+func NewDatabase(config *DatabaseConfig) (*Database, error) {
+	if config.DatabasePath == "" {
+		return &Database{}, errors.New(errInvalidDatabasePath)
+	}
+	db, err := Connect(config.DatabasePath)
+	return &Database{
+		sess:   db,
+		dbPath: config.DatabasePath,
+	}, err
```
bassosimone, line 38: I would rather like see the implementation "implement" the equivalent of Rust's `Result<T>` construct and _either_ return a nil database to the caller (and a non-nil error) _or_ return a valid database (and a nil error).

internal/database/actions.go
```diff
@@ -17,10 +17,41 @@ import (
...
+}
+
+type Database struct {
```
bassosimone, line 41: Please remember to document every public method or struct!

bassosimone (changes requested):
Thank you! This is a good start and I have provided suggestions to improve the diff.

🙌

DecFox:
Thank you for the feedback! I have made the required changes.

internal/database/actions.go
```diff
@@ -17,10 +17,45 @@ import (
...
 
+var (
+	errInvalidDatabasePath = errors.New("database: passed an empty database path")
+)
+
+// DatabaseConfig is the configuration for creating a new database instance
+type DatabaseConfig struct {
+	DatabasePath string
+}
+
```
bassosimone, line 28:
```suggestion
```

I think we're not using this struct anymore (?)

internal/database/actions.go
```diff
@@ -17,10 +17,45 @@ import (
...
+// NewDatabase returns a new database instance
```
bassosimone, line 29:
```suggestion
// New returns a new database instance
```

internal/database/actions_test.go
```diff
@@ -43,6 +44,31 @@ func (lp *locationInfo) ResolverIP() string {
...
 }
 
+func TestNewDatabase(t *testing.T) {
+	t.Run("with empty path", func(t *testing.T) {
+		dbpath := ""
+		db, err := New(dbpath)
+		if db != nil {
+			t.Fatal("unexpected database instance")
+		}
+		if err == nil || !errors.Is(err, errInvalidDatabasePath) {
```
bassosimone, line 54:
```suggestion
		if !errors.Is(err, errInvalidDatabasePath) {
```

The above check also covers the case where `err` is `nil`

internal/database/actions_test.go
```diff
@@ -56,7 +82,7 @@ func TestMeasurementWorkflow(t *testing.T) {
 	}
 	defer os.RemoveAll(tmpdir)
 
-	sess, err := Connect(tmpfile.Name())
+	database, err := New(tmpfile.Name())
```
bassosimone, line 85: Sorry for asking to pile up another change but, now that I re-read the API, I think the following:

* `database.New` is idiomatic (it reads well)

* the stdlib uses `sql.Open` (which is the semantics I would expect for files)

* the code previously used `Connect` (which is the name we have in python's sqlite driver)

* the sqlite3 codebase for Go uses `SqliteDriver.Open`

In light of all of this, I would say we should perhaps use `Open` instead of `New`.

internal/database/actions.go
```diff
@@ -17,10 +17,45 @@ import (
...
+	if dbpath == "" {
+		return nil, errInvalidDatabasePath
+	}
+	db, err := Connect(dbpath)
+	if err != nil {
+		return nil, err
+	}
+	return &Database{
+		sess:   db,
+		dbPath: dbpath,
```
bassosimone, line 40: do we need to store the `dbPath`?

internal/database/actions.go
```diff
@@ -17,10 +17,45 @@ import (
 	"github.com/upper/db/v4"
 )
 
+var (
+	errInvalidDatabasePath = errors.New("database: passed an empty database path")
+)
+
```
bassosimone, line 23: Since the original code did not have this check, I think we can omit it?

bassosimone (approved):
🐳

PR #983, opened 2022-11-17
refactor: pass experiment arguments using a struct

bassosimone:
Closes https://github.com/ooni/probe/issues/2358.

### DIFF
--- a/internal/engine/experiment/dnscheck/dnscheck.go
+++ b/internal/engine/experiment/dnscheck/dnscheck.go
@@ -120,10 +120,11 @@ var (
 )
 
 // Run implements model.ExperimentSession.Run
-func (m *Measurer) Run(
-	ctx context.Context, sess model.ExperimentSession,
-	measurement *model.Measurement, callbacks model.ExperimentCallbacks,
-) error {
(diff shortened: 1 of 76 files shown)


internal/engine/experiment/dnscheck/dnscheck.go
```diff
@@ -120,10 +120,11 @@ var (
 )
 
 // Run implements model.ExperimentSession.Run
-func (m *Measurer) Run(
-	ctx context.Context, sess model.ExperimentSession,
-	measurement *model.Measurement, callbacks model.ExperimentCallbacks,
-) error {
+func (m *Measurer) Run(ctx context.Context, args *model.ExperimentArgs) error {
+	_ = args.Callbacks
```
ainghazal, line 124: no need to assign here no?

bassosimone: No, I left it for clarity to highlight we're not using the callbacks. We may need to use them and the `_` will tell us so.

bassosimone: I mean: previously we were passing a `callbacks` argument and ignoring and I want to keep track of that.

bassosimone:
I am going to merge having received a private "lgtm" from @ainghazal

PR #988, opened 2022-11-23
feat: introduce database abstraction 

DecFox:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2352
- [ ] if you change code inside an experiment, make sure you bump its version number

## Description

This diff introduces a database abstraction in `internal/model`.


cmd/ooniprobe/internal/cli/list/list.go
```diff
@@ -19,7 +19,7 @@ func init() {
 			return err
 		}
 		if *resultID > 0 {
-			measurements, err := probeCLI.DB().ListMeasurements(*resultID)
+			measurements, err := probeCLI.ReadDB().ListMeasurements(*resultID)
```
bassosimone, line 22: ???

bassosimone, line 22: This change seems unnecessary to me. In general, I think the whole diff contains too many changes and we can get away with a much smaller changeset. I would recommend to do the following:

1. the `*database.database` type remains `*database.Database` as it was before;
2. all the code in `cmd/ooniprobe` continues to use a `*database.Database` as it did before.

As such, this specific line of code should continue to be `probeCLI.DB`.

If there are exceptions to these general suggestions, I'll let you know.

cmd/ooniprobe/internal/cli/reset/reset.go
```diff
@@ -20,7 +20,9 @@ func init() {
 		}
 		// We need to first the DB otherwise the DB will be rewritten on close when
 		// we delete the home directory.
-		err = ctx.DB().Close()
+		// Since the writable and readable database are same, we only close the writable
+		// database
+		err = ctx.WriteDB().Close()
```
bassosimone, line 25: Same as above.

cmd/ooniprobe/internal/cli/rm/rm.go
```diff
@@ -8,11 +8,11 @@ import (
 	"github.com/alecthomas/kingpin"
 	"github.com/apex/log"
 	"github.com/ooni/probe-cli/v3/cmd/ooniprobe/internal/cli/root"
-	"github.com/ooni/probe-cli/v3/internal/database"
+	"github.com/ooni/probe-cli/v3/internal/model"
 	"github.com/upper/db/v4"
 )
 
-func deleteAll(d *database.Database, skipInteractive bool) error {
+func deleteAll(d model.WritableDatabase, r model.ReadableDatabase, skipInteractive bool) error {
```
bassosimone, line 15: Same as above.

cmd/ooniprobe/internal/nettests/nettests.go
```diff
@@ -94,7 +93,7 @@ func (c *Controller) BuildAndSetInputIdxMap(testlist []model.OOAPIURLInfo) ([]st
 	urlIDMap := make(map[int64]int64)
 	for idx, url := range testlist {
 		log.Debugf("Going over URL %d", idx)
-		urlID, err := c.Probe.DB().CreateOrUpdateURL(
+		urlID, err := c.Probe.WriteDB().CreateOrUpdateURL(
```
bassosimone, line 96: Same as above with a minor clarification.

In the future, when we will want to unify the `nettests` package with code in the `internal` package, we will need to perform an additional modification, i.e., making `nettests` depend on a `WritableDatabase`. I do not think we need to worry about doing this change right now (it will be a bit messy to do).

However, for this diff we should replace `c.Probe.WriteDB` with `writableDB(c.Probe)` where:

```Go
function writableDB(probe ooni.ProbeCLI) model.WritableDatabase {
  return probe.DB()
}
```

This adapter will _ensure_ that we can write code in terms of `WritableDatabase`.

cmd/ooniprobe/internal/nettests/nettests.go
```diff
@@ -122,7 +121,7 @@ func (c *Controller) SetNettestIndex(i, n int) {
 // This function will continue to run in most cases but will
 // immediately halt if something's wrong with the file system.
 func (c *Controller) Run(builder model.ExperimentBuilder, inputs []string) error {
-	db := c.Probe.DB()
+	db := c.Probe.WriteDB()
```
bassosimone, line 124: Same as above

cmd/ooniprobe/internal/nettests/nettests_test.go
```diff
@@ -52,7 +52,7 @@ func TestRun(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	db := probe.DB()
+	db := probe.WriteDB()
```
bassosimone, line 55: Same as above.

cmd/ooniprobe/internal/nettests/run.go
```diff
@@ -71,7 +71,7 @@ func RunGroup(config RunGroupConfig) error {
 		log.WithError(err).Error("Failed to lookup the location of the probe")
 		return err
 	}
-	db := config.Probe.DB()
+	db := config.Probe.WriteDB()
```
bassosimone, line 74: Same as above.

cmd/ooniprobe/internal/ooni/ooni.go
```diff
@@ -32,7 +32,8 @@ var logger = log.WithFields(log.Fields{
 // ProbeCLI is the OONI Probe CLI context.
 type ProbeCLI interface {
 	Config() *config.Config
-	DB() *database.Database
+	WriteDB() model.WritableDatabase
+	ReadDB() model.ReadableDatabase
```
bassosimone, line 36: Same as above (i.e., code in `cmd/ooniprobe` can continue to use a concrete database)

cmd/ooniprobe/internal/ooni/ooni.go
```diff
@@ -52,7 +53,8 @@ type ProbeEngine interface {
 // Probe contains the ooniprobe CLI context.
 type Probe struct {
 	config  *config.Config
-	db      *database.Database
+	writeDB model.WritableDatabase
+	readDB  model.ReadableDatabase
```
bassosimone, line 57: Same as above

cmd/ooniprobe/internal/ooni/ooni.go
```diff
@@ -84,9 +86,14 @@ func (p *Probe) Config() *config.Config {
...
-func (p *Probe) DB() *database.Database {
-	return p.db
+// WriteDB returns the writable database we're using
+func (p *Probe) WriteDB() model.WritableDatabase {
+	return p.writeDB
+}
+
+// ReadDB returns the readable database we're using
+func (p *Probe) ReadDB() model.ReadableDatabase {
+	return p.readDB
```
bassosimone, line 96: Same as above

cmd/ooniprobe/internal/ooni/ooni.go
```diff
@@ -183,7 +190,8 @@ func (p *Probe) Init(softwareName, softwareVersion, proxy string) error {
 	if err != nil {
 		return err
 	}
-	p.db = db
+	p.writeDB = db
+	p.readDB = db
```
bassosimone, line 194: Same as above

internal/database/actions.go
```diff
@@ -12,203 +12,123 @@ import (
...
 	}
-	return &Database{
+	return &database{
 		sess: db,
 	}, nil
 }
 
 // Database is a database instance to store measurements
-type Database struct {
+type database struct {
```
bassosimone, line 32: As mentioned above, this type should continue to exist as a public type.

internal/database/actions.go
```diff
@@ -12,203 +12,123 @@ import (
...
-			log.Errorf("failed to fetch the measurement %s %s", reportID, measurement.URL.URL.String)
-			return nil, err
-		}
-		defer resp.Body.Close()
-		if err := json.NewDecoder(resp.Body).Decode(&msmtJSON); err != nil {
-			log.Error("failed to unmarshal the measurement_json")
-			return nil, err
-		}
-		return msmtJSON, nil
-	}
```
bassosimone, line 104: I understand that here your intent was to group related functions together. It makes sense to do that but it does not make sense to do that _now_. By reordering the functions you are making what would be a very obvious and short diff much more complex to read and validate. Please, restore the previous order of the function such that the diff is minimal and obvious.

internal/database/models.go
```diff
@@ -1,160 +0,0 @@
...
-
-// PerformanceTestKeys is the result summary for a performance test
-type PerformanceTestKeys struct {
-	Upload   float64 `json:"upload"`
-	Download float64 `json:"download"`
-	Ping     float64 `json:"ping"`
-	Bitrate  float64 `json:"median_bitrate"`
-}
-
-// Finished marks the result as done and sets the runtime
```
bassosimone, line 100: Please, move the refactored functions from the other file back into this file. By doing that, it will become super obvious from the diff that we're just applying small changes and the diff will basically vouch for its own correctness in a much more easy and obvious way. I understand why you wanted to delete this file and the change in itself makes sense. However, by doing this change now, you're making a review for correctness of this diff much more complicated for a human to perform.

internal/model/database.go
```diff
@@ -0,0 +1,170 @@
...
+	ListResults() ([]DatabaseResultNetwork, []DatabaseResultNetwork, error)
+
+	// ListMeasurements given a result ID
+	ListMeasurements(resultID int64) ([]DatabaseMeasurementURLNetwork, error)
+
+	// GetMeasurementJSON returns a map[string]interface{} given a database and a measurementID
+	GetMeasurementJSON(msmtID int64) (map[string]interface{}, error)
+
+	// Close closes the database session
+	Close() error
```
bassosimone, line 191: I don't think `Close` should be part of the `ReadableDatabase` interface.

internal/model/database.go
```diff
@@ -0,0 +1,170 @@
...
+	Failed(msmt *DatabaseMeasurement, failure string) error
+
+	// Close closes the database session
+	Close() error
+}
+
+// ReadableDatabase only supports reading data.
+type ReadableDatabase interface {
+	// Session returns the database session
+	Session() db.Session
```
bassosimone, line 163: IIUC, `Session` allows to write a database, so it should not be part of this interface.

internal/model/mocks/database.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"database/sql"
+
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/upper/db/v4"
+)
+
+// Database allows mocking a database
+type Database struct {
```
bassosimone, line 11: ???

cmd/ooniprobe/internal/cli/list/list.go
```diff
@@ -62,7 +62,7 @@ func init() {
 			}
 			output.MeasurementSummary(msmtSummary)
 		} else {
-			doneResults, incompleteResults, err := probeCLI.DB().ListResults()
+			doneResults, incompleteResults, err := probeCLI.ReadDB().ListResults()
```
bassosimone, line 65: Same as above.

cmd/ooniprobe/internal/cli/show/show.go
```diff
@@ -16,7 +16,7 @@ func init() {
 			log.WithError(err).Error("failed to initialize root context")
 			return err
 		}
-		msmt, err := ctx.DB().GetMeasurementJSON(*msmtID)
+		msmt, err := ctx.ReadDB().GetMeasurementJSON(*msmtID)
```
bassosimone, line 19: Same  as above.

cmd/ooniprobe/internal/oonitest/oonitest.go
```diff
@@ -28,7 +28,7 @@ func (fo *FakeOutput) SectionTitle(s string) {
 // FakeProbeCLI fakes ooni.ProbeCLI
 type FakeProbeCLI struct {
 	FakeConfig         *config.Config
-	FakeDB             *database.Database
+	FakeDB             *mocks.Database
```
bassosimone, line 31: I don't think we need to stop using `*database.Database` at this time.

internal/model/database.go
```diff
@@ -0,0 +1,281 @@
...
+	"database/sql"
+	"time"
+
+	"github.com/upper/db/v4"
+)
+
+// WritabeDatabase supports writing and updating data.
+type WritableDatabase interface {
+	// Session returns the database session
+	Session() db.Session
```
bassosimone, line 17: Is is possible to omit `Session` from this interface? I would say that perhaps it is possible if the functions that we define below can internally use a `db.Session`. If you think about it, it's difficult otherwise to define a `NullDatabase` that does nothing: what would be return for `db.Session`? We would need to also return a mocked implementation of `db.Session` that does nothing. Conversely, it would be much easier if we could just ignore this function as an implementation detail and omit it from the `WritableDatabase` signature.

One of the reasons why I am advocating for _keeping_ a public `*database.Database` is precisely that code that needs the full `Database` can use the pointer to the correct type, while code that only needs some specific interfaces can use the correct interface. By switching such code to use an interface, we would make it independent of the specific `Database` implementation and thus enable moving it inside `internal`.

internal/model/database.go
```diff
@@ -0,0 +1,281 @@
...
+	// AddTestKeys writes the summary to the measurement
+	//
+	// Arguments:
+	//
+	// - msmt is the database measurement to update
+	//
+	// - tk is the testkeys
+	//
+	// Returns a non-nil error if measurement update failed
+	AddTestKeys(msmt *DatabaseMeasurement, tk interface{}) error
```
bassosimone, line 114:
```suggestion
	AddTestKeys(msmt *DatabaseMeasurement, tk any) error
```

Since go1.18, we should use `any` rather than `interface{}`.

internal/model/database.go
```diff
@@ -0,0 +1,281 @@
...
+	//
+	// - msmt is the database measurement to update
+	//
+	// - failure is the error string to use
+	//
+	// Returns a non-nil error if the measurement update failed
+	Failed(msmt *DatabaseMeasurement, failure string) error
+
+	// Close closes the database session
+	Close() error
```
bassosimone, line 157: The same argument used above for `Session` applies to `Close`. Users of a database don't need to care about closing it. I think it's the piece of code that OWNS the database (i.e., the `ooni.Probe` struct) that should eventually be responsible of closing the database. All the rest of the codebase does not need to worry about closing the database and actually it should not even be capable of doing so, because it's outside its remit.

bassosimone (changes requested):
Thank you! This is a good start but there's need to apply some changes before we could consider this diff ready for merging!

DecFox:
Thank you for the feedback! I have made the requested changes

internal/model/database.go
```diff
@@ -0,0 +1,267 @@
...
+//
+// Database
+//
+
+import (
+	"database/sql"
+	"time"
+)
+
+// WritabeDatabase supports writing and updating data.
```
bassosimone, line 12:
```suggestion
// WritableDatabase supports writing and updating data.
```

internal/model/database.go
```diff
@@ -0,0 +1,267 @@
+package model
+
+//
+// Database
```
bassosimone, line 4:
```suggestion
// Results database
```

bassosimone (approved):
✨ thank you!

PR #994, opened 2022-11-30
refactor(probeservices): use internal testing

bassosimone:
I was trying to (1) either move data types definitions inside probeservices to internal/model or (2) make them private to reduce the amount of exported definitions. However, I was severely limited in doing that by the fact that probeservices uses external testing.

To overcome this issue, I'm switching to internal testing first and then I will resume with the two points above, which are in line with my goal for today of moving https://github.com/ooni/probe/issues/2372 forward.

### DIFF
--- a/internal/engine/mockable/mockable.go
+++ b/internal/engine/mockable/mockable.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"net/url"
 
-	"github.com/ooni/probe-cli/v3/internal/engine/probeservices"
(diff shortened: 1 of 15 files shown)


internal/engine/mockable/mockable.go
```diff
@@ -150,4 +149,3 @@ func (sess *Session) UserAgent() string {
 }
 
 var _ model.ExperimentSession = &Session{}
-var _ probeservices.Session = &Session{}
```
bassosimone, line 153: I had to zap this to avoid an import cycle inside tests.

PR #996, opened 2022-11-30
probeservices: sync up with v3.14.0 ooapi package

bassosimone:
This diff ensures that the definitions we're using are in sync with the model used by [v3.14.0's ooapi package](
https://github.com/ooni/probe-cli/tree/v3.14.0/internal/ooapi/apimodel), which was an experiment at autogenerating APIs and cross checking with ooni/api's swagger JSON file.

The only difference that remains pertains the name of the query string parameter for /api/v1/test-list/urls, which must be plural (as it is in the probeservices package) rather than singular (as it is in the ooapi/apimodel package).

So, after merging this diff, we can say we have done part of the work specified by https://github.com/ooni/probe/issues/2372. However, we want to also check with the _current_ swagger.

(Because we already had a model that was consistent with the swagger _at the time we wrote ooapi_, it seems reasonable for us to start syncing up with that model, which is easy, and then focus on importing the swagger and comparing with it.)

### DIFF
--- a/internal/engine/probeservices/statefile.go
+++ b/internal/engine/probeservices/statefile.go
@@ -36,7 +36,7 @@ func (s State) Credentials() *model.OOAPILoginCredentials {
 	if s.Password == "" {
 		return nil
 	}
-	return &model.OOAPILoginCredentials{ClientID: s.ClientID, Password: s.Password}
+	return &model.OOAPILoginCredentials{Username: s.ClientID, Password: s.Password}
 }
 
 // StateFile is the orchestra state file. It is backed by
--- a/internal/engine/probeservices/urls.go
+++ b/internal/engine/probeservices/urls.go
@@ -21,6 +21,8 @@ func (c Client) FetchURLList(ctx context.Context, config model.OOAPIURLListConfi
 		query.Set("limit", fmt.Sprintf("%d", config.Limit))
 	}
 	if len(config.Categories) > 0 {
+		// Note: ooapi (the unused package in v3.14.0 that implemented automatic API
+		// generation) used `category_code` (singular) here, but that's wrong.
 		query.Set("category_codes", strings.Join(config.Categories, ","))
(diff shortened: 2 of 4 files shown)


internal/engine/probeservices/statefile.go
```diff
@@ -36,7 +36,7 @@ func (s State) Credentials() *model.OOAPILoginCredentials {
 	if s.Password == "" {
 		return nil
 	}
-	return &model.OOAPILoginCredentials{ClientID: s.ClientID, Password: s.Password}
+	return &model.OOAPILoginCredentials{Username: s.ClientID, Password: s.Password}
```
bassosimone, line 39: This diff only changes the Go visible identifier: the JSON name is unchanged.

internal/engine/probeservices/urls.go
```diff
@@ -21,6 +21,8 @@ func (c Client) FetchURLList(ctx context.Context, config model.OOAPIURLListConfi
 		query.Set("limit", fmt.Sprintf("%d", config.Limit))
 	}
 	if len(config.Categories) > 0 {
+		// Note: ooapi (the unused package in v3.14.0 that implemented automatic API
+		// generation) used `category_code` (singular) here, but that's wrong.
```
bassosimone, line 25:
```suggestion
		// generation) used `category_code` (singular) here, but that's wrong. The plural
		// name is the correct name as I've just verified -- 2022-11-30.
```

internal/model/ooapi.go
```diff
@@ -198,7 +214,7 @@ type OOAPICollectorUpdateResponse struct {
 
 // OOAPILoginCredentials contains the login credentials
 type OOAPILoginCredentials struct {
-	ClientID string `json:"username"`
+	Username string `json:"username"`
```
bassosimone, line 217: See the comment above.

PR #1018, opened 2023-01-06
feat(engine): make netxlite.utlsConn public

kelmenhorst:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2378https://github.com/ooni/probe/issues/2378

## Description

This PR exposes the `netxlite.utlsConn` to the experiments by making it public (--> `UTLSConn`) `echcheck` can use `netxlite.UTLSConn` now so we don't need to re-implement utlsConn there any longer.
Note that `NewConnUTLSWithHelloID` returns `*UTLSConn` now, instead of `TLSConn`, to make it more convenient for experiment developers.

### DIFF
--- a/internal/netxlite/utls.go
+++ b/internal/netxlite/utls.go
@@ -30,12 +30,12 @@ import (
 // Passing a nil `id` will make this function panic.
 func NewTLSHandshakerUTLS(logger model.DebugLogger, id *utls.ClientHelloID) model.TLSHandshaker {
 	return newTLSHandshaker(&tlsHandshakerConfigurable{
-		NewConn: newConnUTLS(id),
+		NewConn: newUTLSConnFactory(id),
 	}, logger)
 }
 
-// utlsConn implements TLSConn and uses a utls UConn as its underlying connection
-type utlsConn struct {
+// UTLSConn implements TLSConn and uses a utls UConn as its underlying connection
+type UTLSConn struct {
 	// We include the real UConn
 	*utls.UConn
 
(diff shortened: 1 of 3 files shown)


internal/netxlite/utls.go
```diff
@@ -46,22 +46,22 @@ type utlsConn struct {
...
 
-// Ensures that a utlsConn implements the TLSConn interface.
-var _ TLSConn = &utlsConn{}
+// Ensures that a UTLSConn implements the TLSConn interface.
+var _ TLSConn = &UTLSConn{}
 
-// newConnUTLS returns a NewConn function for creating utlsConn instances.
-func newConnUTLS(clientHello *utls.ClientHelloID) func(conn net.Conn, config *tls.Config) (TLSConn, error) {
+// NewConnUTLS returns a NewConn function for creating UTLSConn instances.
+func NewConnUTLS(clientHello *utls.ClientHelloID) func(conn net.Conn, config *tls.Config) (TLSConn, error) {
```
bassosimone, line 53: The name of this function was ~fine as long as it was private. Now that it is public, I am wondering why its name seems to imply that it creates a new connection, while instead it _returns_ a function for creating a new connection. I suppose we can probably find a better name. Additionally, or instead of doing that, we could also consider keeping this function private. It seems to me an adapter to allow creating uTLS connections inside `netxlite/tls.go` and I think most users want instead to call `NewConnUTLSWithHelloID`. Regardless of what we choose to do, perhaps `NewConnUTLS` is the right name for `NewConnUTLSWithHelloID`?

bassosimone (commented):
Looks good to me! It seems we can also seize the opportunity for choosing which functions to expose and for slightly improving the naming of functions.

internal/netxlite/utls.go
kelmenhorst: > It seems to me an adapter to allow creating uTLS connections inside netxlite/tls.go and I think most users want instead to call NewConnUTLSWithHelloID.

I have the same understanding.

> perhaps NewConnUTLS is the right name for NewConnUTLSWithHelloID?

Makes sense. 
And I agree, the adapter should rather be private. I can't think of a better name for it though, but I'll think about it

kelmenhorst: Maybe 
`NewConnUTLSWithHelloID` --> `NewUTLSConn`,
`NewConnUTLS` --> `newConnWithUTLS`

bassosimone: How about:

`NewConnUTLSWithHelloID` --> `NewUTLSConn`,
`newConnUTLS` --> `newUTLSConnFactory`

Not exactly as before but at least it removes the ambiguity of two function only differing in their first letter capitalization and doing a different business?

kelmenhorst: Yes, I think that's good 👍

internal/engine/experiment/echcheck/utls.go
```diff
@@ -31,60 +34,16 @@ func newHandshakerWithExtensions(extensions []utls.TLSExtension) func(dl model.D
 
 func (t *tlsHandshakerWithExtensions) Handshake(ctx context.Context, conn net.Conn, tlsConfig *tls.Config) (
 	net.Conn, tls.ConnectionState, error) {
-	t.conn = newConnUTLSWithHelloID(conn, tlsConfig, t.id)
+	var err error
+	t.conn, err = netxlite.NewUTLSConn(conn, tlsConfig, t.id)
+	runtimex.Assert(err == nil, "unexpected error when creating UTLSConn")
```
bassosimone, line 39: ✔️

bassosimone (approved):
🐳

PR #1024, opened 2023-01-10
android: compile custom libtor.a without using go-libtor

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2365

## Description

This pull request changes the android build to compile and use a custom `libtor.a` rather than using [go-libtor](https://github.com/ooni/go-libtor). The reason for implementing this change is explained below in the "motivation" section. (We will eventually migrate all this information into an issue.)

### Motivation

As far as I know, go-libtor is ~working as intended. There are some (solvable) issues around updating it from a `darwin/arm64` system. When investigating these issues, I came across more issues that made me consider dropping this library entirely and just doing the good old `./configure && make && make install` cross compiling dance.

The first issue I identified is subtle. Build systems for Unix (being them using the GNU style `./configure && make && make install`) or the OpenSSL style (configuration written in Perl) were historically written with the assumption that the user configures, builds, and installs a product. Then, once such a product has been installed, the user proceeds to build a dependency of the original product, whose configuration step locates the dependency. However, the way in which `go-libtor` works is different. It just runs the `./configure` (or equivalent) step and then derives build rules by analyzing what the compiler would do. What bothers me about this strategy is that the `./configure` script of a dependency is not using the exact dependency package that was just configured and built, but may end up using system-installed packages. As far as I can tell, this procedure is not causing any issue _right now_ but I don't like this fact that, when using go-libtor, the burden of the proof that the build is sound _is on me_. In short, I think the current implementation of go-libtor is possibly subject to _subtle_ issues caused by optimistically peeking into a Unix-style build system (which often times is a mess of not super robust bash scripts trying to configure a package written in C). I do not like that enough to continue using this library in its current form (but I am grateful it helped us for such a long time). I think that, perhaps, it would be more robust for go-libtor to always run the real build and declare itself as the compiler to properly intercept all command line arguments and be sure the build happens against the real dependencies. However, I do not have energy to write such a patch for go-libtor at the time (it seems difficult), so let's explore running `./configure` and friends manually.

The second issue is related to compiler flags. Many Unix packages do not set hardening flags by default, even though `tor` is a nice exception. It is difficult to know whether hardening flags would work in general, so it makes sense for go-libtor or CGO to avoid setting these flags by default. That said, if we compile `tor` from sources using `./configure`, we get hardening flags. When using go-libtor we don't. So the resulting binary is lacking a bunch of security features it could have. To fix, we can either use `CGO_CFLAGS` with CGO or run `./configure` manually.

So, because at this stage I now have two arguments in favour of run `./configure` manually, I think it makes sense for me to explore how this change would look like and compare what we would get with what libtor provides us with. This comparison will help making an informed decision on whether to migrate to the `./configure` approach or improve `go-libtor`.

### Structure of this patch

We introduce the `CDEPS` directory. It contains build scripts for C code and patches to apply to the code. I went a bit overboard and patched each dependency to introduce checks such that we're sure we're using the right headers. We add a `#define` to core headers and we break the build when using such headers if the define is missing.

The `internal/libtor` package is a new Go package that handles linking the C code. Its `enabled.go` file is probably the most tricky part of this pull request to review, because it contains CGO code. We should probably make sure we can always build this package for Linux, such that we can write comprehensive testing of that file.

We change the `internal/tunnel` package to select this embedding strategy for Android and (when special build flags are present) for Linux (so that we can have testing). We didn't change iOS for now, but we will do this pretty soon.

### Testing miniooni with embedded tor on linux/amd64

On Ubuntu you'll need these packages installed `built-essential`, `autoconf`, `automake`, `libtool`. (It may be that more packages are missing, in which case we'll update this list &mdash; FWIW I tested this branch on Ubuntu 22.04.)

```bash
./CLI/go-build-linux-amd64-embed-tor ./internal/cmd/miniooni
```

Then you can run:

```bash
./miniooni --tunnel=tor -n example
```

If you want to test also snowflake, replace `--tunnel=tor` with `--tunnel=torsf`.


.eslintrc.json
```diff
@@ -1,13 +0,0 @@
-{
```
bassosimone, line 1: This file was unused

CDEPS/build-linux-amd64
```diff
@@ -0,0 +1,27 @@
...
+case "$(uname -m)" in
+x86_64)
+	GOARCH=amd64
+	export OPENSSL_COMPILER=linux-x86_64
+	;;
+*)
+	echo "FATAL: unsupported architecture" 1>&2
+	exit 1
+	;;
+esac
```
bassosimone, line 18: TODO: we need to set proper compiler flags

bassosimone: Done

internal/libtor/enabled.go
```diff
@@ -0,0 +1,213 @@
...
+	// Route messages from and to the control connection
+	go sendrecv(cc, filep)
+	go sendrecv(filep, cc)
+
+	// Tell user that startup is successful.
+	p.startErr <- nil // nonblocking chan
+
+	// Run tor until completion. Note that return codes are not
+	// currently documented and they're never zero (WTF?!).
+	_ = C.tor_run_main(config)
```
bassosimone, line 204: This is actually wrong: there is a case that I missed (🤦) where the code returns `0`. So we should consider any value different from `0` here to be an execution error for `tor`.

.github/workflows/libtorlinux.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: echo ::set-output name=version::$(cat GOVERSION)
+
+      - uses: magnetikonline/action-golang-cache@v2
+        with:
+          go-version: "${{ steps.goversion.outputs.version }}"
+          cache-key-suffix: "-libtorlinux-${{ steps.goversion.outputs.version }}"
+
+      - run: ./CDEPS/build-linux-amd64 zlib openssl libevent tor
+
+      - run: go test -count 1 -v -cover -tags ooni_libtor -race ./internal/libtor/...
```
bassosimone, line 29: We also need to test that we're building with the expected compiler flags.

internal/libtor/enabled.go
```diff
@@ -0,0 +1,277 @@
...
+// #include <limits.h>
+// #include <stdbool.h>
+// #include <stdlib.h>
+//
+// #include <tor_api.h>
+//
+// /* Note: we need to define inline helpers because we cannot index C arrays in Go. */
+//
+// static char **cstringArrayNew(size_t size) {
+//     char **argv = calloc(sizeof(char *), size);
```
kelmenhorst, line 30: IIUC, `size` is the length of the string array, correct?
In this case, I think it should be `char **argv = calloc(size, sizeof(char *));` https://en.cppreference.com/w/c/memory/calloc
Or not?

internal/libtor/enabled.go
```diff
@@ -0,0 +1,277 @@
...
+//         abort();
+//     }
+//     return argv;
+// }
+//
+// static void cstringArraySet(char **argv, size_t index, char *entry) {
+//     argv[index] = entry;
+// }
+//
+// static void cstringArrayFree(char **argv, size_t size) {
```
kelmenhorst, line 41: Did you think about adding some safety checks here? 
I do understand that the way you use these functions, argv and the entries will not be NULL.

internal/libtor/enabled.go
bassosimone: They're both `size_t`. The first one is `count` and the second one is `size`. The man page on macOS says: "allocates enough space for count objects that are size bytes of memory each". So, swapping the arguments as you suggest is definitely more consistent with the spirit of the lib call, good catch 🙌 💯 🔝. (In practical terms, the code was nonetheless working as intended because the two `size_t` values are multiplied (with overflow checks) to obtain the total amount of bytes to allocate.)

internal/libtor/enabled.go
bassosimone: `free(NULL)` is working as intended in virtually all systems nowadays. However, historically, that was a portability concern, but it seems this is not a concern anymore. (See https://github.com/maxmind/geoip-api-c/pull/100#pullrequestreview-143264986 for a discussion on this topic.)

kelmenhorst: Got it, thanks for clarifying!

.github/workflows/libtorlinux.yml
bassosimone: done for android

bassosimone:
We merged most of this branch in https://github.com/ooni/probe-cli/pull/1052. What remains is the part related to compiling for Android, which is still shell script based but very valuable.

bassosimone:
Superseded by https://github.com/ooni/probe-cli/pull/1061

PR #1039, opened 2023-01-17
feat: enable A/B testing using check-in API

bassosimone:
See https://github.com/ooni/probe/issues/2396

### DIFF
--- a/internal/model/ooapi.go
+++ b/internal/model/ooapi.go
@@ -85,8 +85,8 @@ type OOAPICheckInResult struct {
 
 // OOAPICheckInResultConfig contains configuration.
 type OOAPICheckInResultConfig struct {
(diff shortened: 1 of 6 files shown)


internal/model/ooapi.go
```diff
@@ -85,8 +85,8 @@ type OOAPICheckInResult struct {
 
 // OOAPICheckInResultConfig contains configuration.
 type OOAPICheckInResultConfig struct {
-	// Feature contains feature flags.
-	Feature map[string]bool `json:"feature"`
+	// Features contains feature flags.
+	Features map[string]bool `json:"features"`
```
bassosimone, line 89: This was a typo that we didn't notice because we have not completed https://github.com/ooni/probe/issues/2372 yet.

PR #1051, opened 2023-01-25
feat: rewrite generic build rules in Go

bassosimone:
This diff rewrites the generic build rules in Go.

See https://github.com/ooni/probe/issues/2401.


bassosimone:
The failure in nextlite windows coverage is because of a flaky test.

PR #1068, opened 2023-02-01
cleanup(sessionresolver): don't depend on netx

bassosimone:
This diff closes https://github.com/ooni/probe/issues/2121 because it removes the last unnecessary netx usage. All the other packages that currently use netx are network experiments. We will eventually convert them to not use netx as part of other tracked issues.

This diff closes https://github.com/ooni/probe/issues/2135 because now the sessionresolver does not depend on netx anymore.

We refactor the way in which we conditionally configure byte counting for HTTP such that we use a free function rather than a method (we can use methods with nil receiver in Go, but the free function seems to be a better choice in this case).

We introduce and use a new bytecounter specifically for the system resolver. This byte counter is very imprecise but seems still better than using the system resolver doesn't use any network I/O.

We stop printing the sessionresolver stats when we close a session, since this component has been in production for years now.

We improve the model.UnderlyingNetwork model to add support for changing the root cert pool, which helps with writing some integration tests.

We modify the protocol to use and modify the netxlite tproxy (a singleton instance of UnderlyingNetwork) to make it goroutine safe.

We introduce a new file inside sessionresolver, factory.go, which creates and properly wraps the resolvers. This code replaces code for which we previously used netx, and is the core change introduced here.

While there, we refactor how we log in the session resolver to use the operation logger as we do in some experiments.

We write some additional tests that take advantage of the new netxlite tproxy mocking functionality to ensure the sessionresolver continues to work in two extreme use cases: no resolver is available and just the system resolver is available. I introduced these new tests because I originally broke the system resolver when I introduced factory.go and I felt like it was useful to have more robustness here.

### DIFF
--- a/internal/sessionresolver/factory.go
+++ b/internal/sessionresolver/factory.go
@@ -24,7 +24,7 @@ var errUnsupportedResolverScheme = errors.New("unsupported resolver scheme")
 //
 // - logger is the MANDATORY logger;
 //
-// - URL is the MANDATORY HTTP/3 URL to use;
+// - URL is the MANDATORY URL to use (a DoH URL or system:///);
 //
 // - http3Enabled indicates whether to use HTTP/3;
 //
--- a/internal/sessionresolver/factory.go
+++ b/internal/sessionresolver/factory.go
@@ -1,0 +1,100 @@
+package sessionresolver
+
+import (
+	"errors"
+	"net/url"
+
+	"github.com/ooni/probe-cli/v3/internal/bytecounter"
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/netxlite"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
+// errCannotUseHTTP3WithAProxyURL means we cannot construct a new
+// child resolver using HTTP/3 with a proxy URL.
+var errCannotUseHTTP3WithAProxyURL = errors.New("cannot use HTTP/3 with a proxy URL")
+
+// errUnsupportedResolverScheme means we don't support the
+// given resolver scheme. We only support https, http and system.
+var errUnsupportedResolverScheme = errors.New("unsupported resolver scheme")
+
+// newChildResolver constructs a new child resolver.
+//
+// Arguments:
+//
+// - logger is the MANDATORY logger;
+//
+// - URL is the MANDATORY HTTP/3 URL to use;
+//
+// - http3Enabled indicates whether to use HTTP/3;
+//
+// - counter is the OPTIONAL byte counter;
+//
+// - proxyURL is the OPTIONAL proxy URL.
+//
+// Using a proxy URL is incompatible with using HTTP/3 and this
+// factory will return an error if that happens.
+//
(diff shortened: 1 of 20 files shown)


internal/sessionresolver/resolvermaker.go
```diff
@@ -1,15 +1,14 @@
 package sessionresolver
 
 //
-// Code for creating a new child resolver
+// High-level for creating a new child resolver
```
bassosimone, line 4:
```suggestion
// High-level code for creating a new child resolver
```

internal/sessionresolver/factory.go
```diff
@@ -0,0 +1,100 @@
...
+// given resolver scheme. We only support https, http and system.
+var errUnsupportedResolverScheme = errors.New("unsupported resolver scheme")
+
+// newChildResolver constructs a new child resolver.
+//
+// Arguments:
+//
+// - logger is the MANDATORY logger;
+//
+// - URL is the MANDATORY HTTP/3 URL to use;
```
bassosimone, line 27:
```suggestion
// - URL is the MANDATORY URL to use (a DoH URL or system:///);
```

internal/sessionresolver/factory.go
```diff
@@ -0,0 +1,100 @@
...
+	if http3Enabled && proxyURL != nil {
+		return nil, errCannotUseHTTP3WithAProxyURL
+	}
+	parsed, err := url.Parse(URL)
+	if err != nil {
+		return nil, err
+	}
+	var reso model.Resolver
+	switch parsed.Scheme {
+	case "http", "https":
```
bassosimone, line 57:
```suggestion
	case "http", "https": // http is here for testing
```

internal/sessionresolver/factory.go
```diff
@@ -0,0 +1,100 @@
...
+		dialer := netxlite.MaybeWrapWithProxyDialer(
+			netxlite.NewDialerWithStdlibResolver(logger),
+			proxyURL, // handles correctly the case where proxyURL is nil
+		)
+		thx := netxlite.NewTLSHandshakerStdlib(logger)
+		tlsDialer := netxlite.NewTLSDialer(dialer, thx)
+		txp = netxlite.NewHTTPTransport(logger, dialer, tlsDialer)
+	case true:
+		// TODO(bassosimone): to test this arm we need to further extend
+		// netxlite to override the default list of certificates
```
bassosimone, line 92:
```suggestion
```

internal/sessionresolver/factory_test.go
```diff
@@ -0,0 +1,354 @@
...
+			addrs, err := reso.LookupHost(context.Background(), "dns.google")
+			if err == nil || err.Error() != netxlite.FailureConnectionReset {
+				t.Fatal("unexpected error", err)
+			}
+			if len(addrs) != 0 {
+				t.Fatal("expected zero length addrs here")
+			}
+		})
+
+		t.Run("what we get is an action DNS-over-HTTPS resolver", func(t *testing.T) {
```
bassosimone, line 145:
```suggestion
		t.Run("what we get is a DNS-over-HTTPS resolver", func(t *testing.T) {
```

PR #1073, opened 2023-02-03
poc: investigate tor SIGABRT with go-libtor-like code

bassosimone:
See https://github.com/ooni/probe/issues/2405


bassosimone:
We're not merging this PR. It has only been useful to pin the commits discussed at https://github.com/ooni/probe/issues/2405#issuecomment-1416085962.

PR #1078, opened 2023-02-09
fix: avoid submitting when tor binary is missing

bassosimone:
This diff modifies tunnel, torsf, and vanilla such that:

1. we return a special error when tor is not found on PATH

2. we use such an error to avoid submitting

By doing that, we save some useless submissions.

Part of https://github.com/ooni/probe/issues/2406 because I wrote this diff while investigating it.


bassosimone:
Thanks @DecFox for finishing this pull request! 🙌 💯 🔝 🙏

bassosimone:
I removed a test that is broken when `tor` is installed. I am now going to merge this PR, thanks again @DecFox!

PR #1083, opened 2023-02-13
refactor(mlablocatev2): modernize before adding DASH support

bassosimone:
This commit modernizes the mlablocatev2 API. This mainly means
preferring using pointers for structures in most cases. It also
entails avoiding to group variables and constants declarations
because this leads to more readable documentation.

In addition, make sure we use the mocks package rather than
rolling out private mocks. While there, also remove the example
because we do not have examples in most packages and, while it
may be nice to always have examples, history so far has thaught
us that we don't have capacity to write enough examples.

Part of https://github.com/ooni/probe/issues/2398


internal/mlablocatev2/mlablocatev2_test.go
```diff
@@ -150,35 +160,43 @@ func TestEmptyResponse(t *testing.T) {
...
+	client := NewClient(http.DefaultClient, model.DiscardLogger, "miniooni/0.1.0-dev")
 	client.HTTPClient = &http.Client{
-		Transport: FakeTransport{
-			Resp: &http.Response{
-				StatusCode: 404,
-				Body:       FakeBody{Err: io.EOF},
+		Transport: &mocks.HTTPTransport{
+			MockRoundTrip: func(req *http.Request) (*http.Response, error) {
+				resp := &http.Response{
+					StatusCode: 400,
```
bassosimone, line 168:
```suggestion
					StatusCode: 404,
```

PR #1113, opened 2023-02-28
poc: run integration tests using gvisor

bassosimone:
This diff introduces a new package called `internal/netem`. This package allows you to create clients and servers using gvisor-based user-space network stacks. You can attach each stack to a NIC and you can link together NICs using links. Each link has specific delay characteristics allowing one to emulate faster or slower links (do not expect realistic emulation, all that matters is that we can turn a knob and get one order of magnitude worst performance). It's also possible to emulate a backbone that ties all the links together. It's also possible to attach DPI rules to the servers (which is not what one would do in the real world but allows us to keep much less DPI state). The `./internal/cmd/netemu` command showcases how to use the new code that I am introducing in this PoC.

## Checklist

## Description


ainghazal:
I know you're trying to keep everything in go-land, and I understand the need to have all tests able to run via github actions, _but_...

I suspect this also opens the door to be able to connect the simulated DPI engine with _real_ servers running in a vm. If I'm not wrong, we could leverage something like https://github.com/containers/gvisor-tap-vsock to plug servers into a more complex SDN - or even borrow some concepts from theirs.

bassosimone:
We created github.com/ooni/netem and internal/netemx, which supersede this pull request.

PR #1125, opened 2023-03-20
RiseupVPN tests improvements

cyBerta:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1928 and https://github.com/ooni/probe/issues/1842
- [x] if you changed anything related to how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: https://github.com/ooni/spec/pull/273

## Description

This MR  aims to improve the data quality of RiseupVPN tests by reduce the amount of false positives due to server side issues.
Major changes are:

* API failures are reported, but not necessarily a reason to stop the test early
* invalid CA's are indicated but not considered a failure and the tests will continue without TLS verification
* a blocked API will be indicated as well, but tried to be circumvented with a Tor+Snowflake tunnel.
* the API is only considered blocked if the main configuration end point is unreachable, RiseupVPN has multiple ways to get around other other API blockages or server side misbehavior (targets https://github.com/ooni/probe/issues/1842#issuecomment-952783818)
* overloaded gateways are not tested anymore
* only a subset of gateways are tested

### DIFF
--- a/internal/experiment/riseupvpn/riseupvpn.go
+++ b/internal/experiment/riseupvpn/riseupvpn.go
@@ -6,13 +6,11 @@ package riseupvpn
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"time"
 
 	"github.com/ooni/probe-cli/v3/internal/experiment/urlgetter"
 	"github.com/ooni/probe-cli/v3/internal/model"
 	"github.com/ooni/probe-cli/v3/internal/netxlite"
-	"github.com/ooni/probe-cli/v3/internal/tracex"
 )
 
 const (
@@ -24,19 +22,19 @@ const (
 	tcpConnect    = "tcpconnect://"
 )
 
-// EipService is the main JSON object of eip-service.json.
-type EipService struct {
+// EipServiceV3 is the main JSON object of eip-service.json.
+type EipServiceV3 struct {
 	Gateways []GatewayV3
 }
 
-// Capabilities is a list of transports a gateway supports
-type Capabilities struct {
+// CapabilitiesV3 is a list of transports a gateway supports
+type CapabilitiesV3 struct {
 	Transport []TransportV3
 }
 
 // GatewayV3 describes a gateway.
 type GatewayV3 struct {
-	Capabilities Capabilities
+	Capabilities CapabilitiesV3
 	Host         string
 	IPAddress    string `json:"ip_address"`
 	Location     string `json:"location"`
@@ -48,13 +46,6 @@ type TransportV3 struct {
 	Protocols []string
 	Ports     []string
 	Options   map[string]string
-}
-
-// GatewayConnection describes the connection to a riseupvpn gateway.
-type GatewayConnection struct {
-	IP            string `json:"ip"`
-	Port          int    `json:"port"`
-	TransportType string `json:"transport_type"`
 }
 
 // GatewayLoad describes the load of a single Gateway.
@@ -83,21 +74,15 @@ type Config struct {
 // TestKeys contains riseupvpn test keys.
 type TestKeys struct {
 	urlgetter.TestKeys
-	APIFailure      []string            `json:"api_failure"`
-	APIStatus       string              `json:"api_status"`
-	CACertStatus    bool                `json:"ca_cert_status"`
-	FailingGateways []GatewayConnection `json:"failing_gateways"`
-	TransportStatus map[string]string   `json:"transport_status"`
+	APIFailure   []string `json:"api_failure"`
+	CACertStatus bool     `json:"ca_cert_status"`
 }
 
 // NewTestKeys creates new riseupvpn TestKeys.
 func NewTestKeys() *TestKeys {
 	return &TestKeys{
-		APIFailure:      nil,
-		APIStatus:       "ok",
-		CACertStatus:    true,
-		FailingGateways: nil,
-		TransportStatus: nil,
+		APIFailure:   nil,
+		CACertStatus: true,
 	}
 }
 
@@ -109,11 +94,6 @@ func (tk *TestKeys) UpdateProviderAPITestKeys(v urlgetter.MultiOutput) {
 	tk.TCPConnect = append(tk.TCPConnect, v.TestKeys.TCPConnect...)
 	tk.TLSHandshakes = append(tk.TLSHandshakes, v.TestKeys.TLSHandshakes...)
 	if v.TestKeys.Failure != nil {
-		for _, request := range v.TestKeys.Requests {
-			if request.Request.URL == eipServiceURL && request.Failure != nil {
-				tk.APIStatus = "blocked"
-			}
-		}
 		tk.APIFailure = append(tk.APIFailure, *v.TestKeys.Failure)
 		return
 	}
@@ -125,42 +105,6 @@ func (tk *TestKeys) UpdateProviderAPITestKeys(v urlgetter.MultiOutput) {
 func (tk *TestKeys) AddGatewayConnectTestKeys(v urlgetter.MultiOutput, transportType string) {
 	tk.NetworkEvents = append(tk.NetworkEvents, v.TestKeys.NetworkEvents...)
 	tk.TCPConnect = append(tk.TCPConnect, v.TestKeys.TCPConnect...)
-	for _, tcpConnect := range v.TestKeys.TCPConnect {
-		if !tcpConnect.Status.Success {
-			gatewayConnection := newGatewayConnection(tcpConnect, transportType)
-			tk.FailingGateways = append(tk.FailingGateways, *gatewayConnection)
-		}
-	}
-}
-
-func (tk *TestKeys) updateTransportStatus(openvpnGatewayCount, obfs4GatewayCount int) {
-	failingOpenvpnGateways, failingObfs4Gateways := 0, 0
-	for _, gw := range tk.FailingGateways {
-		if gw.TransportType == "openvpn" {
-			failingOpenvpnGateways++
-		} else if gw.TransportType == "obfs4" {
-			failingObfs4Gateways++
-		}
-	}
-	if failingOpenvpnGateways < openvpnGatewayCount {
-		tk.TransportStatus["openvpn"] = "ok"
-	} else {
-		tk.TransportStatus["openvpn"] = "blocked"
-	}
-	if failingObfs4Gateways < obfs4GatewayCount {
-		tk.TransportStatus["obfs4"] = "ok"
-	} else {
-		tk.TransportStatus["obfs4"] = "blocked"
-	}
-}
-
-func newGatewayConnection(
-	tcpConnect tracex.TCPConnectEntry, transportType string) *GatewayConnection {
-	return &GatewayConnection{
-		IP:            tcpConnect.IP,
-		Port:          tcpConnect.Port,
-		TransportType: transportType,
-	}
 }
 
 // AddCACertFetchTestKeys adds generic urlgetter.Get() testKeys to riseupvpn specific test keys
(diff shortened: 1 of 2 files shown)


cyBerta:
Hm... the coverage CI action wasn't failing before, when I played with the code on my personal github repo and filed a MR to myself.

From the above failing report - RiseupVPN's coverage:  
`github.com/ooni/probe-cli/v3/internal/experiment/riseupvpn	0.684s	coverage: 99.4% of statements`

cyBerta:
Hi @bassosimone @hellais, is there anything I could help with to support the review process?

cyBerta:
@bassosimone I've rebased the MR to that latest master branch. I'm sure you know that: to test the MR, you'll need to build mini-ooni. proble-cli tests will fail without an redirection in go.mod or until the changes landed on master. We would love to see this MR getting in soon, it will help us a lot with improving measurements for LEAP VPN / RiseupVPN.

bassosimone:
@cyBerta I will take care of doing a review of this MR! Thank you for updating riseupvpn!

bassosimone:
Thank you for your patience @cyBerta! I am currently finishing the review, running tests, and generally think about this experiment and how we can move it forward! I estimate I'll publish the review next week. 🙏

bassosimone (commented):
Thank you for improving the riseupvpn experiment! I started a detailed code review and tested the old and the new implementations side by side for a few hours. I did not finish my review because I realized that there was a more fundamental issue that we needed to discuss.

While testing I noticed that fetching eip-service.json and the geo service JSON has been very unreliable across my testing. It happened to me several times that the connection timed out when trying to connect to api.black.riseup.net. Your diff calls these APIs at lines [249](https://github.com/cyBerta/probe-cli/blob/e3561a1d8a04a2ab5c5e503e399850d14d9d56ae/internal/experiment/riseupvpn/riseupvpn.go#L249) and [255](https://github.com/cyBerta/probe-cli/blob/e3561a1d8a04a2ab5c5e503e399850d14d9d56ae/internal/experiment/riseupvpn/riseupvpn.go#L255) respectively. You process the results by calling the UpdateProviderAPITestKeys method on line [264](https://github.com/cyBerta/probe-cli/blob/e3561a1d8a04a2ab5c5e503e399850d14d9d56ae/internal/experiment/riseupvpn/riseupvpn.go#L255). This method sets the APIStatus as blocked on line [114](https://github.com/cyBerta/probe-cli/blob/e3561a1d8a04a2ab5c5e503e399850d14d9d56ae/internal/experiment/riseupvpn/riseupvpn.go#L114) if and only if the URL was that of the eip-service.json. In turn, this causes the whole measurement to be flagged as anomaly on line [493](https://github.com/cyBerta/probe-cli/blob/e3561a1d8a04a2ab5c5e503e399850d14d9d56ae/internal/experiment/riseupvpn/riseupvpn.go#L493). If I was a user and I had run and submitted those measurements, I would have believed that riseupvpn was blocked, while the underlying cause was the flaky service serving eip-service.json. If this event occurred so frequently that I could notice it, I am afraid it could have also happened for several other users. And we previously disabled riseupvpn because of too frequent false positives.

By reading the diff, I see how you did great work to try to reduce the frequency of false positives, which incidentally complicated the implementation and added extra complexity that we would need to maintain. At this point, I am convinced that we’re fighting an uphill battle and I want to instead propose a different approach that should make both my life (as the probe-cli maintainer) and your life (as the experiment maintainer) significantly easier. What’s more, my proposal would help  expedite the review and reintegration of the riseupvpn experiment into ooniprobe.

Before explaining my proposal, I would like to explain what, in my opinion, the goal of this experiment should be, both now and in the next two years. We would like to collect data useful to understand whether (1) riseupvpn endpoints are accessible and (2) the riseupvpn app could bootstrap by fetching the data it needs to do so. We are much less concerned with providing immediate feedback to users than we are about collecting data that can be useful to you folks to plan and reason about the global VPN service you provide.

For this reason, here’s what I propose to do in the short term:

1. We drop all the top-level keys from the experiment and we only keep the results generated by urlgetter. At the same time, we simplify the code that generates the summary so that it always says that there was no anomaly. This change removes the possibility that individual users would be triggered by flaky behavior of api.black.riseup.net and shifts the focus on the experiment purely on collecting data that should only be analyzed at scale (like we do for dnscheck).

2. We change the logic of the experiment such that, if any of the APIs required to bootstrap fail, we stop and submit a smaller JSON to the backend. This change removes the extra complexity that was already there and the additional one you introduced in this commit. Arguably, this would also make the experiment less complete than the one you implemented in this pull request, but I also have a plan to address that. However, this plan is not for the short term, but luckily it intersects with my other work commitments. I expand on this plan in detail below.

3. We modify the code to avoid submitting the JSON returned by the geo service, given that it includes geolocation data that it would be better not to submit. Granted, there is a mechanism in place with ooniprobe that strips the user’s IP address from the measurement. But I can see a bunch of cases where different geolocation mechanisms (the one used by OONI and the one used by riseup) may not agree because the IP address is different for some reasons and I would not like for us to increase the likelihood of a leak.

With these three changes in place, the riseupvpn experiment would be smaller and easier to maintain. So, we can add it back to ooniprobe in the experimental suite. By running it again, we can collect more data at scale that is certainly useful to you folks. At the same time, we do not need to worry anymore about false positives because now we focus on collecting data.

In the medium term, I am going to propose the following:

1. We are developing a mechanism by which the OONI backend can serve to ooniprobe a DSL describing what a specific network experiment needs to do. We are planning on using this mechanism for selected experiments (surely dnscheck and the IM nettests). I am suggesting we apply this mechanism also to riseupvpn. You can see a complete prototype of how this would work implemented in [ooni/2023-05-richer-input#12](https://github.com/ooni/2023-05-richer-input/pull/12) and documented at [probe#2502](https://github.com/ooni/probe/issues/2502). 

2. When riseupvpn is rewritten to use the DSL, the backend will periodically invoke the relevant riseupvpn APIs to fetch gateways to test. The backend will then serve to ooniprobes a DSL that contains instructions to measure the services required to bootstrap and the gateways. Because the backend generates the DSL, we have now removed the functional coupling between the first part of the riseupvpn experiment (bootstrap) and the second part (gateways). These two parts can now run independently of each other with moderate parallelism. Yay for simplicity!

3. The DSL will include a feature-flag mechanism allowing the selection of the best operation to perform depending on the probe capabilities. Once @ainghazal has landed minivpn support in ooniprobe, we will then be able to ask the ooniprobes that can do that, to perform the openvpn handshake. More generally, the DSL will decouple the definition of an experiment to the probe version out there, thus requiring fewer emergency releases.

In the long term (with all the usual caveats borrowed from J.M. Keynes), I suggest that we evaluate again the decision to score measurements locally based on the collected data. By then, incidentally, I suspect we would have migrated to a model where the OONI collector provides the data analysis and scoring for a measurement, which would further decouple releasing a new version of ooniprobe and tweaking data analysis.

What do you think?

cyBerta:
Wow, Thanks for this extensive review! I appreciate a lot the time you took for it. I can definitely change the given MR according to your short term proposals, they sound reasonable to me. Also awesome that you apparently already started to work on the midterm proposals as I understand https://github.com/ooni/probe/issues/2502. 
I haven't spent time yet to dig into the DSL approach, but will do so soon, so that I can hopefully help to maintain the shiny new solution.

bassosimone:
Thank you, @cyBerta !

cyBerta:
ok, I think I've addressed your proposals. 


>   1.  We drop all the top-level keys from the experiment and we only keep the results generated by urlgetter.

I've left 2 TestKeys that helps within the test to use tls-no-verify in case fetching the CA cert fails or the CA is expired. Next to the slow response time Riseup's API server it was a common case for false positives. Also we're using these keys in a fetch-and-parsing pipeline for prometheus, so keeping these keys would help us to change less code  at other places as well.

> At the same time, we simplify the code that generates the summary so that it always says that there was no anomaly. 

I've removed the summary keys and the tests always return false for IsAnomaly

> This change removes the possibility that individual users would be triggered by flaky behavior of api.black.riseup.net and shifts the focus on the experiment purely on collecting data that should only be analyzed at scale (like we do for dnscheck).

yey, should be the case now

>    2.  We change the logic of the experiment such that, if any of the APIs required to bootstrap fail, we stop and submit a smaller JSON to the backend. This change removes the extra complexity that was already there and the additional one you introduced in this commit. Arguably, this would also make the experiment less complete than the one you implemented in this pull request, but I also have a plan to address that. However, this plan is not for the short term, but luckily it intersects with my other work commitments. I expand on this plan in detail below.

I've changed the test so that it fails and returns early. 

>    3. We modify the code to avoid submitting the JSON returned by the geo service, given that it includes geolocation data that it would be better not to submit. Granted, there is a mechanism in place with ooniprobe that strips the user’s IP address from the measurement. But I can see a bunch of cases where different geolocation mechanisms (the one used by OONI and the one used by riseup) may not agree because the IP address is different for some reasons and I would not like for us to increase the likelihood of a leak.

while I was at it I've removed all response bodies, since they aren't interesting for us and they unnecessarily leak data

bassosimone:
I mistakenly pushed some of my comments in this branch, but I reverted it now.

bassosimone:
@cyBerta I started merging code from this PR into the master branch. The related PRs are https://github.com/ooni/probe-cli/pull/1360, https://github.com/ooni/probe-cli/pull/1363, and https://github.com/ooni/probe-cli/pull/1364.

https://github.com/ooni/probe-cli/pull/1363#issue-1938109948 in particular explains my rationale for pulling changes and what I chose not to include for the 3.19 release. It also explains how I think we can use some form of richer input, such as the DSL described above, or perhaps something simpler (and easier to implement), to decouple testing services useful to bootstrap riseupvpn and getting the endpoints we should be testing.

I think this PR should probably remain open because it contains the work about finding out which subset of the gateways to test, which I think in the future should be an algorithm that runs in the backend to find out which gateways probes should test.

If you have no objections, I'd like to go ahead and merge your changes with master, so it's more clear what has not been added into master and the algorithms about parsing the geo service results stand out more clearly.

Thank you!

internal/experiment/riseupvpn/riseupvpn.go
```diff
@@ -230,46 +188,62 @@ func (m Measurer) Run(ctx context.Context, args *model.ExperimentArgs) error {
...
+	if testkeys.APIFailure != nil {
+		// scrub reponse bodys before early returning
+		var scrubbedRequests = []model.ArchivalHTTPRequestResult{}
+		for _, requestEntry := range testkeys.Requests {
+			requestEntry.Response.Body = model.ArchivalHTTPBody{Value: "[scrubbed]"}
+			scrubbedRequests = append(scrubbedRequests, requestEntry)
+		}
+		testkeys.Requests = scrubbedRequests
+		return nil
+	}
```
bassosimone, line 220: Regarding scrubbing, I should mention here that since https://github.com/ooni/probe-cli/commit/7a6fd910df7bbfd11f9f257cbd778f686d2994ca we should be unconditionally scrubbing HTTP headers and bodies by removing anything that looks like an IP address or endpoint.

cyBerta:
> @cyBerta I started merging code from this PR into the master branch. The related PRs are #1360, #1363, and #1364.
> 
> [#1363 (comment)](https://github.com/ooni/probe-cli/pull/1363#issue-1938109948) in particular explains my rationale for pulling changes and what I chose not to include for the 3.19 release. It also explains how I think we can use some form of richer input, such as the DSL described above, or perhaps something simpler (and easier to implement), to decouple testing services useful to bootstrap riseupvpn and getting the endpoints we should be testing.
> 
> I think this PR should probably remain open because it contains the work about finding out which subset of the gateways to test, which I think in the future should be an algorithm that runs in the backend to find out which gateways probes should test.
> 
> If you have no objections, I'd like to go ahead and merge your changes with master, so it's more clear what has not been added into master and the algorithms about parsing the geo service results stand out more clearly.
> 
> Thank you!

hi @bassosimone 
Thank you so much for your thoughts, effort and work you've put into this review and the subsequent PRs! I'm curious to see how this test suite evolves. please let me know how I can support the development in any way.

PR #1128, opened 2023-03-22
fix(dslx): Remove calls to NewDefaultCertPool

kelmenhorst:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2402 

## Description
We don't want to use `NewDefaultCertPool` anymore outside of netxlite (https://github.com/ooni/probe/issues/2444).


bassosimone (changes requested):
Rather than just removing the `RootCAs` fields, could you keep them as `nil` and add before the struct you're initializing the same disclaimer we added in other places:

```Go
	// See https://github.com/ooni/probe/issues/2413 to understand
	// why we're using nil to force netxlite to use the cached
	// default Mozilla cert pool.
```

bassosimone (approved):
🐳

PR #1135, opened 2023-05-07
signal: remove check for api.directory.signal.org

simoncpu:
The reference to the `api.directory.signal.org` domain was removed in signalapp/Signal-TLS-Proxy@461d2d1aeec601a4d62100949a227f0cef2f6154, so the check for Signal fails. This commit removes the reference to the domain in OONI Probe as well.


bassosimone (approved):
🐳

bassosimone:
Thank you, I will include this PR into the next release

PR #1137, opened 2023-05-20
Benchmark using cachegrind

FedericoCeratto:
Implement basic reproducible benchmarking

See https://pythonspeed.com/articles/consistent-benchmarking-in-ci/

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -1,6 +1,7 @@
 # Runs quality assurance checks
 name: "qa"
 on:
+  pull_request:
   push:
     branches:
       - "release/**"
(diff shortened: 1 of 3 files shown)


.github/workflows/qa.yml
```diff
@@ -1,6 +1,7 @@
 # Runs quality assurance checks
 name: "qa"
 on:
+  pull_request:
```
bassosimone, line 4: We don't want to run this workflow on every pull request because it is slow and it will hence slow down development. We usually only want to run QA tests when we are inside a release branch.

bassosimone (changes requested):
I don't think we should conflate the scripts we use for performing QA with cachegrind runs because that would most likely make the QA scripts (which are already slow and flaky) even slower and more annoying to run during releases.

Go contains [an extensive framework for benchmarking](https://go.dev/doc/diagnostics), but it is unclear to me whether there is an equivalent of using cachegrind. Could you check? (I'd rather really use native Go tools for benchmarking than external tools.)

What were you trying to benchmark? Perhaps, I can help by adding support to miniooni for generating performance traces that we can analyze with Go tools.

PR #1150, opened 2023-06-05
refactor:geolocate package to fundamentals

DecFox:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2491

## Description

This diff divides the `geolocate` package functionality to separate `iplookup` and `resolverlookup` packages. It also shifts the usability policy of the geolocate results to `internal/engine`


internal/engine/geolocate.go
```diff
@@ -74,8 +75,8 @@ type Config struct {
 	UserAgent string
 }
 
-// NewTask creates a new instance of Task from config.
-func NewTask(config Config) *Task {
+// NewIplookupTask creates a new instance of IpLookupTask from config.
+func NewGeolocateTask(config GeolocateConfig) *GeolocateTask {
```
bassosimone, line 79: Something to think about: we may want to pass a `*GeolocateConfig` here (which is what we try to do in many cases to have consistent coding style). However, in such a case, we should refactor the code below such that we always avoid mutating the `config` structure that was passed to us (for correctness). We may also want to consider checking what the calls of `NewGeolocateTask` do and, if they always initialize all the fields, then we can just avoid setting default values and make all the fields in the config struct mandatory.

bassosimone (changes requested):
Thank you! This is a good starting point to perform the refactoring I had in mind! I have provided feedback explaining what we're missing to land this PR.

PR #1159, opened 2023-06-12
fix(.github): avoid using deprecated ::set-output

bassosimone:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


bassosimone:
Merging because it is obvious at this point that the change is working as intended. We're using the correct Go version and we have removed the warnings emitted by GitHub while running.

PR #1162, opened 2023-06-24
feat(netxlite): add HTTP/HTTPS proxy dialer

Murphy-OrangeMud:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2552

## Description
 
Add support for HTTP/HTTPS proxy dialing in netxlite.

Even though we can already pass `--proxy` with HTTP/HTTPS proxies to `ooniprobe`, e.g.,.

```
./ooniprobe run all --proxy http://127.0.0.1:7890
```

The implementation of this functionality added as part of https://github.com/ooni/probe/issues/2531#issuecomment-1743087781 does not allow to have per-domain specific proxies or to perform proxy hopping, as explained in https://github.com/ooni/probe/issues/2552. By adding an HTTP/HTTPS proxy dialer to `netxlite`, instead, we would make these kind of advanced behavior possible.


bassosimone:
Thanks for contributing! I'll take a look at this diff when preparing v3.19.x!

bassosimone (changes requested):
I have implemented HTTP/HTTPS proxies by brutally reusing the standard library functionality as documented at https://github.com/ooni/probe/issues/2531#issuecomment-1743087781, because I was working on related code and that seemed very easy to do.

The solution I implemented has some drawbacks that suggest that it would still be useful to have a way of connecting to a proxy along the lines you sketched out here. In particular, I believe this functionality would allow us to specify proxies in a more fine grained way and it would also allow us to dynamically switch proxy. For this reason, I am still very interested in seeing this code landing in the master branch sometime in the future.

However, I think this pull request is probably trying to do too many things for the objectives that we have in OONI and I have provided some initial recommendations to simplify and streamline the code. Please, start by applying the changes I recommended, such that we can move a step forward and can gradually make this branch ready for landing.

Thank you for your work and effort! 🙏 🙂

PR #1176, opened 2023-07-06
feat(oonimkall): start sketching out oonirun v2 API

bassosimone:
## Checklist

## Description

### DIFF
--- a/pkg/oonimkall/xoonirun.go
+++ b/pkg/oonimkall/xoonirun.go
@@ -1,0 +1,193 @@
+package oonimkall
+
+//
+// eXperimental OONI Run code.
+//
+
+import (
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


bassosimone:
We discussed with @aanorbel and it seems it would be simpler if the API call just returned a `java.lang.String` to the caller rather than attempting to parse the JSON and return back an object that it's difficult to manage in Java.

bassosimone:
> We discussed with @aanorbel and it seems it would be simpler if the API call just returned a `java.lang.String` to the caller rather than attempting to parse the JSON and return back an object that it's difficult to manage in Java.

Done in https://github.com/ooni/probe-cli/pull/1176/commits/55936f35c5820081c5b807c6a37ca7293c6a648b

PR #1184, opened 2023-07-18
chore: use go1.20.6

bassosimone:
See https://github.com/ooni/probe/issues/2503


bassosimone:
We see the following coverage failure:

```
2023/07/18 13:57:46 failed to increase receive buffer size (wanted: 2048 kiB, got 0 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Receive-Buffer-Size for details.
2023/07/18 13:57:46 http: TLS handshake error from 10.0.0.17:61842: EOF
2023/07/18 13:57:51 http: TLS handshake error from 10.0.0.17:21879: use of closed network connection
2023/07/18 13:57:51 http: TLS handshake error from 10.0.0.17:21878: use of closed network connection
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x82e0a2]

goroutine 176 [running]:
net/http.(*conn).serve(0xc000632bd0, {0x105d3c0, 0xc000367530})
	/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:1847 +0x82
created by net/http.(*Server).Serve
	/opt/hostedtoolcache/go/1.20.6/x64/src/net/http/server.go:3089 +0x818
FAIL	github.com/ooni/probe-cli/v3/internal/experiment/tlsping	6.593s
```

We should understand and address it before merging to master.

bassosimone:
The crash happens on [line 1847 of server.go](https://github.com/golang/go/blob/go1.20.6/src/net/http/server.go#L1847), which reads:

Screenshot 2023-07-18 at 16 11 02

So, it seems we need to make sure there is a valid non-nil address. (I also ran the tests locally and it was working for me, so it may be to be the error proper was just transient&mdash;nonetheless, if I can fix it, that's much better!)

bassosimone:
I re-run coverage and it succeeded. I know sometimes gvisor could behave slightly differently than normal networking and so I am going to ignore this kind of failure right now under the assumption it was an unlikely one. I'll look more closely in case the above kind of failure starts happening more and more frequently.

PR #1185, opened 2023-07-18
Add netemx tests for webconnectivitylte

kelmenhorst:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2461

## Description
This diff implements netemx testing for webconnectivitylte by doing the following high-level things:
* Move most code in internal/cmd/oohelperd to internal/oohelperd such that this library can be used by our tests.
* Refactor netxlite w.r.t. the underlying network (this change enables us to run multiple stacks in the `QAEnv`)
   * stdlib dialers/resolvers/handshakers now optionally own an instance of `model.UnderlyingNetwork` that, if set, is used instead of `tproxySingleton()`
   * internal/netxlite/net.go now contains the high-level struct `Net` which wraps an `UnderlyingNetwork`. Its methods are constructors that mimic constructors in netxlite, e.g. `NewDialerWithResolver`. These constructor methods return dialers/resolvers/handshakers which use this `Net`'s `UnderlyingNetwork`.
* Add a success case and DPI case as integration tests for webconnectivitylte. Note that we can now run both the probe as well as the testhelper as separate stacks in the testing environment.


internal/netemx/qaenv.go
```diff
@@ -295,34 +326,40 @@ func (env *QAEnv) mustNewHTTPServers(config *qaEnvConfig) (closables []io.Closer
 				RightToLeftDelay: time.Millisecond,
 			},
 		))
+		env.serverStacks[addr] = stack
 
-		ipAddr := net.ParseIP(addr)
-		runtimex.Assert(ipAddr != nil, "invalid IP addr")
+		handler := factory.NewHandler(stack)
+		closables = env.serverListen(stack, handler, addr)
```
bassosimone, line 332: I think here we should append to the closables rather than assigning.

bassosimone:
I have split this PR in several smaller PRs and integrated the code in master.

PR #1193, opened 2023-08-18
fix(tutorial): avoid broken pipe caused by passing non-JSON to jq

imw:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2523

## Description

Following the tutorial code results in a broken pipe. This is a simple fixup that adds a call to tail so that jq is only given valid input.


bassosimone (approved):
🐳 🙏

PR #1208, opened 2023-08-24
feat(webconnectivitylte): use webconnectivityqa for testing

bassosimone:
This diff removes previous (and rather simplistic) netem-based testing for webconnectivitylte and replaces it with the test harness defined by the webconnectivityqa package. Reference issue: https://github.com/ooni/probe/issues/2461.


internal/experiment/webconnectivitylte/analysiscore.go
```diff
@@ -122,6 +122,8 @@ func (tk *TestKeys) analysisToplevel(logger model.Logger) {
 			tk.BlockingFlags, tk.Accessible, tk.Blocking,
 		)
 
+	// assigning "http-failure" for both TLS and HTTP blocking is legacy
```
bassosimone, line 125:
```suggestion
	// Assigning "http-failure" for both TLS and HTTP blocking is a legacy behavior
```

PR #1217, opened 2023-08-30
feat: add echcheck to the experimental suite

kelmenhorst:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1453 https://github.com/ooni/probe/issues/2547
- [x] if you changed anything related to how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: https://github.com/kelmenhorst/spec/tree/echcheck-spec 

## Description
This diff does two things:
1. Add `echcheck` to the experimental nettest suite. We do not provide input which causes the experiment to use the default URL `https://crypto.cloudflare.com/cdn-cgi/trace` (URL [proposed](https://github.com/ooni/probe/issues/1453#issuecomment-1699380640) by @eighthave). With this single input configuration we can collect the first experimental ECH measurements.
2. Add netem tests to `echcheck`, while keeping a "real-internet" test to connect to an ECH-enabled server.


eighthave:
I don't know this codebase, or Go for that matter, so I can't really follow this.  But the idea sounds good.  I would recommend including the data that https://crypto.cloudflare.com/cdn-cgi/trace returns, especially the results of the `sni=` field.  For example:

```properties
fl=125f76
h=crypto.cloudflare.com
ip=23.153.248.38
ts=1693413463.195
visit_scheme=https
uag=Mozilla/5.0 (X11; Linux x86_64; rv:102.0) Gecko/20100101 Firefox/102.0
colo=MCI
sliver=none
http=http/2
loc=T1
tls=TLSv1.3
sni=plaintext
warp=off
gateway=off
rbi=off
kex=X25519
```

You can see my Firefox does not have ECH enabled, since `sni=plaintext`

kelmenhorst:
@eighthave Yes I was thinking the same. However, at the moment, echcheck only handshakes. That said, I don't see a reason why we should not do HTTP GET here.

internal/experiment/echcheck/measure_test.go
```diff
@@ -4,92 +4,156 @@ import (
...
+	err := measurer.Run(context.Background(), args)
 	if err == nil {
 		t.Fatal("expected an error here")
 	}
+	if err.Error() != "input is not an URL" {
+		t.Fatal("unexpected error type")
+	}
+}
+
+func TestMeasurerSuccess(t *testing.T) {
```
bassosimone, line 92:
```suggestion
func TestMeasurerExperimentFailsBecauseNetemDoesNotSupportECH(t *testing.T) {
```

internal/experiment/echcheck/measure_test.go
```diff
@@ -4,92 +4,156 @@ import (
...
+			t.Fatal("unexpected error GetSummaryKeys", err)
+		}
+		if summary.(SummaryKeys).IsAnomaly != false {
+			t.Fatal("expected false")
+		}
+		tk := msrmnt.TestKeys.(TestKeys)
+		if tk.Control.Failure == nil || tk.Target.Failure == nil {
+			// we expect errors here because we cannot do ECH with netem yet
+			t.Fatal("expected an error here")
+		}
```
bassosimone, line 123:
```suggestion
		}
		t.Skip("netem does not support ECH so we cannot declare this test as successful")
```

cmd/ooniprobe/internal/nettests/echcheck.go
```diff
@@ -0,0 +1,14 @@
...
+// ECHCheck nettest implementation.
+type ECHCheck struct{}
+
+// Run starts the nettest.
+func (n ECHCheck) Run(ctl *Controller) error {
+	builder, err := ctl.Session.NewExperimentBuilder("echcheck")
+	if err != nil {
+		return err
+	}
+	// providing empty input prompts the experiment to use the default URL https://example.org
```
bassosimone, line 12:
```suggestion
	// providing empty input causes the experiment to use the default URL https://example.org
```

internal/experiment/echcheck/measure_test.go
```diff
@@ -4,94 +4,126 @@ import (
...
+		cancel() // immediately cancel the context
+
+		// create measurer
+		measurer := NewExperimentMeasurer(Config{})
+		args := &model.ExperimentArgs{
+			Callbacks:   model.NewPrinterCallbacks(model.DiscardLogger),
+			Measurement: &model.Measurement{},
+			Session:     &mocks.Session{MockLogger: func() model.Logger { return model.DiscardLogger }},
+		}
+		// run measurement
```
bassosimone, line 56:
```suggestion

		// run measurement
```

bassosimone:
@kelmenhorst the reason why we should not fetch the webpage here was that the result included the IP address of the probe and we would like to avoid that. We used to have a mitigation where we scrubbed the IP address known to us, i.e., the one we discovered, from the measurement result. Since v3.19.0-alpha, there's now a much more robust mitigation where we scrub any IP address or endpoint, which means now it is much safer to fetch this page.

bassosimone (approved):
🐳

PR #1221, opened 2023-09-04
refactor(netemx): move towards unified factory/server

bassosimone:
This diff starts refactoring netemx such that any kind of service we want to create for a given IP address uses the same interface, which is a generic factory to create a generic server working depending on a netstack.

The overall objective here is to enable the same IP address to serve DNS, HTTP/HTTPS/HTTP3, and possibly arbitrary other services. Currently, we can only choose one of these services per IP address. This limitation, for example, makes it impossible to have 8.8.8.8 handle both Do53 and DoH.

I will move towards this objective incrementally. This diff just refactors the way in which we manage the echo server to follow a model where there is a factory for creating a server, which is what HTTP code is using.

Once this diff lands, I will refactor HTTP and DNS to follow this model. And, after that, I will try to allow multiple factories for each IP addr.

Reference issue: https://github.com/ooni/probe/issues/1803

## Checklist

- [x] reference issue for this pull request: see above

### DIFF
--- a/internal/netemx/tcpecho.go
+++ b/internal/netemx/tcpecho.go
@@ -10,80 +10,94 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/runtimex"
 )
 
-// TCPEchoNetStack is a [QAEnvNetStackHandler] implementing a TCP echo service.
-func TCPEchoNetStack(logger model.Logger, ports ...uint16) QAEnvNetStackHandler {
-	return &tcpEchoNetStack{
-		closers: []io.Closer{},
-		logger:  logger,
-		mu:      sync.Mutex{},
-		ports:   ports,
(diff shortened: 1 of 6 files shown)


internal/netemx/tcpecho.go
```diff
@@ -10,80 +10,94 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/runtimex"
 )
 
-// TCPEchoNetStack is a [QAEnvNetStackHandler] implementing a TCP echo service.
-func TCPEchoNetStack(logger model.Logger, ports ...uint16) QAEnvNetStackHandler {
-	return &tcpEchoNetStack{
+// NewTCPEchoServerFactory is a [QAEnvNetStackHandler] implementing a TCP echo service.
```
bassosimone, line 13:
```suggestion
// NewTCPEchoServerFactory is a [NetStackServerFactory] for the TCP echo service.
```

PR #1234, opened 2023-09-05
feat: start merging filtering into testingx

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1803

## Description

To continue rewriting Jafar based tests to use netemx, the next step is to adapt the proxies used by Jafar to be usable from within netemx to implement equivalent test cases.

In turn, netemx is using testingx for general-purpose test servers that could also be interesting for other packages.

We also have the netxlite/filtering package, which is ~fine but has a string-based API, where an interface-based API would be more proper and easier to compose. (We historically use a string-based API there because we previously attempted to replace Jafar with code in userspace usinf netxlite/filtering.)

The first step in this quest is therefore to rewrite the DNS code inside netxlite/filtering and move it to testingx.

While there rename netemx.UDPResolverFactory to DNSOverUDPServerFactory for consistency.


internal/netxlite/dnsoverudp_test.go
```diff
@@ -311,7 +307,7 @@ func TestDNSOverUDPTransport(t *testing.T) {
 			if _, err := rch.DecodeLookupHost(); err != nil {
 				t.Fatal(err)
 			}
-			// Now wait for the delayed response to arrive. We don't care much
+			// Now wait for the delayed responses to arrive. We don't care much
```
bassosimone, line 310:
```suggestion
			// Now wait for the delayed response to arrive. We don't care much
```

internal/testingx/dnssimulategfw.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/ooni/netem"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
+// DNSNumBogusResponses is a type indicating the number of bogus responses
+// the [DNSSimulateGWFListener] should emit for each round trip.
+type DNSNumBogusResponses int
+
+// DNSSimulateGWFListener is a DNS-over-UDP listener that simulates the GFW behavior by
+// responding with three answers, where the first two answers are invalid for the domain
```
bassosimone, line 17:
```suggestion
// responding with N+1 answers, where the first N answers are invalid for the domain
```

PR #1235, opened 2023-09-05
refactor: merge filtering's HTTP code with testingx

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/1803

## Description

This diff merges internal/filtering's code with testingx. While doing this, we refactor the code to (1) be a drop-in replacement for the net/http/httptest package and (2) use interfaces rather than string values to select the proper behavior, which makes the code more composable than it was before.

The overall objective of this effort is to consolidate filtering into testingx such that we can later merge the proxy capabilities implemented by Jafar into the testingx package, which is a stepping stone for ditching Jafar.


internal/testingx/httptestx.go
```diff
@@ -0,0 +1,160 @@
...
+// Bug: this handler does not WAI when using [github.com/ooni/netem]. The reason why this happens
+// is that gvisor.io supports SO_LINGER but there's no *gonet.TCPConn.SetLinger.
+func HTTPHandlerReset() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		httpHandlerHijack(w, r, "reset")
+	})
+}
+
+// HTTPHandlerTimeout returns a handler that never returns a response and instead
+// blocks on the request context, this causing a client timeout.
```
bassosimone, line 132:
```suggestion
// blocks on the request context, thus causing a client timeout.
```

internal/testingx/tcpx.go
```diff
@@ -0,0 +1,44 @@
+package testingx
+
+import "net"
+
+// TCPListener creates TCP connections for HTTP. This type should work both
```
bassosimone, line 5:
```suggestion
// TCPListener creates TCP connections for HTTP, TLS, etc. This type should work both
```

internal/testingx/tcpx.go
```diff
@@ -0,0 +1,44 @@
...
+
+var _ TCPListener = &TCPListenerStdlib{}
+
+// ListenTCP implements HTTPListener.
+func (*TCPListenerStdlib) ListenTCP(network string, addr *net.TCPAddr) (net.Listener, error) {
+	return net.ListenTCP(network, addr)
+}
+
+// tcpMaybeResetNetConn is a portable mechanism to reset a net.Conn that takes into account
+// both TLS wrapping with any library and stdlib vs. netem concerns.
```
bassosimone, line 22:
```suggestion
// both TLS wrapping with any library and stdlib vs. netem concerns.
//
// Bug: netem is not WAI because there's no *gonet.TCPConn.SetLinger method.
```

internal/testingx/httptestx_test.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			// make sure we should not skip this test
+			if tc.reasonToSkip != "" {
+				t.Skip(tc.reasonToSkip)
+			}
+
+			// create a start topology for hosting the test
```
bassosimone, line 377:
```suggestion
			// create a star topology for hosting the test
```

PR #1161, opened 2023-06-19
probe-cli: Update to quic-go v0.37.3 

kelmenhorst:
## Checklist

- [x] I have read the [contribution guidelines](https://github.com/ooni/probe-cli/blob/master/CONTRIBUTING.md)
- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2492

## Description

This diff updates our QUIC code, mainly in netxlite, to quic-go v0.35.1, including the new API introduced in v0.35.0.
The migration has been done according to https://github.com/ooni/probe/issues/2492.

Major changes: 
* Replaced call to `quic.DialEarlyContext` with call to `quic.DialEarly` (netxlite/quic.go).
* Removed now unnecessary address string in `DialEarly` calls (netxlite/quic.go).
* Replaced `quic.Listener` with `*quic.Listener` (simplequicping_test.go).
* Adapted signature of `HandshakeComplete` such that it returns a channel rather than a context (quic code in netxlite, measurex, mocks).
* Renamed `QUICListener` to `UDPListener`.

~**Attention**: Tests using a failing `MockSyscallConn` (e.g. in measurexlite) fail, due to [issues with closing the quic.Transport quic-go](https://github.com/quic-go/quic-go/issues/3906).~

**Update (07.08.2023):**
Upgraded to quic-go 0.37.3 with the following changes:
* `quic.ConnectionState.TLS` points directly to a `tls.ConnectionState` instead of pointing to a `qtls.ConnectionState` instance. Background: Starting with Go 1.21, quic-go will use the QUIC support provided by the standard library's TLS implementation. Thus, quic-go will no longer fork crypto/tls anymore.
* changed signature of `quic.Connection`'s `ReceiveMessage` method

PR #1245, opened 2023-09-08
fix(webconnectivityqa): Close the QAEnv

kelmenhorst:
## Checklist

- [x] reference for this pull request: Error emerged here: https://github.com/ooni/probe-cli/pull/1161

## Description

This diff fixes the panics caused by the new quic-go API. 
We did not close the `QAEnv` in some of our webconnectivityqa tests. Therefore, the QUIC Server would not close the underlying UDP socket (`net.PacketConn`) and the local address of this socket would not be removed from the QUIC multiplexer. 
In the subsequent test case, the new quic-go API (see https://github.com/ooni/probe-cli/pull/1161) would panic because the local address of the server would still be registered in the global QUIC multiplexer.


bassosimone (approved):
🐳

PR #1270, opened 2023-09-14
feat(netxlite): add flexible HTTP transport factory

bassosimone:
Now that we've clearly labeled and packaged technical debt, we can copy existing technical-debt-ridden code and modify it to obtain a flexible factory for creating HTTP transports, `NewHTTPTransportWithOptions`.

In particular, this factory uses sensible defaults for measuring and there are options that you can pass it to customize details such as the backend proxy that we previously unconditionally configured.

More in detail, this is a side-by-side comparison between new code's defaults and old code:

| Setting Name        | httpquirks.go (old code) | httpfactory.go  (new code) |
| ------------------- | ------------------------ | -------------------------- |
| .Proxy              | nil                      | nil                        |
| .MaxConnsPerHost    | 1                        | ooni/oohttp's default      |
| .DisableCompression | true                     | true                       |
| .ForceAttemptHTTP2  | true                     | true                       |

So, basically, the biggest change is that we've removed the limitation of the max number of connections per host being set to 1. In any case, the new code allows to configure each of these fields.

This factory will be the starting point for having custom network functions for the engine in line with https://github.com/ooni/probe/issues/2531.

While there, acknowledge that `NewHTTPClient` contains technical debt because it does unconditionally disable cookies, to document that and move it inside of the proper file (`httpquirks.go`).

### DIFF
--- a/internal/netxlite/httpfactory.go
+++ b/internal/netxlite/httpfactory.go
@@ -27,7 +27,12 @@ type HTTPTransportOption func(txp *oohttp.Transport)
 // HTTP_PROXY and HTTPS_PROXY environment variables, which is required if
 // we want to use this code for measuring;
 //
-// 2. the .ForceAttemptHTTP2 field is set to true.
+// 2. the .ForceAttemptHTTP2 field is set to true;
+//
+// 3. the .DisableCompression field is set to true, again required if we
+// want to use this code for measuring, and we should make sure the defaults
+// we're using are suitable for measuring, since the impact of making a
+// mistake in measuring code is a data quality issue 😅.
 //
 // The returned transport supports logging and error wrapping because
 // internally this function calls [WrapHTTPTransport] before we return.
@@ -46,6 +51,7 @@ func NewHTTPTransportWithOptions(logger model.Logger,
 	txp.DialTLSContext = tlsDialer.DialTLSContext
 
 	// As documented, disable proxies and force HTTP/2
+	txp.DisableCompression = true
 	txp.Proxy = nil
 	txp.ForceAttemptHTTP2 = true
 
--- a/internal/netxlite/httpfactory.go
+++ b/internal/netxlite/httpfactory.go
@@ -1,12 +1,89 @@
 package netxlite
 
 import (
-	"net/http"
+	"net/url"
 
+	oohttp "github.com/ooni/oohttp"
 	"github.com/ooni/probe-cli/v3/internal/model"
 )
 
-// NewHTTPClient creates a new, wrapped HTTPClient using the given transport.
-func NewHTTPClient(txp model.HTTPTransport) model.HTTPClient {
-	return WrapHTTPClient(&http.Client{Transport: txp})
+// HTTPTransportOption is an initialization option for [NewHTTPTransport].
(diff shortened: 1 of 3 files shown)


internal/netxlite/httpfactory.go
```diff
@@ -1,12 +1,89 @@
...
+
+// NewHTTPTransport is the high-level factory to create a [model.HTTPTransport] using
+// github.com/ooni/oohttp as the HTTP library with HTTP/1.1 and HTTP2 support.
+//
+// This transport is suitable for HTTP2 and HTTP/1.1 using any TLS
+// library, including, e.g., github.com/ooni/oocrypto.
+//
+// This factory clones the default github.com/ooni/oohttp transport and
+// configures the provided dialer and TLS dialer by setting the .DialContext
+// and .DialTLSContext fields of the transport. We'll wrap the provided
```
bassosimone, line 21:
```suggestion
// and .DialTLSContext fields of the transport. We also wrap the provided
```

internal/netxlite/httpfactory.go
```diff
@@ -1,12 +1,89 @@
...
+		TLSDialer:     tlsDialer,
+	})
+}
+
+// HTTPTransportOptionProxyURL configures the transport to use the given proxyURL
+// or disables proxying (already the default) if the proxyURL is nil.
+func HTTPTransportOptionProxyURL(proxyURL *url.URL) HTTPTransportOption {
+	return func(txp *oohttp.Transport) {
+		txp.Proxy = func(r *oohttp.Request) (*url.URL, error) {
+			return proxyURL, nil
```
bassosimone, line 70:
```suggestion
			// "If Proxy is nil or returns a nil *URL, no proxy is used."
			return proxyURL, nil
```

internal/netxlite/httpfactory.go
```diff
@@ -1,12 +1,96 @@
...
+// HTTPTransportOptionMaxConnsPerHost configures the .MaxConnPerHosts field, which
+// otherwise uses the default set in github.com/ooni/oohttp.
+func HTTPTransportOptionMaxConnsPerHost(value int) HTTPTransportOption {
+	return func(txp *oohttp.Transport) {
+		txp.MaxConnsPerHost = value
+	}
+}
+
+// HTTPTransportOptionDisableCompression configures the .DisableCompression field, which
+// otherwise uses the default set in github.com/ooni/oohttp, i.e., false.
```
bassosimone, line 91:
```suggestion
// otherwise is set to true, so that this code is ready for measuring out of the box.
```

internal/netxlite/httpfactory_test.go
```diff
@@ -0,0 +1,170 @@
...
+			t.Fatal("invalid TLS dialer")
+		}
+
+		// make sure there's the stdlib adapter
+		stdlibAdapter := txpCloser.HTTPTransport.(*httpTransportStdlib)
+		oohttpStdlibAdapter := stdlibAdapter.StdlibTransport
+		underlying := oohttpStdlibAdapter.Transport
+
+		// now let's check that everything is configured as intended
+		// by applying the expected changes to a cloned default transport
```
bassosimone, line 49:
```suggestion
```

PR #1274, opened 2023-09-15
feat(testingx): introduce more comprehensive HTTP(S) proxy

bassosimone:
We want more comprehensive testing of how we use proxies during the bootstrap. Tests should encompass both SOCKS5 and HTTP(S) proxies. Tests should support using both the host network and using netem.

This diff starts paving the way for improving proxy testing, by introducing in `./internal/testingx/httpproxy.go` an HTTP(S) proxy implementation supporting both the case of HTTP over HTTP(S) proxy (where you use the host header) and HTTPS over HTTP(S) proxy (where the client issues a `CONNECT` request to the remote endpoint).

The previous implementation (in `./internal/testingx/httptestx.go`) is still there, for now. We used the previous implementation, which only supported the host header, as a starting point for writing the new one.

More in detail, this diff introduces the new proxy and its tests. Because testing the proxy functionality is a bit complex, I chose to use a separate package and also write tests for the tests. Obviously, we're still using `netxlite` as the underlying library, so there is some circularity in testing, but `netxlite` also has its own tests, so we should probably be fine.

The separate package with tests is `./internal/testingproxy`.

While working on this package, I realized the need to forward the CA used by the proxy. This happens in `./internal/testingproxy/hosthttps.go`. If we do not do this, we see the following failure:

```
=== RUN   TestWorkingAsIntended/fetching_https://www.example.com/_using_the_host_network_and_an_HTTPS_proxy
2023/09/15 15:28:39 debug > GET https://www.example.com/
2023/09/15 15:28:39 debug >                        

2023/09/15 15:28:39 dialerWithAssertions: verified that the network is tcp as expected
2023/09/15 15:28:39 dialerWithAssertions: verified that the address is 127.0.0.1 as expected

2023/09/15 15:28:39 debug dial 127.0.0.1:61772/tcp...
2023/09/15 15:28:39 debug dial_address 127.0.0.1:61772/tcp...
2023/09/15 15:28:39 debug dial_address 127.0.0.1:61772/tcp... ok in 212.083µs
2023/09/15 15:28:39 debug dial 127.0.0.1:61772/tcp... ok in 222.208µs

2023/09/15 15:28:39 debug tls {sni=127.0.0.1 next=[]}...
2023/09/15 15:28:39 debug tls {sni=127.0.0.1 next=[]}... ssl_unknown_authority in 3.687875ms

[...]
```

In other words, the TLS config we're using does not know the proxy CA. So, in this diff you also see the required work to forward the proxy CA, including the related netem changes in https://github.com/ooni/netem/pull/38.

The reference issue is https://github.com/ooni/probe/issues/2531.


internal/testingproxy/hosthttp.go
```diff
@@ -0,0 +1,74 @@
...
+func (tc *hostNetworkTestCaseWithHTTP) Run(t *testing.T) {
+	// create an instance of Netx where the underlying network is nil,
+	// which means we're using the host's network
+	netx := &netxlite.Netx{Underlying: nil}
+
+	// create the proxy server using the host network
+	proxyServer := testingx.MustNewHTTPServer(testingx.NewHTTPProxyHandler(log.Log, netx))
+	defer proxyServer.Close()
+
+	log.SetLevel(log.DebugLevel)
```
bassosimone, line 51:
```suggestion
	//log.SetLevel(log.DebugLevel)
```

internal/testingproxy/dialer.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"fmt"
+	"log"
+	"net"
+
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
+type dialerWithAssertions struct {
```
bassosimone, line 13:
```suggestion
// dialerWithAssertions ensures that we're dialing with the proxy address.
type dialerWithAssertions struct {
```

internal/testingproxy/httputils.go
```diff
@@ -0,0 +1,53 @@
...
+	t.MockFatal(v...)
+}
+
+// Logf implements httpTestingT.
+func (t *httpTestingTMock) Logf(format string, v ...any) {
+	t.MockLogf(format, v...)
+}
+
+func httpCheckResponse(t httpTestingT, client httpClient, targetURL string) {
+	// get the homepage and assert we're getting a succesful response
```
bassosimone, line 43:
```suggestion
```

PR #1282, opened 2023-09-19
feat(enginenetx): support HTTP and HTTPS proxies

bassosimone:
This diff completes the work we have been doing for a few days now and provides HTTP and HTTPS proxy support, in addition to SOCKS5 support, for the engine-specific network.

We did this work in the context of https://github.com/ooni/probe/issues/2531 and https://github.com/ooni/probe/issues/1955.

BTW, the fact that tests used `measurexlite` and tracing is very nice here. It means the idea to write `measurexlite` based on context and tracing was good and could be used beyond its original design goals.

### DIFF
--- a/internal/enginenetx/http_test.go
+++ b/internal/enginenetx/http_test.go
@@ -165,7 +165,7 @@ func TestHTTPTransportWAI(t *testing.T) {
 				t.Fatal("unexpected status code")
 			}
 
-			// make sure that we only connected to the SOCKS5 proxy
+			// make sure that we only connected to the HTTP proxy
 			tcpConnects := trace.TCPConnects()
 			if len(tcpConnects) <= 0 {
 				t.Fatal("expected at least one TCP connect")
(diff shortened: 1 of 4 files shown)


internal/enginenetx/http_test.go
```diff
@@ -1,34 +1,263 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			t.Logf("%+v", resp)
+			defer resp.Body.Close()
+			if resp.StatusCode != 200 {
+				t.Fatal("unexpected status code")
+			}
+
+			// make sure that we only connected to the SOCKS5 proxy
```
bassosimone, line 168:
```suggestion
			// make sure that we only connected to the HTTP proxy
```

internal/enginenetx/http_test.go
```diff
@@ -1,34 +1,263 @@
...
```
bassosimone, line 237:
```suggestion
			// make sure that we only connected to the HTTPS proxy
```

PR #1297, opened 2023-09-22
refactor(enginenetx): make static/loadable policy easier to use

bassosimone:
This diff makes the policy previously known as loadable and now know as the static policy easier to use, by having a constructor that reads from a key-value store and by passing it a fallback policy to use.

With this design, it should be possible to have code that uses the static policy if applied, falling back to whatever policy we are otherwise constructing into the NewNetwork constructor.

In a subsequent commit, I will hook this code into NewNetwork, so that we can override the policy by changing the filesystem.

While there, notice several failures in the test suite and apply workarounds (see https://github.com/ooni/probe/issues/2539, https://github.com/ooni/probe/issues/2540, https://github.com/ooni/probe/issues/2541).

Part of https://github.com/ooni/probe/issues/2531.


internal/enginenetx/httpsdialerstatic_test.go
```diff
@@ -0,0 +1,239 @@
...
+				{
+					Endpoint:       "162.55.247.208:443",
+					InitialDelay:   0,
+					SNI:            "www.example.org",
+					VerifyHostname: "api.ooni.io",
+				},
+			}
+
+			p := &HTTPSDialerStaticPolicy{
+				Fallback: nil, // explicitly such that there is a panic if we access it
```
bassosimone, line 183:
```suggestion
				Fallback: nil, // explicitly nil such that there is a panic if we access it
```

bassosimone:
I see this failure in the CI:

```
2023/09/22 13:34:42  info New reportID:            
2023/09/22 13:34:42  info New reportID:            
--- FAIL: TestMaybeLoginIdempotent (0.20s)
    login_test.go:62: httpx: request failed: 500 Internal Server Error
--- FAIL: TestFetchPsiphonConfig (0.20s)
    psiphon_test.go:16: httpx: request failed: 500 Internal Server Error
--- FAIL: TestFetchTorTargetsSetsQueryString (0.20s)
    tor_test.go:69: httpx: request failed: 500 Internal Server Error
FAIL
	github.com/ooni/probe-cli/v3/internal/probeservices	coverage: 91.9% of statements
FAIL	github.com/ooni/probe-cli/v3/internal/probeservices	4.103s
```

It seems the testing backend is having issues.

bassosimone:
There is obviously a coverage drop, but not in the package we wanted to modify in this PR:

Screenshot 2023-09-22 at 17 31 01

The drop, in fact, stems for the workarounds for broken tests I needed to commit.

PR #1299, opened 2023-09-25
feat(enginenetx): track operations and collect stats

bassosimone:
This diff modifies the enginenetx package such that we track the results of operations we perform and collect statistics.

The statistics will be saved in the `StateDir` where we already save other statistics, e.g., the ones related to DoH.

While there, realize that we should also count the number of interrupted operations. Also, realize that the TLS verification cannot be interrupted, so there's no point in passing a `ctx` to the related stats manager callback.

### DIFF
--- a/internal/enginenetx/httpsdialerstats.go
+++ b/internal/enginenetx/httpsdialerstats.go
@@ -1,0 +1,314 @@
+package enginenetx
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/ooni/probe-cli/v3/internal/model"
+	"github.com/ooni/probe-cli/v3/internal/runtimex"
+)
+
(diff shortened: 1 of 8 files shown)


internal/enginenetx/httpsdialerstats.go
```diff
@@ -0,0 +1,314 @@
...
+	// update stats
+	record.CountStarted++
+	record.LastUpdated = mt.TimeNow()
+}
+
+// OnTCPConnectError implements HTTPSDialerStatsManager.
+func (mt *HTTPSDialerStatsManager) OnTCPConnectError(ctx context.Context, tactic *HTTPSDialerTactic, err error) {
+	// avoid updating stats if the operation has been interrupted
+	if ctx.Err() != nil {
+		return
```
bassosimone, line 218: Maybe we should have an "interrupted" counter here, otherwise we would be missing some number of operations from the overall statistics, which would confuse people looking at the data.

PR #1307, opened 2023-09-26
refactor(enginenetx): group by domain's endpoint

bassosimone:
There's no point in grouping tactics by "www.example.com" for stats stored on disk and user-provided static policies. By grouping by domain alone, we'll end up with mixed tactics for different ports all being together. Which means that, code that needs to filter, needs to explicitly filter by port. If we used "www.example.com:443" as the group-by key, conversely, we would not need to write specialized filtering code to only get data for the correct port. So, let's do that and let's simplify future code that will need to get tactics known for work for a given domain _and_ port.

### DIFF
--- a/internal/enginenetx/httpsdialercore.go
+++ b/internal/enginenetx/httpsdialercore.go
@@ -71,6 +71,11 @@ func (dt *HTTPSDialerTactic) Summary() string {
 	return fmt.Sprintf("%v sni=%v verify=%v", net.JoinHostPort(dt.Address, dt.Port), dt.SNI, dt.VerifyHostname)
 }
 
+// domainEndpointKey returns the key domain endpoint key for storing into a map.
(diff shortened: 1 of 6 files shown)


internal/enginenetx/httpsdialercore.go
```diff
@@ -71,6 +71,11 @@ func (dt *HTTPSDialerTactic) Summary() string {
 	return fmt.Sprintf("%v sni=%v verify=%v", net.JoinHostPort(dt.Address, dt.Port), dt.SNI, dt.VerifyHostname)
 }
 
+// domainEndpointKey returns the key domain endpoint key for storing into a map.
```
bassosimone, line 74:
```suggestion
// domainEndpointKey returns the domain's endpoint string key for storing into a map.
```

PR #1312, opened 2023-09-26
feat(enginenetx): add policy based on stats

bassosimone:
This policy first checks the stats and sorts the entries for which we have the highest success rate. Then, we use an underlying policy to produce suitable tactics for the HTTPS dialer.

While there, rename statstracker{,_test}.go to statsmanager{,_test}.go because the type inside it is called "state manager". The state tracker is an abstract concept only used by the https dialer code.


internal/enginenetx/statspolicy.go
```diff
@@ -0,0 +1,92 @@
+package enginenetx
+
+//
+// Schedling policy based on stats that fallbacks to
```
bassosimone, line 4:
```suggestion
// Scheduling policy based on stats that fallbacks to
```

internal/enginenetx/statspolicy.go
```diff
@@ -0,0 +1,92 @@
...
+	out := make(chan *httpsDialerTactic)
+
+	go func() {
+		index := 0
+		defer close(out)
+
+		// make sure we don't emit two equal policy in a single run
+		uniq := make(map[string]int)
+
+		// function that maybeEmitTactic a given tactic
```
bassosimone, line 41:
```suggestion
		// function that emits a given tactic unless we already emitted it
```

PR #1316, opened 2023-09-27
fix(enginenetx): gracefully handle more nil cases

bassosimone:
This diff adjusts code inside the enginenetx package such that we gracefully deal with nil values in more cases.

### DIFF
--- a/internal/enginenetx/statspolicy_test.go
+++ b/internal/enginenetx/statspolicy_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ooni/probe-cli/v3/internal/netemx"
 	"github.com/ooni/probe-cli/v3/internal/netxlite"
 	"github.com/ooni/probe-cli/v3/internal/runtimex"
+	"github.com/ooni/probe-cli/v3/internal/testingx"
 )
 
 func TestStatsPolicyWorkingAsIntended(t *testing.T) {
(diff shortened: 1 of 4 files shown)


internal/enginenetx/statspolicy_test.go
```diff
@@ -314,3 +315,43 @@ var _ httpsDialerPolicy = &mocksPolicy{}
 func (p *mocksPolicy) LookupTactics(ctx context.Context, domain string, port string) <-chan *httpsDialerTactic {
 	return p.MockLookupTactics(ctx, domain, port)
 }
+
+func TestStatsPolicyPostProcessTactics(t *testing.T) {
+	t.Run("we do nothing when bool is false", func(t *testing.T) {
```
bassosimone, line 320:
```suggestion
	t.Run("we do nothing when good is false", func(t *testing.T) {
```

internal/enginenetx/statspolicy_test.go
```diff
@@ -314,3 +315,43 @@ var _ httpsDialerPolicy = &mocksPolicy{}
...
+		tactics := statsPolicyPostProcessTactics(nil, false)
+		if len(tactics) != 0 {
+			t.Fatal("expected zero-lenght return value")
+		}
+	})
+
+	t.Run("we filter out cases in which a .Tactic field is nil", func(t *testing.T) {
```
bassosimone, line 327:
```suggestion
	t.Run("we filter out cases in which t or t.Tactic are nil", func(t *testing.T) {
```

PR #1317, opened 2023-09-27
fix(enginenetx): periodically trim statistics

bassosimone:
If we're running the enginenetx code inside an environment that sprays ~random IP addresses during DNS lookup, we're going to end up using quite a large amount of statistics. While this amount of statistics could potentially be great in general, we need to be mindful of memory and disk occupation as well as of the overhead of serializing and deserializing.

To address these concerns, add the arbitrary rule that we don't want more than ten domain-endpoint entries and implement code to zap old and/or always-failing entries. The code will run when loading the entries, when serializing them, and periodically while the `*Network` is up and running.

See https://github.com/ooni/probe/issues/2531

### DIFF
--- a/internal/enginenetx/statsmanager.go
+++ b/internal/enginenetx/statsmanager.go
@@ -89,10 +89,10 @@ type statsTactic struct {
 }
 
 // statsNilSafeSuccessRate is a convenience function for computing the success rate
-// which returns zero as the success rate if CountStarted is zero
+// which returns zero as the success rate if CountStarted is zero.
 //
-// for robustness, be paranoid about nils here because the stats are
-// written on the disk and a user could potentially edit them
+// For robustness, be paranoid about nils here because the stats are
+// written on the disk and a user could potentially edit them.
 func statsNilSafeSuccessRate(t *statsTactic) (rate float64) {
 	if t != nil && t.CountStarted > 0 {
 		rate = float64(t.CountSuccess) / float64(t.CountStarted)
@@ -379,10 +379,10 @@ type statsManager struct {
 	mu sync.Mutex
 
 	// pruned is a channel pruned on a best effort basis
-	// by the background goroutine that prunes
+	// by the background goroutine that prunes.
 	pruned chan any
 
-	// wg tells us when the background goroutine joined
+	// wg tells us when the background goroutine joined.
 	wg *sync.WaitGroup
 }
 
@@ -629,7 +629,7 @@ func (mt *statsManager) trim(ctx context.Context, interval time.Duration) {
 			mt.mu.Unlock()
 
 			// notify whoever's concerned that we pruned
-			// on a best effort basis
+			// and do that best effort because it may be that nobody is concerned
 			select {
 			case mt.pruned <- true:
 			default:
--- a/internal/enginenetx/statsmanager.go
+++ b/internal/enginenetx/statsmanager.go
@@ -149,10 +149,10 @@ func statsDefensivelySortTacticsByDescendingSuccessRateWithAcceptPredicate(
 		if statsNilSafeSuccessRate(work[i]) > statsNilSafeSuccessRate(work[j]) {
 			return true
 		}
-		if statsNilSafeLastUpdated(work[i]).Sub(statsNilSafeLastUpdated(work[j])) > 0 {
+		if statsNilSafeCountSuccess(work[i]) > statsNilSafeCountSuccess(work[j]) {
 			return true
 		}
-		if statsNilSafeCountSuccess(work[i]) > statsNilSafeCountSuccess(work[j]) {
+		if statsNilSafeLastUpdated(work[i]).Sub(statsNilSafeLastUpdated(work[j])) > 0 {
 			return true
 		}
 		return false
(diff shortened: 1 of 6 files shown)


internal/enginenetx/statsmanager_test.go
```diff
@@ -1043,3 +1074,562 @@ func TestStatsContainer(t *testing.T) {
...
+		st := &statsTactic{
+			LastUpdated: expect,
+		}
+		if statsNilSafeLastUpdated(st) != expect {
+			t.Fatal("unexpected result")
+		}
+	})
+}
+
+func TestStatsNilSafeCounSuccess(t *testing.T) {
```
bassosimone, line 1116:
```suggestion
func TestStatsNilSafeCountSuccess(t *testing.T) {
```

internal/enginenetx/statsmanager.go
```diff
@@ -87,6 +88,86 @@ type statsTactic struct {
 	Tactic *httpsDialerTactic
 }
 
+// statsNilSafeSuccessRate is a convenience function for computing the success rate
+// which returns zero as the success rate if CountStarted is zero
```
bassosimone, line 92:
```suggestion
// which returns zero as the success rate if CountStarted is zero.
```

internal/enginenetx/statsmanager.go
```diff
@@ -87,6 +88,86 @@ type statsTactic struct {
...
+//
+// for robustness, be paranoid about nils here because the stats are
+// written on the disk and a user could potentially edit them
```
bassosimone, line 95:
```suggestion
// For robustness, be paranoid about nils here because the stats are
// written on the disk and a user could potentially edit them.
```

internal/enginenetx/statsmanager.go
```diff
@@ -265,6 +377,13 @@ type statsManager struct {
 
 	// mu provides mutual exclusion when accessing the stats.
 	mu sync.Mutex
+
+	// pruned is a channel pruned on a best effort basis
+	// by the background goroutine that prunes
```
bassosimone, line 382:
```suggestion
	// by the background goroutine that prunes.
```

internal/enginenetx/statsmanager.go
```diff
@@ -265,6 +377,13 @@ type statsManager struct {
...
+	pruned chan any
+
+	// wg tells us when the background goroutine joined
```
bassosimone, line 385:
```suggestion
	// wg tells us when the background goroutine joined.
```

internal/enginenetx/statsmanager_test.go
```diff
@@ -1043,3 +1074,562 @@ func TestStatsContainer(t *testing.T) {
...
+			DomainEndpoints: map[string]*statsDomainEndpoint{},
+			Version:         statsContainerVersion,
+		}
+
+		if diff := cmp.Diff(expect, output); diff != "" {
+			t.Fatal(diff)
+		}
+	})
+
+	t.Run("we avoid including into the results empty .Tactics", func(t *testing.T) {
```
bassosimone, line 1490:
```suggestion
	t.Run("we avoid including into the results expired entries", func(t *testing.T) {
```

internal/enginenetx/statsmanager_test.go
```diff
@@ -1043,3 +1074,562 @@ func TestStatsContainer(t *testing.T) {
...
+		}
+
+		if diff := cmp.Diff(expect, output); diff != "" {
+			t.Fatal(diff)
+		}
+	})
+}
+
+func TestStatsManagerTrimEntriesConcurrently(t *testing.T) {
+	// start stats manger that trims very frequently
```
bassosimone, line 1583:
```suggestion
	// start stats manager that trims very frequently
```

internal/enginenetx/statsmanager_test.go
```diff
@@ -1043,3 +1074,562 @@ func TestStatsContainer(t *testing.T) {
...
+	<-sm.pruned
+
+	// order the background goroutine to shutdown
+	// and wait for the shutdown to complete
+	if err := sm.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	// now check what actually ended up being written; note that we expect
+	// to see empty domain endpoitns because we added a too old entry
```
bassosimone, line 1626:
```suggestion
	// to see empty domain endpoints because we added a too old entry
```

internal/enginenetx/statsmanager_test.go
```diff
@@ -1043,3 +1074,562 @@ func TestStatsContainer(t *testing.T) {
...
+	expecteData := []byte(`{"DomainEndpoints":{},"Version":5}`)
```
bassosimone, line 1627:
```suggestion
	expectedData := []byte(`{"DomainEndpoints":{},"Version":5}`)
```

internal/enginenetx/statsmanager_test.go
```diff
@@ -1043,3 +1074,562 @@ func TestStatsContainer(t *testing.T) {
...
+	data, err := store.Get(statsKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if diff := cmp.Diff(expecteData, data); diff != "" {
```
bassosimone, line 1632:
```suggestion
	if diff := cmp.Diff(expectedData, data); diff != "" {
```

PR #1321, opened 2023-09-28
feat(model): introduce ArchivalBinaryData

bassosimone:
There are cases where we know we have binary data in input and we want the output to be binary data using the dictionary encoding like `{"format":"base64","data":"..."}`.

Such cases are, for example, DNS messages bodies.

There is no need for us to pass through the MaybeArchivalBinaryData in such cases. Additionally, this makes MaybeArchivalBinaryData a bit more complex than it probably needs to be.

What's more, ArchivalBinaryData is for when you do not require to scrub the data. I want to introduce new data types that automatically perform scrubbing when they're used.

But this puts even more pressure on MaybeArchivalBinaryData, and hence this commit, to start differentiating between:

1. always binary data vs maybe binary data

2. no scrubbing required vs scrubbing required

The rationale for doing this set of changes has been explained in https://github.com/ooni/probe-cli/pull/1319.

The reference issue is https://github.com/ooni/probe/issues/2531.

For now, this commit just adds the new type and tests for it without using the type, which we'll do later once we have added better marshal/unmarshal testing for the interested types.

### DIFF
--- a/internal/model/archival_test.go
+++ b/internal/model/archival_test.go
@@ -70,7 +70,7 @@ func TestArchivalBinaryData(t *testing.T) {
 			expectErr:  nil,
 			expectData: []byte("null"),
 		}, {
-			name:       "with .Value being simple binary string",
+			name:       "with .Value being a simple binary string",
 			input:      model.ArchivalBinaryData{Value: []byte("Elliot")},
 			expectErr:  nil,
 			expectData: []byte(`{"data":"RWxsaW90","format":"base64"}`),
@@ -115,7 +115,7 @@ func TestArchivalBinaryData(t *testing.T) {
 	})
 
 	// This test verifies that we correctly parse binary data to JSON by
-	// readeding from null | {"format":"base64","data":"<base64>"}
+	// reading from null | {"format":"base64","data":"<base64>"}
 	t.Run("UnmarshalJSON", func(t *testing.T) {
 		// testcase is a test case defined by this function
 		type testcase struct {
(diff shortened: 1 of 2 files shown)


internal/model/archival_test.go
```diff
@@ -37,6 +40,259 @@ var archivalBinaryInput = []uint8{
...
+			input:      model.ArchivalBinaryData{Value: nil},
+			expectErr:  nil,
+			expectData: []byte("null"),
+		}, {
+			name:       "with zero length .Value",
+			input:      model.ArchivalBinaryData{Value: []byte{}},
+			expectErr:  nil,
+			expectData: []byte("null"),
+		}, {
+			name:       "with .Value being simple binary string",
```
bassosimone, line 73:
```suggestion
			name:       "with .Value being a simple binary string",
```

internal/model/archival_test.go
```diff
@@ -37,6 +40,259 @@ var archivalBinaryInput = []uint8{
...
+
+				if diff := cmp.Diff(tc.expectData, output); diff != "" {
+					t.Fatal(diff)
+				}
+			})
+		}
+	})
+
+	// This test verifies that we correctly parse binary data to JSON by
+	// readeding from null | {"format":"base64","data":"<base64>"}
```
bassosimone, line 118:
```suggestion
	// reading from null | {"format":"base64","data":"<base64>"}
```

internal/model/archival_test.go
```diff
@@ -37,6 +40,259 @@ var archivalBinaryInput = []uint8{
...
+		}
+
+		cases := []testcase{{
+			name:  "with nil .Value",
+			input: model.ArchivalBinaryData{Value: nil},
+		}, {
+			name:  "with zero length .Value",
+			input: model.ArchivalBinaryData{Value: []byte{}},
+		}, {
+			name:  "with .Value being simple binary string",
```
bassosimone, line 254:
```suggestion
			name:  "with .Value being a simple binary string",
```

PR #1339, opened 2023-10-04
chore: upgrade C dependencies

bassosimone:
This diff updates C dependencies to the latest version.

I'm using `androidbuild` as the PR branch name such that we trigger an Android build.

Part of https://github.com/ooni/probe/issues/2524


bassosimone:
I think it's worth mentioning I _expect_ this PR to fail, since it the Android build has failed locally for me 😅

bassosimone:
Now it also works locally 🤷. What changed is that I ran `make android` as opposed to running individual commands going on memory. Perhaps, this is the reason why before it has failed. Marking this PR as ready again.

PR #1347, opened 2023-10-06
chore: update obfs4, goptlib, snowflake

bassosimone:
This diff upgrades obfs4, goptlib, and snowflake. I tried to import lyrebird instead of obfs4, but that did not work as intended. It seems tags aren't compatible with the format expected by Go (vX.Y.Z) and I did not want to pin to the latest commit. I will report this issue to lyrebird developers later today.

While there, also upgrade x/net and x/exp.

This work is part of https://github.com/ooni/probe/issues/2524.

Here's what we could not upgrade (based on the content of `go.mod` after we committed this diff and ran `go get -u -v -d ./...`):

```diff
diff --git a/go.mod b/go.mod
index 67e0af02..18175d73 100644
--- a/go.mod
+++ b/go.mod
@@ -80,7 +80,7 @@ require (
 	golang.org/x/exp/typeparams v0.0.0-20230522175609-2e198f4a06a1 // indirect
 	golang.org/x/time v0.3.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	gvisor.dev/gvisor v0.0.0-20230603040744-5c9219dedd33 // indirect
+	gvisor.dev/gvisor v0.0.0-20231006032704-15cc3fcbbd77 // indirect
 )

 require (
@@ -89,7 +89,7 @@ require (
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
 	github.com/armon/go-proxyproto v0.0.0-20210323213023-7e956b284f0a // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/bifurcation/mint v0.0.0-20180306135233-198357931e61 // indirect
+	github.com/bifurcation/mint v0.0.0-20210616192047-fd18df995463 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/cheekybits/genny v1.0.0 // indirect
 	github.com/cognusion/go-cache-lru v0.0.0-20170419142635-f73e2280ecea // indirect
@@ -132,7 +132,7 @@ require (
 	github.com/prometheus/common v0.44.0 // indirect
 	github.com/prometheus/procfs v0.12.0 // indirect
 	github.com/refraction-networking/gotapdance v1.7.4 // indirect
-	github.com/refraction-networking/utls v1.3.3 // indirect
+	github.com/refraction-networking/utls v1.5.3 // indirect
[...]

PR #1348, opened 2023-10-06
chore: replace obfs4 with lyrebird

bassosimone:
This diff replaces obfs4 and uses lyrebird instead. See https://github.com/ooni/probe-cli/pull/1347#issuecomment-1750086765 to understand why my previous assessment that we could not import lyrebird was actually caused by PEBCAK.

The build currently fails because the set of dependencies required by Psiphon conflict with lyrebird ones, so upgrading isn't possible. Here is how the build fails:

```console
$ go build -v ./...
github.com/Psiphon-Labs/psiphon-tunnel-core/psiphon/common/protocol
/home/sbs/go/pkg/mod/github.com/!psiphon-!labs/psiphon-tunnel-core@v1.0.11-0.20230822172011-3f91b1b804b1/psiphon/common/protocol/customTLSProfiles.go:180:16: undefined: utls.UtlsExtendedMasterSecretExtension
```

So, I think the underlying issue here may be that utls has removed some symbols without bumping their major version number.

If it was possible to merge this work right now, it would have been part of the https://github.com/ooni/probe/issues/2524 issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,9 +35,9 @@ require (
 	github.com/rubenv/sql-migrate v1.5.2
 	github.com/schollz/progressbar/v3 v3.13.1
 	github.com/upper/db/v4 v4.6.0
-	gitlab.com/yawning/obfs4.git v0.0.0-20231005123604-19f5a37fe427
 	gitlab.com/yawning/utls.git v0.0.12-1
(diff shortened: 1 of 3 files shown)


bassosimone:
This PR served to document why we cannot update to lyrebird right now. I'm going to close it now.

PR #1352, opened 2023-10-09
refactor: go test -short does not use the host network

bassosimone:
This diff refactors the codebase so that we avoid using the host network when running `go test -short`. This change is good in general, because now coverage tells us the amount of code we're covering without depending on interactions with an existing network, which means these tests behave in the same way in ~any place.

I expect a coverage drop from this PR, because there's some coverage made with integration testing (if we consider integration tests the tests that require the host network interface with uncensored internet access).

To make this happen, I needed to modify the `quictesting` package (now moved to toplevel and renamed `testingquic`) such that it attempts to get a known-to-work-well endpoint for QUIC _only_ when the developer using the package really needs it, rather than on import. Before doing this, there were several tests that panicked because `quictesting` could not figure out which IP address to use when you disable the WiFi or run inside another netns. Now we only figure this IP address out the first time a test requires us to give it either the domain or the endpoint that we should use.

To be sure we continue to honour the promise that `go test -short` does not use the host network, I needed to refactor the CI such that we measure coverage inside a new network namespace with only localhost support. I think this compromise is acceptable, since the original ask was to avoid flaky network tests (see https://github.com/ooni/probe/issues/2426).

Because of this change in how we run the coverage checks, I am tentatively enabling also running all tests for pull requests, otherwise we don't know if a contribution breaks tests using the network. Hopefully, we should be fine because we are caching previous runs, so a bunch of tests should already be cached.

Closes https://github.com/ooni/probe/issues/2426.

While there, enable again some backend integration tests, and close https://github.com/ooni/probe/issues/2539.


script/linuxcoveragerun.bash
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+#
+# Script invoked by ./script/linuxcoverage.bash to run tests
+# with coverage using a separate namespace with loopback
```
bassosimone, line 5:
```suggestion
# with coverage using a separate network namespace with only loopback support.
```

script/linuxcoveragerun.bash
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+#
+# Script invoked by ./script/linuxcoverage.bash to run tests
+# with coverage using a separate network namespace with only loopback support.
```
bassosimone, line 5:
```suggestion
# Script invoked by ./script/linuxcoverage.bash to run tests with coverage
# using a separate network namespace with only loopback support.
```

PR #1364, opened 2023-10-11
feat(riseupvpn): include bootstrap in progress

bassosimone:
This diff modifies riseupvpn to include bootstrap in progress.

We define as riseupvpn's bootstrap the process of fetching the CA and the various JSON files needed to find out the gateways to measure.

I previously stopped including bootstrap in progress in https://github.com/ooni/probe-cli/pull/1363, which I did because the progress was not monotonic.

This diff introduces helpers that make it very obvious to emit monotonic bootstrap through the ./internal/progress package.

This work is part of https://github.com/ooni/probe/issues/1432.

### DIFF
--- a/internal/progress/progress_test.go
+++ b/internal/progress/progress_test.go
@@ -1,0 +1,87 @@
+package progress
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/ooni/probe-cli/v3/internal/model"
+)
+
+type capturerCallbacks struct {
+	value float64
+}
+
+var _ model.ExperimentCallbacks = &capturerCallbacks{}
+
+// OnProgress implements model.ExperimentCallbacks.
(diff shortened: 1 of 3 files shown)


internal/progress/progress_test.go
```diff
@@ -0,0 +1,87 @@
...
+		emit:   []float64{0, 0.2, 0.4, 0.6, 0.8, 1},
+		expect: []float64{0, 0.1, 0.2, 0.3, 0.4, 0.5},
+	}, {
+		name:   "with offset==0.5 and total=1",
+		offset: 0.5,
+		total:  1,
+		emit:   []float64{0, 0.2, 0.4, 0.6, 0.8, 1},
+		expect: []float64{0.5, 0.6, 0.7, 0.8, 0.9, 1},
+	}, {
+		name:   "with offset==0.2 and total=0.8",
```
bassosimone, line 60:
```suggestion
		name:   "with offset==0.2 and total=0.7",
```

PR #1366, opened 2023-10-12
feat(ios): replace go-libtor with ./internal/libtor

bassosimone:
This diff replaces go-libtor with ./internal/libtor.

Closes https://github.com/ooni/probe/issues/2564


bassosimone:
Alright, enough hammering for the build to succeed. Obviously, a successful build does not imply we're also producing correct binaries, but a failed build is definitely not providing binaries, so this counts as progress 😅

PR #1369, opened 2023-10-12
fix(buildtool): let libevent use OpenSSL's pkgconfig

bassosimone:
This diff modifies `./internal/cmd/buildtool` to (1) stop removing OpenSSL's pkgconfig, (2) use such a pkgconfig during libevent's `./configure` step, and (3) make sure libevent's install does not remove OpenSSL's pkgconfig.

We're implementing changes (1) and (2) because libevent fails to properly configure for iOS if you have a system with homebrew with OpenSSL@3 installed. Unless we force the `PKG_CONFIG_PATH`, libevent's `configure` uses the OpenSSL@3 headers and library installed with homebrew rather than the ones we cross compiled.

Change (3) is there because otherwise we cannot recompile libevent without recompiling OpenSSL, since the libevent build fails unless we have a suitable OpenSSL pkgconfig, as explained before.

This diff was extracted and refined from https://github.com/ooni/probe-cli/pull/1366.

The reference issue is: https://github.com/ooni/probe/issues/2564.


internal/cmd/buildtool/cdepslibevent.go
```diff
@@ -49,6 +49,10 @@ func cdepsLibeventBuildMain(globalEnv *cBuildEnv, deps buildtoolmodel.Dependenci
 	}
 	envp := cBuildExportAutotools(cBuildMerge(globalEnv, localEnv))
 
+	// On iOS, we need PKG_CONFIG_PATH to convince libevent to use the OpenSSL we built but]
```
bassosimone, line 52:
```suggestion
	// On iOS, we need PKG_CONFIG_PATH to convince libevent to use the OpenSSL we built but
```

internal/cmd/buildtool/cdepslibevent.go
```diff
@@ -49,6 +49,10 @@ func cdepsLibeventBuildMain(globalEnv *cBuildEnv, deps buildtoolmodel.Dependenci
...
+	// always letting libevent's configure be pkgconfig aware would probably be fine
```
bassosimone, line 53:
```suggestion
	// always letting libevent's configure use pkgconfig is actually fine.
```

internal/cmd/buildtool/cdepsopenssl.go
```diff
@@ -82,5 +82,8 @@ func cdepsOpenSSLBuildMain(globalEnv *cBuildEnv, deps buildtoolmodel.Dependencie
 	))
 
 	must.Run(log.Log, "make", "DESTDIR="+globalEnv.DESTDIR, "install_dev")
-	must.Run(log.Log, "rm", "-rf", filepath.Join(globalEnv.DESTDIR, "lib", "pkgconfig"))
+
+	// We used to delete the pkgconfig but it turns out this is import for iOS builds, which
```
bassosimone, line 86:
```suggestion
	// We used to delete the pkgconfig but it turns out this is important for libevent iOS builds, which
```

internal/cmd/buildtool/cdepsopenssl.go
```diff
@@ -82,5 +82,8 @@ func cdepsOpenSSLBuildMain(globalEnv *cBuildEnv, deps buildtoolmodel.Dependencie
 	))
 
 	must.Run(log.Log, "make", "DESTDIR="+globalEnv.DESTDIR, "install_dev")
-	must.Run(log.Log, "rm", "-rf", filepath.Join(globalEnv.DESTDIR, "lib", "pkgconfig"))
+
+	// We used to delete the pkgconfig but it turns out this is important for libevent iOS builds, which
+	// means now we need to keep it. See https://github.com/ooni/probe-cli/pull/1366 for details.
```
bassosimone, line 87:
```suggestion
	// means now we need to keep it. See https://github.com/ooni/probe-cli/pull/1369 for details.
```

internal/cmd/buildtool/cdepslibevent.go
```diff
@@ -59,7 +63,17 @@ func cdepsLibeventBuildMain(globalEnv *cBuildEnv, deps buildtoolmodel.Dependenci
 	must.Run(log.Log, "make", "V=1", "-j", strconv.Itoa(runtime.NumCPU()))
 	must.Run(log.Log, "make", "DESTDIR="+globalEnv.DESTDIR, "install")
 	must.Run(log.Log, "rm", "-rf", filepath.Join(globalEnv.DESTDIR, "bin"))
-	must.Run(log.Log, "rm", "-rf", filepath.Join(globalEnv.DESTDIR, "lib", "pkgconfig"))
+
+	// We used to delete the whole pkgconfig directory but libevent's build needs OpenSSL's
+	// pkgconfig, so removing the whole directory means rebuilding libevent requires building
+	// OpenSSL because its pkgconfig is also removed. We discovered this need when working
+	// on the https://github.com/ooni/probe-cli/pull/1366 MVP. To keep the build idempotent,
```
bassosimone, line 70:
```suggestion
	// on the https://github.com/ooni/probe-cli/pull/1369 PR. To keep the build idempotent,
```

PR #1371, opened 2023-10-12
chore(buildtool): add ios-build unit tests

bassosimone:
We needed to move the runtimex.Assert for darwin, otherwise we cannot run unit tests on linux 😅. While there, be consistent and make sure we avoid panicking on Windows if one runs unit tests there.

Part of https://github.com/ooni/probe/issues/2564

### DIFF
--- a/internal/cmd/buildtool/internal/buildtooltest/buildtooltest.go
+++ b/internal/cmd/buildtool/internal/buildtooltest/buildtooltest.go
@@ -85,7 +85,7 @@ func CheckSingleCommand(cmd *execabs.Cmd, tee ExecExpectations) error {
 		return err
 	}
 	if err := CompareEnv(tee.Env, shellxtesting.CmdEnvironMinusOsEnviron(cmd)); err != nil {
-		return err
+		return fmt.Errorf("in %v: %w", tee.Argv, err)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


internal/cmd/buildtool/internal/buildtooltest/buildtooltest.go
```diff
@@ -253,12 +253,12 @@ func (*DependenciesCallCounter) XCRun(args ...string) string {
...
 		runtimex.Assert(len(args) == 3, "expected three arguments")
 		runtimex.Assert(args[2] == "--show-sdk-path", "the third argument must be --show-sdk-path")
-		return filepath.Join("Developer", "SDKs", args[1])
+		return string(filepath.Separator) + filepath.Join("Developer", "SDKs", args[1])
 
 	case "-find":
 		runtimex.Assert(len(args) == 4, "expected four arguments")
 		runtimex.Assert(args[1] == "-sdk", "the second argument must be -sdk")
-		return filepath.Join("Developer", "SDKs", args[2], "bin", args[3])
+		return string(filepath.Separator) + filepath.Join("", "Developer", "SDKs", args[2], "bin", args[3])
```
bassosimone, line 261:
```suggestion
		return string(filepath.Separator) + filepath.Join("Developer", "SDKs", args[2], "bin", args[3])
```

PR #1376, opened 2023-10-20
refactor(dslx): allow running w/o measurexlite as the backend

bassosimone:
Summary of the changes:

* rename `ConnPool` (`struct`) to `Runtime` (`interface`) and retain the same functionality;

* introduce a "minimal" runtime that does not collect observations along with a measurexlite runtime;

* use the "runtime" to carry the logger and other bits of information that it seems more rational to keep into the runtime rather than keeping inside each pipeline stage result (I also did this in the richer-input PoC);

* make sure `measurexlite`'s `Trace` can be fully accessed through an interface.

Closes https://github.com/ooni/probe/issues/2580


internal/dslx/tls.go
```diff
@@ -76,12 +74,12 @@ type tlsHandshakeFunc struct {
 	// NextProto contains the ALPNs to negotiate.
 	NextProto []string
 
-	// Pool is the Pool that owns us.
-	Pool *ConnPool
-
 	// RootCAs contains the Root CAs to use.
 	RootCAs *x509.CertPool
 
+	// Pool is the Pool that owns us.
```
bassosimone, line 80:
```suggestion
	// Rt is the Runtime that owns us.
```

bassosimone:
Something seems to have changed in the OONI `-test` backend, which causes `alltests` to fail.

I will investigate at a later time, since it may just be a transient issue.

PR #1377, opened 2023-10-20
cleanup(dslx): use model.UnderlyingNetwork for testing

bassosimone:
Closes https://github.com/ooni/probe/issues/2582


bassosimone:
Merging with red `alltests`, which is fine given that we're not close to a release and we're seen some backend issues that probably are transient and caused by ongoing `-test` deployment of which I am aware.

PR #1380, opened 2023-10-25
refactor(dslx): prepare for making HTTPRequest stateless

bassosimone:
- avoid storing state for creating an http.Request in httpRequestFunc

- avoid using methods for constructing an http.Request where we can use pure functions instead

- update and adapt the unit test suite

Reference issue: https://github.com/ooni/probe/issues/2612

### DIFF
--- a/internal/dslx/http_test.go
+++ b/internal/dslx/http_test.go
@@ -251,7 +251,7 @@ func TestHTTPRequest(t *testing.T) {
 			}
 		})
 
-		t.Run("with invalid URL path", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/dslx/http_test.go
```diff
@@ -115,20 +251,20 @@ func TestHTTPRequest(t *testing.T) {
 			}
 		})
 
-		t.Run("with invalid method", func(t *testing.T) {
+		t.Run("with invalid URL path", func(t *testing.T) {
```
bassosimone, line 254:
```suggestion
		t.Run("with invalid domain", func(t *testing.T) {
```

PR #1392, opened 2023-11-22
feat: make sure LTE and v0.4 emit the same test keys

bassosimone:
This diff switches LTE to use the classic "engine", removes restrictions regarding comparing test keys, and ensures with the existing test suite that LTE and v0.4 produce the same top-level test keys under the same conditions.

While there, let's also bump LTE's version number, lest we forget.

See https://github.com/ooni/probe/issues/2634


internal/experiment/webconnectivityqa/badssl.go
```diff
@@ -43,8 +43,8 @@ func badSSLWithWrongServerName() *TestCase {
 			HTTPExperimentFailure: "ssl_invalid_hostname",
 			XStatus:               16, // StatusAnomalyControlFailure
 			XNullNullFlags:        4,  // analysisFlagNullNullTLSMisconfigured
-			Accessible:            nil,
-			Blocking:              nil,
+			Accessible:            false,
+			Blocking:              false,
```
bassosimone, line 47: This is an improvement rather than a regression. Web Connectivity LTE correctly flags the website as being down because of misconfigured TLS, while Web Connectivity v0.4 is not able to do that.

internal/experiment/webconnectivityqa/badssl.go
```diff
@@ -64,8 +64,8 @@ func badSSLWithUnknownAuthorityWithConsistentDNS() *TestCase {
 			HTTPExperimentFailure: "ssl_unknown_authority",
 			XStatus:               16, // StatusAnomalyControlFailure
 			XNullNullFlags:        4,  // analysisFlagNullNullTLSMisconfigured
-			Accessible:            nil,
-			Blocking:              nil,
+			Accessible:            false,
+			Blocking:              false,
```
bassosimone, line 68: This is an improvement rather than a regression. Web Connectivity LTE correctly flags the website as being down because of misconfigured TLS, while Web Connectivity v0.4 is not able to do that.

internal/experiment/webconnectivityqa/badssl.go
```diff
@@ -89,6 +89,10 @@ func badSSLWithUnknownAuthorityWithInconsistentDNS() *TestCase {
 		},
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
+			BodyLengthMatch:       true,
+			StatusCodeMatch:       true,
+			HeadersMatch:          true,
+			TitleMatch:            true,
```
bassosimone, line 95: This is not a regression. Web Connectivity LTE includes code to periodically and opportunistically perform DNS-over-HTTPS (DoH) measurements. The first run in a list of URLs always uses DoH. After the first DoH attempt, we do not try again for a given number of seconds. Because the QA suite always runs with the first run, we are getting and using a valid IP address for the domain and hence we're able to fetch the right webpage.

Note that, even without DoH, if we can communicate with the test helper, we're getting legit IP address that we are going to use. Therefore, this test case would _also_ fetch the correct web page w/o DoH.

internal/experiment/webconnectivityqa/control.go
```diff
@@ -44,6 +44,7 @@ func controlFailureWithSuccessfulHTTPWebsite() *TestCase {
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
 			ControlFailure: "unknown_failure: httpapi: all endpoints failed: [ connection_reset; connection_reset; connection_reset; connection_reset;]",
+			DNSConsistency: "consistent",
```
bassosimone, line 47: This is not a regression. Web Connectivity LTE performs TLS connections with the website even when the URL is HTTP to check whether we can consider the IP addresses as valid. Because TLS succeeds, we end up considering the DNS as consistent.

internal/experiment/webconnectivityqa/control.go
```diff
@@ -89,8 +90,9 @@ func controlFailureWithSuccessfulHTTPSWebsite() *TestCase {
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
 			ControlFailure: "unknown_failure: httpapi: all endpoints failed: [ connection_reset; connection_reset; connection_reset; connection_reset;]",
-			XStatus:        1, // StatusSuccessSecure
-			XNullNullFlags: 8, // analysisFlagNullNullSuccessfulHTTPS
+			DNSConsistency: "consistent",
```
bassosimone, line 93: We conclude that the DNS is consistent because TLS works. This is not a regression, rather a way in which Web Connectivity LTE is outperforming v0.4.

internal/experiment/webconnectivityqa/dnsblocking.go
```diff
@@ -47,12 +51,15 @@ func dnsBlockingNXDOMAIN() *TestCase {
...
-			// converted into android_dns_cache_no_data by the emulation layer
+			// remove the record so that the DNS query returns NXDOMAIN
 			env.ISPResolverConfig().RemoveRecord("www.example.com")
 		},
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
+			BodyLengthMatch:      true,
+			StatusCodeMatch:      true,
+			HeadersMatch:         true,
+			TitleMatch:           true,
```
bassosimone, line 62: This is not a regression. LTE is able to fetch a webpage because it uses multiple resolvers.

internal/experiment/webconnectivityqa/tcpblocking.go
```diff
@@ -63,6 +63,10 @@ func tcpBlockingConnectionRefusedWithInconsistentDNS() *TestCase {
 		},
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
+			BodyLengthMatch:       true,
+			StatusCodeMatch:       true,
+			HeadersMatch:          true,
+			TitleMatch:            true,
```
bassosimone, line 69: This is not a regression. LTE tries harder and is able to bypass the blocking. But it still records the existing blocking mechanisms. So, we are actually doing a better job here.

internal/experiment/webconnectivityqa/redirect.go
```diff
@@ -37,7 +37,7 @@ func redirectWithConsistentDNSAndThenConnectionRefusedForHTTP() *TestCase {
 			HTTPExperimentFailure: "connection_refused",
 			XStatus:               8320, // StatusExperimentHTTP | StatusAnomalyConnect
 			XDNSFlags:             0,
-			XBlockingFlags:        32, // analysisFlagSuccess
+			XBlockingFlags:        8, // analysisFlagHTTPBlocking
```
bassosimone, line 40: This used to be a regression. Now LTE yields the same result of v0.4.

internal/experiment/webconnectivityqa/redirect.go
```diff
@@ -125,7 +125,7 @@ func redirectWithConsistentDNSAndThenConnectionResetForHTTP() *TestCase {
 func redirectWithConsistentDNSAndThenConnectionResetForHTTPS() *TestCase {
 	return &TestCase{
 		Name:  "redirectWithConsistentDNSAndThenConnectionResetForHTTPS",
-		Flags: TestCaseFlagNoLTE, // BUG: LTE thinks this website is accessible (WTF?!)
+		Flags: 0,
```
bassosimone, line 128: This used to be a regression. Now LTE yields the same result of v0.4.

internal/experiment/webconnectivityqa/redirect.go
```diff
@@ -258,7 +258,7 @@ func redirectWithConsistentDNSAndThenEOFForHTTPS() *TestCase {
 			HTTPExperimentFailure: "eof_error",
 			XStatus:               8448, // StatusExperimentHTTP | StatusAnomalyReadWrite
 			XDNSFlags:             0,
-			XBlockingFlags:        32, // analysisFlagSuccess
+			XBlockingFlags:        8, // analysisFlagHTTPBlocking
```
bassosimone, line 261: This used to be a regression. Now LTE yields the same result of v0.4.

internal/experiment/webconnectivityqa/redirect.go
```diff
@@ -336,7 +336,7 @@ func redirectWithConsistentDNSAndThenTimeoutForHTTPS() *TestCase {
 			HTTPExperimentFailure: "generic_timeout_error",
 			XStatus:               8704, // StatusExperimentHTTP | StatusAnomalyUnknown
 			XDNSFlags:             0,
-			XBlockingFlags:        32, // analysisFlagSuccess
+			XBlockingFlags:        8, // analysisFlagHTTPBlocking
```
bassosimone, line 339: This used to be a regression. Now LTE yields the same result of v0.4.

internal/experiment/webconnectivityqa/httpdiff.go
```diff
@@ -92,7 +92,7 @@ func httpDiffWithInconsistentDNS() *TestCase {
 			TitleMatch:            false,
 			XStatus:               96, // StatusAnomalyHTTPDiff | StatusAnomalyDNS
 			XDNSFlags:             4,  // AnalysisDNSUnexpectedAddrs
-			XBlockingFlags:        35, // analysisFlagSuccess | analysisFlagDNSBlocking | analysisFlagTCPIPBlocking
+			XBlockingFlags:        17, // analysisFlagDNSBlocking | analysisFlagHTTPDiff
```
bassosimone, line 95: This seems to be a regression that we have fixed. The correctly expected result is to see that there is DNS blocking (because we're spoofing the IP address) and HTTP diff (because we're injecting a response).

It should also be noted that this test case seems flaky when running it locally. Because there is a race condition involved in spoofing the webpage, we would probably need to increase the RTTs to make sure that the censor always wins the race with ordinary packets in flight.

internal/experiment/webconnectivityqa/dnshijacking.go
```diff
@@ -64,7 +64,7 @@ func dnsHijackingToProxyWithHTTPSURL() *TestCase {
 		},
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
-			DNSConsistency:  "inconsistent",
+			DNSConsistency:  "consistent",
```
bassosimone, line 67: Here the DNS flips from inconsistent to consistent because we can perform a TLS handshake for the target domain to the spoofed IP address, so we say that the IP address is good for the domain.

internal/experiment/webconnectivityqa/dnshijacking.go
```diff
@@ -33,11 +33,11 @@ func dnsHijackingToProxyWithHTTPURL() *TestCase {
...
-			XStatus:         2, // StatusSuccessCleartext
-			XDNSFlags:       0,
-			XBlockingFlags:  32, // analysisFlagSuccess
-			Accessible:      true,
-			Blocking:        false,
+			XStatus:         2,     // StatusSuccessCleartext
+			XDNSFlags:       4,     // AnalysisDNSUnexpectedAddrs
+			XBlockingFlags:  33,    // AnalysisFlagDNSBlocking | analysisFlagSuccess
+			Accessible:      false, // FIXME: this is probably incorrect!
+			Blocking:        "dns",
```
bassosimone, line 40: This case seems tricky. We can see the result oscillating. In some cases, we get the expectations that are listed above, which changed since `master`, and which we'll call case `A`. In some other cases, we see the following error, which we'll call case `B`:

```
qa_test.go:20: test keys mismatch:   &webconnectivityqa.testKeys{
          	... // 1 ignored field
          	DNSExperimentFailure: nil,
        - 	DNSConsistency:       string("inconsistent"),
        + 	DNSConsistency:       string("consistent"),
          	ControlFailure:       nil,
          	... // 3 ignored and 4 identical fields
        - 	XDNSFlags:      4,
        + 	XDNSFlags:      0,
        - 	XBlockingFlags: 33,
        + 	XBlockingFlags: 32,
          	XNullNullFlags: 0,
        - 	Accessible:     bool(false),
        + 	Accessible:     bool(true),
        - 	Blocking:       string("dns"),
        + 	Blocking:       bool(false),
          }
```

In the former case, the analysis is this:

```JSON
{
  "DNSExperimentFailure": null,
  "DNSTransactionsWithBogons": {},
  "DNSTransactionsWithUnexpectedFailures": {},
  "DNSPossiblyInvalidAddrs": {
    "130.192.182.17": true
  },
  "HTTPDiffBodyProportionFactor": 1,
  "HTTPDiffStatusCodeMatch": true,
  "HTTPDiffTitleDifferentLongWords": {},
  "HTTPDiffUncommonHeadersIntersection": {
    "content-length": true
  },
  "HTTPFinalResponses": {
    "4": true
  },
  "HTTPFinalResponsesWithTLS": {},
  "TCPTransactionsWithUnexpectedTCPConnectFailures": {},
  "TCPTransactionsWithUnexpectedTLSHandshakeFailures": {},
  "TCPTransactionsWithUnexpectedHTTPFailures": {},
  "TCPTransactionsWithUnexplainedUnexpectedFailures": {}
}
```

In the latter case, the analysis is:

```JSON
{
  "DNSExperimentFailure": null,
  "DNSTransactionsWithBogons": {},
  "DNSTransactionsWithUnexpectedFailures": {},
  "DNSPossiblyInvalidAddrs": {},
  "HTTPDiffBodyProportionFactor": 1,
  "HTTPDiffStatusCodeMatch": true,
  "HTTPDiffTitleDifferentLongWords": {},
  "HTTPDiffUncommonHeadersIntersection": {
    "content-length": true
  },
  "HTTPFinalResponses": {
    "4": true
  },
  "HTTPFinalResponsesWithTLS": {},
  "TCPTransactionsWithUnexpectedTCPConnectFailures": {},
  "TCPTransactionsWithUnexpectedTLSHandshakeFailures": {},
  "TCPTransactionsWithUnexpectedHTTPFailures": {},
  "TCPTransactionsWithUnexplainedUnexpectedFailures": {}
}
```

The difference between the two cases is:

```diff
--- A	2023-11-27 14:01:02
+++ B	2023-11-27 14:00:48
@@ -2,9 +2,7 @@
   "DNSExperimentFailure": null,
   "DNSTransactionsWithBogons": {},
   "DNSTransactionsWithUnexpectedFailures": {},
-  "DNSPossiblyInvalidAddrs": {
-    "130.192.182.17": true
-  },
+  "DNSPossiblyInvalidAddrs": {},
   "HTTPDiffBodyProportionFactor": 1,
   "HTTPDiffStatusCodeMatch": true,
   "HTTPDiffTitleDifferentLongWords": {},
```

The difference between the observations is the following:

```diff
@@ -40,12 +40,12 @@
       "ControlHTTPResponseHeadersKeys": null,
       "ControlHTTPResponseTitle": null
     },
-    "2": {
+    "3": {
       "DNSTransactionIDs": [
-        2
+        3
       ],
       "DNSDomain": "www.example.com",
-      "DNSLookupFailure": "dns_no_answer",
+      "DNSLookupFailure": "dns_nxdomain_error",
       "DNSQueryType": "AAAA",
       "DNSEngine": "doh",
       "IPAddress": null,
@@ -84,7 +84,7 @@
   "KnownTCPEndpoints": {
     "4": {
       "DNSTransactionIDs": [
-        3,
+        2,
         1
       ],
       "DNSDomain": "www.example.com",
@@ -134,23 +134,24 @@
     },
     "5": {
       "DNSTransactionIDs": [
-        2
+        2,
+        1
       ],
       "DNSDomain": "www.example.com",
       "DNSLookupFailure": "",
       "DNSQueryType": null,
       "DNSEngine": null,
-      "IPAddress": "93.184.216.34",
-      "IPAddressASN": 15133,
-      "IPAddressOrg": "Edgecast Inc.",
+      "IPAddress": "130.192.182.17",
+      "IPAddressASN": 137,
+      "IPAddressOrg": "Consortium GARR",
       "IPAddressBogon": false,
       "EndpointTransactionID": 5,
       "EndpointProto": "tcp",
-      "EndpointPort": "80",
-      "EndpointAddress": "93.184.216.34:80",
+      "EndpointPort": "443",
+      "EndpointAddress": "130.192.182.17:443",
       "TCPConnectFailure": "",
-      "TLSHandshakeFailure": null,
-      "TLSServerName": null,
+      "TLSHandshakeFailure": "",
+      "TLSServerName": "www.example.com",
       "HTTPRequestURL": null,
       "HTTPFailure": null,
       "HTTPResponseStatusCode": null,
@@ -163,9 +164,9 @@
       "ControlDNSDomain": null,
       "ControlDNSLookupFailure": null,
       "ControlTCPConnectFailure": "",
-      "MatchWithControlIPAddress": true,
-      "MatchWithControlIPAddressASN": true,
-      "ControlTLSHandshakeFailure": null,
+      "MatchWithControlIPAddress": false,
+      "MatchWithControlIPAddressASN": false,
+      "ControlTLSHandshakeFailure": "",
       "ControlHTTPFailure": "",
       "ControlHTTPResponseStatusCode": 200,
       "ControlHTTPResponseBodyLength": 1533,
@@ -178,25 +179,22 @@
       "ControlHTTPResponseTitle": "Default Web Page"
     },
     "6": {
-      "DNSTransactionIDs": [
-        3,
-        1
-      ],
-      "DNSDomain": "www.example.com",
-      "DNSLookupFailure": "",
+      "DNSTransactionIDs": [],
+      "DNSDomain": null,
+      "DNSLookupFailure": null,
       "DNSQueryType": null,
       "DNSEngine": null,
-      "IPAddress": "130.192.182.17",
-      "IPAddressASN": 137,
-      "IPAddressOrg": "Consortium GARR",
+      "IPAddress": "93.184.216.34",
+      "IPAddressASN": 15133,
+      "IPAddressOrg": "Edgecast Inc.",
       "IPAddressBogon": false,
       "EndpointTransactionID": 6,
       "EndpointProto": "tcp",
-      "EndpointPort": "443",
-      "EndpointAddress": "130.192.182.17:443",
+      "EndpointPort": "80",
+      "EndpointAddress": "93.184.216.34:80",
       "TCPConnectFailure": "",
-      "TLSHandshakeFailure": "",
-      "TLSServerName": "www.example.com",
+      "TLSHandshakeFailure": null,
+      "TLSServerName": null,
       "HTTPRequestURL": null,
       "HTTPFailure": null,
       "HTTPResponseStatusCode": null,
@@ -209,9 +207,9 @@
       "ControlDNSDomain": null,
       "ControlDNSLookupFailure": null,
       "ControlTCPConnectFailure": "",
-      "MatchWithControlIPAddress": false,
-      "MatchWithControlIPAddressASN": false,
-      "ControlTLSHandshakeFailure": "",
+      "MatchWithControlIPAddress": null,
+      "MatchWithControlIPAddressASN": null,
+      "ControlTLSHandshakeFailure": null,
       "ControlHTTPFailure": "",
       "ControlHTTPResponseStatusCode": 200,
       "ControlHTTPResponseBodyLength": 1533,
@@ -224,11 +222,9 @@
       "ControlHTTPResponseTitle": "Default Web Page"
     },
     "7": {
-      "DNSTransactionIDs": [
-        2
-      ],
-      "DNSDomain": "www.example.com",
-      "DNSLookupFailure": "",
+      "DNSTransactionIDs": [],
+      "DNSDomain": null,
+      "DNSLookupFailure": null,
       "DNSQueryType": null,
       "DNSEngine": null,
       "IPAddress": "93.184.216.34",
@@ -254,8 +250,8 @@
       "ControlDNSDomain": null,
       "ControlDNSLookupFailure": null,
       "ControlTCPConnectFailure": "",
-      "MatchWithControlIPAddress": true,
-      "MatchWithControlIPAddressASN": true,
+      "MatchWithControlIPAddress": null,
+      "MatchWithControlIPAddressASN": null,
       "ControlTLSHandshakeFailure": "",
       "ControlHTTPFailure": "",
       "ControlHTTPResponseStatusCode": 200,
```

bassosimone: Interestingly, both runs seem to produce the same observations (as far as I can tell). So, either I am missing something or the algorithm producing the results must contain something subtle that depends on ordering.

bassosimone: Fixed by https://github.com/ooni/probe-cli/commit/8d25a65d8eb13f4b457e3b29b8dd38f040ec4f74

internal/experiment/webconnectivityqa/dnshijacking.go
```diff
@@ -27,7 +27,7 @@ func dnsHijackingToProxyWithHTTPURL() *TestCase {
 		},
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
-			DNSConsistency:  "inconsistent",
+			DNSConsistency:  "consistent",
```
bassosimone, line 30: Here the DNS flips from inconsistent to consistent because we can perform a TLS handshake for the target domain to the spoofed IP address, so we say that the IP address is good for the domain.

internal/experiment/webconnectivityqa/httpdiff.go
```diff
@@ -83,18 +83,18 @@ func httpDiffWithInconsistentDNS() *TestCase {
 		ExpectErr: false,
 		ExpectTestKeys: &testKeys{
 			DNSExperimentFailure:  nil,
-			DNSConsistency:        "inconsistent",
+			DNSConsistency:        "consistent",
```
bassosimone, line 86: Here the DNS flips with consistent because the proxy we're redirecting to supports TLS.

bassosimone: **TODO** split this test case into two test cases

PR #1402, opened 2023-11-29
feat(minipipeline): introduce classic filtering

bassosimone:
Classic filtering is an `WebObservationsContainer` filtering technique that takes in input a `WebObservationsContainer` and only keeps DNS lookups using `getaddrinfo` and endpoints whose IP address has been discovered using `getaddrinfo`. By applying this technique, we reduce the richer dataset produced by Web Connectivity LTE to a smaller dataset comparable to what Web Connectivity v0.4 would return. In turn, by focusing the analysis on the reduced dataset, we hope to emulate the results produced by v0.4 for backward compatible test keys.

Part of https://github.com/ooni/probe/issues/2634.


internal/cmd/qatool/main.go
```diff
@@ -79,14 +79,23 @@ func runWebConnectivityLTE(tc *webconnectivityqa.TestCase) {
 		// serialize the observations
 		mustSerializeMkdirAllAndWriteFile(actualDestdir, "observations.json", observationsContainer)
 
+		// convert to classic analysis
```
bassosimone, line 82:
```suggestion
		// convert to classic observations
```

PR #1424, opened 2023-12-12
feat: allow go1.21 builds

bassosimone:
We can ~safely compile our net/http and crypto/tls forks with a later version of Go, because they use public packages in the standard library, which should be covered by the Go 1.x stability guarantee. Whether that is 100% safe, I don't know, but I suppose we should be fine unless there's really something subtle. (We will continue building our packages with the correct version of Go, but I also understand how distribution maintainers have other needs.)

The real price to pay for compiling with go1.21 is disabling Psiphon. There isn't much we can do here. It's not possible to have a build configuration with fixed flags that disable GQUIC for Psiphon, so I have two choices (a) either ooniprobe does not build for go1.21 because of Psiphon and I need to tell people to apply a flag; or (b) when you compile with go1.21, everything is WAI but for Psiphon. I chose the latter in this pull request.

Note that I originally tried to make oohttp and oocrypto optional. However, it turns out making oohttp optional and using net/http instead cripples ooniprobe entirely. This is why I concluded that maybe compiling go1.20 code using the go1.21 compiler and stdlib is not that bad, considering that our stdlib forks do not (obviously) depend on internals.

Part of https://github.com/ooni/probe/issues/2556; closes https://github.com/ooni/probe/issues/2548.


bassosimone:
So, the code up until bd0129622d9dae7d01346cf9e6900f02970e7251 is such that OONI Probe compiles using Go 1.21, however, unfortunately, the generated binary does not work. The reason why it does not work is exemplified by the following failing log message:

```
[      0.725700] <info> sessionresolver: lookup dkyhjv0wpi2dk.cloudfront.net using https://dns.google/dns-query... started
[      1.227024] <info> sessionresolver: lookup dkyhjv0wpi2dk.cloudfront.net using https://dns.google/dns-query... in progress
[      1.935738] <info> sessionresolver: lookup dkyhjv0wpi2dk.cloudfront.net using https://dns.google/dns-query... unknown_failure: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x12\x04\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00d\x00\x04\x00\x10\x00\x00\x00\x06\x00\x01\x00\x00\x00\x00\x04\b\x00\x00\x00\x00\x00\x00\x0f\x00\x01\x00\x00\x1e\a\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01http2_handshake_failed"
```

The TL;DR is that we're using `["h2", "http/1.1"]` inside the TLS handshake. However, the Go TLS stack is not prepared to handle `h2`, therefore we end up failing. Now, one could argue that I can stop sending "h2" as part of the handshake, but this would actually be quite bad because it would further change the TLS handshake.

This means I've reached a bit of a stalemate. The current code cannot fundamentally work. 🤔

bassosimone:
Okay, I've realized I can ~safely use Go 1.20 forked libraries with Go 1.21. This should mostly be fine because of the Go 1.x compatibility guarantee. This means we only need to featurize Psiphon. I'm going to update the original PR text.

PR #1434, opened 2023-12-13
chore: update C deps

bassosimone:
Part of https://github.com/ooni/probe/issues/2556


bassosimone:
We also need to introduce a workaround for https://github.com/mattn/go-sqlite3/issues/1164 😓

PR #1423, opened 2023-12-12
chore: publish code used during the splintercon demo

bassosimone:
During [the 2023 SplinterCon edition](https://splintercon.net/), I demoed using OONI technologies for research, ranging from invoking `miniooni` to using OONI Run v2, to using scripts to iterate faster. This branch contains all the code that I used during the demo, consisting mainly of fixes for existing code and of a preview of the JavaScript engine.

I am going to create separate pull requests for each functionality and then close this PR. However, my intent in creating this PR is a container to illustrate the work I did preparing for SplinterCon. In turn, most of the JavaScript code here evolved from code that I wrote researching richer input and later refactored to better support using JavaScript.

PR #1435, opened 2023-12-13
fix: make sure we can parse integers from JSON

bassosimone:
This commit ensures we can parse integer values provided using OONI Run v2 descriptors into actual integers.

Values are originally parsed as float64, so we need to add a specific conversion case for that scenario.

Diff extracted from: https://github.com/ooni/probe-cli/pull/1423

Closes https://github.com/ooni/probe/issues/2645

### DIFF
--- a/internal/registry/factory.go
+++ b/internal/registry/factory.go
@@ -7,6 +7,7 @@ package registry
 import (
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"reflect"
 	"strconv"
@@ -107,6 +108,12 @@ func (b *Factory) setOptionBool(field reflect.Value, value any) error {
 	}
 }
 
+// With JSON we're limited by the 52 bits in the mantissa
+const (
+	jsonMaxInteger = 1<<53 - 1
+	jsonMinInteger = -1<<53 + 1
+)
+
 // setOptionInt sets an int option
 func (b *Factory) setOptionInt(field reflect.Value, value any) error {
 	switch v := value.(type) {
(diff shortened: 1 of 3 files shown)


internal/registry/factory.go
```diff
@@ -132,6 +139,18 @@ func (b *Factory) setOptionInt(field reflect.Value, value any) error {
 		}
 		field.SetInt(number)
 		return nil
+	case float64:
+		if math.IsNaN(v) || math.IsInf(v, 0) {
+			return fmt.Errorf("%w from NaN, +-Inf: %v", ErrCannotSetIntegerOption, value)
```
bassosimone, line 144:
```suggestion
			return fmt.Errorf("%w from: %v", ErrCannotSetIntegerOption, value)
```

internal/registry/factory.go
```diff
@@ -132,6 +139,18 @@ func (b *Factory) setOptionInt(field reflect.Value, value any) error {
...
+		}
+		if math.Trunc(v) != v {
+			return fmt.Errorf("%w from a value of type %T", ErrCannotSetIntegerOption, value)
```
bassosimone, line 147:
```suggestion
			return fmt.Errorf("%w from: %v", ErrCannotSetIntegerOption, value)
```

internal/registry/factory.go
```diff
@@ -132,6 +139,18 @@ func (b *Factory) setOptionInt(field reflect.Value, value any) error {
...
+		}
+		if v > jsonMaxInteger || v < jsonMinInteger {
+			return fmt.Errorf("%w from a too large or too small float64: %v", ErrCannotSetIntegerOption, value)
```
bassosimone, line 150:
```suggestion
			return fmt.Errorf("%w from: %v", ErrCannotSetIntegerOption, value)
```

PR #1450, opened 2024-01-16
chore(minipipeline): increase code coverage

bassosimone:
I spent some time increasing code coverage for `internal/minipipeline`.

This is KTLO work as part of https://github.com/ooni/probe/issues/2640.

### DIFF
--- a/internal/minipipeline/observation_test.go
+++ b/internal/minipipeline/observation_test.go
@@ -156,7 +156,7 @@ func TestWebObservationsContainerIngestControlMessages(t *testing.T) {
 		}
 	})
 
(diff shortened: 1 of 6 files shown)


internal/minipipeline/observation_test.go
```diff
@@ -154,4 +155,52 @@ func TestWebObservationsContainerIngestControlMessages(t *testing.T) {
 			t.Fatal("ControlTLSHandshakeFailure should be none")
 		}
 	})
+
+	t.Run("we correctly handle an inconsisten control DNS lookup result", func(t *testing.T) {
```
bassosimone, line 159:
```suggestion
	t.Run("we correctly handle an inconsistent control DNS lookup result", func(t *testing.T) {
```

PR #1453, opened 2024-01-19
feat(minipipeline): add DNS and HTTP unexplained failures

bassosimone:
We need to know about all possible unexplained failures. With this information and information about ControlFinalResponseExpectations introduced in #1451 and #1452, we are able to transform unexplained failures (i.e., failures occurring during redirects) into explained failures when we have a control-based expectation to compare to.

Part of https://github.com/ooni/probe/issues/2640


internal/minipipeline/analysis.go
```diff
@@ -377,6 +382,16 @@ func (wa *WebAnalysis) dnsComputeFailureMetrics(c *WebObservationsContainer) {
...
 
+		// lookups once we started following redirects does not have corresponding
+		// control information, so failures end up being unexplained
+		if obs.TagDepth.IsNone() || obs.TagDepth.Unwrap() != 0 {
+			if obs.DNSLookupFailure.Unwrap() != "" {
+				wa.DNSLookupUnexplainedFailure.Add(obs.DNSTransactionID.Unwrap())
+				continue
+			}
+			continue
+		}
```
bassosimone, line 393: Moving this block of code down here should be fine, given that we're just moving checks that cause us to skip processing above and we end up skipping the entry with `continue` in any case in the `if` block.

PR #1457, opened 2024-01-22
feat(webconnectivitylte): handle ghost DNS censorship

bassosimone:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2652
- [x] if you changed anything related to how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: no need
- [x] if you changed code inside an experiment, make sure you bump its version number: already bumped for 3.21.x

## Description

This commit modifies webconnectivitylte to handle ghost DNS censorship.

We define ghost DNS censorship the case where the original domain does not exist anymore but the censor continues to return an IP address for the original domain nonetheless.

We used to have null-null handling for this case in the "orig" engine and a reference issue as https://github.com/ooni/probe/issues/2307.

With this commit, we modify the "classic" engine to correctly handle this case.

To this end, we need to:

1. mark DNS inconsistency when we have successful probe lookups and no IP address has been resolved by the test helper, which is indeed the case of ghost DNS censorship.

2. specific handling on the case in which the website seems down where we also ask ourselves the question of whether the culprit could be the DNS and otherwise set accessible = false and blocking = false.

Note that, with those two changes, we depart from strict v0.4-is-always-right orthodoxy.

So, while there, let's recognize that always setting HTTPExprimentFailure is probably for the greater good.

### DIFF
--- a/internal/experiment/webconnectivityqa/ghost.go
+++ b/internal/experiment/webconnectivityqa/ghost.go
@@ -44,7 +44,7 @@ func ghostDNSBlockingWithHTTP() *TestCase {
 }
 
 // ghostDNSBlockingWithHTTPS is the case where the domain does not exist anymore but
-// there's still ghost censorship because of the system resolver configuration, which
+// there's still ghost censorship because of the censor DNS censoring configuration, which
 // says that we should censor the domain by returning a specific IP address.
 //
 // See https://github.com/ooni/probe/issues/2308.
--- a/internal/experiment/webconnectivityqa/ghost.go
+++ b/internal/experiment/webconnectivityqa/ghost.go
@@ -1,0 +1,81 @@
+package webconnectivityqa
+
+import (
+	"github.com/apex/log"
+	"github.com/ooni/netem"
+	"github.com/ooni/probe-cli/v3/internal/netemx"
+)
+
+// ghostDNSBlockingWithHTTP is the case where the domain does not exist anymore but
+// there's still ghost censorship because of the system resolver configuration, which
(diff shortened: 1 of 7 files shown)


internal/experiment/webconnectivityqa/ghost.go
```diff
@@ -0,0 +1,81 @@
+package webconnectivityqa
+
+import (
+	"github.com/apex/log"
+	"github.com/ooni/netem"
+	"github.com/ooni/probe-cli/v3/internal/netemx"
+)
+
+// ghostDNSBlockingWithHTTP is the case where the domain does not exist anymore but
+// there's still ghost censorship because of the system resolver configuration, which
```
bassosimone, line 10:
```suggestion
// there's still ghost censorship because of the censor DNS censoring configuration, which
```

internal/experiment/webconnectivityqa/ghost.go
```diff
@@ -0,0 +1,81 @@
...
+			XNullNullFlags:       16, // AnalysisFlagNullNullUnexpectedDNSLookupSuccess
+			XStatus:              16, // StatusAnomalyControlFailure
+			Accessible:           false,
+			Blocking:             "dns",
+		},
+	}
+}
+
+// ghostDNSBlockingWithHTTPS is the case where the domain does not exist anymore but
+// there's still ghost censorship because of the system resolver configuration, which
```
bassosimone, line 47:
```suggestion
// there's still ghost censorship because of the censor DNS censoring configuration, which
```

PR #1459, opened 2024-01-22
minipipeline: fix expected TCP & TLS failures

bassosimone:
This diff corrects an embarrassing bugs in the logic we were using for expected TCP and TLS failures.

We spotted this issue with the new testcase we added called websiteDownTCPConnect, which is also included.

Additionally, we updated some tests because of the embarrassing bugs we fixed.

Additionally, we fixed some comments that were outdated.

Part of https://github.com/ooni/probe/issues/2652.

Closes https://github.com/ooni/probe/issues/2299.


internal/minipipeline/analysis.go
```diff
@@ -555,18 +552,15 @@ func (wa *WebAnalysis) tlsComputeMetrics(c *WebObservationsContainer) {
 		}
 
 		// handle the case where both the probe and the control fail
-		if obs.TLSHandshakeFailure.Unwrap() != "" && obs.ControlTCPConnectFailure.Unwrap() != "" {
+		//
+		// See https://github.com/ooni/probe/issues/2300.
+		if obs.TLSHandshakeFailure.Unwrap() != "" && obs.ControlTLSHandshakeFailure.Unwrap() != "" {
```
bassosimone, line 557: Bug `#1` is that we were checking against `ControlTCPConnectFailure`, which was wrong.

internal/minipipeline/analysis.go
```diff
@@ -555,18 +552,15 @@ func (wa *WebAnalysis) tlsComputeMetrics(c *WebObservationsContainer) {
...
 		}
 
 		// handle the case where the control fails
 		if obs.ControlTLSHandshakeFailure.Unwrap() != "" {
-			// If also the probe failed mark this failure as expected.
-			//
-			// See https://github.com/ooni/probe/issues/2300.
-			if obs.TLSHandshakeFailure.Unwrap() != "" {
-				wa.TLSHandshakeExpectedFailure.Add(obs.EndpointTransactionID.Unwrap())
-			}
```
bassosimone, line 569: Bug `#2` is that this code should have been in the `if` above.

internal/minipipeline/analysis.go
```diff
@@ -491,21 +491,18 @@ func (wa *WebAnalysis) tcpComputeMetrics(c *WebObservationsContainer) {
...
 		if obs.ControlTCPConnectFailure.Unwrap() != "" {
-			// If also the probe failed mark this failure as expected.
-			//
-			// See https://explorer.ooni.org/measurement/20220911T105037Z_webconnectivity_IT_30722_n1_ruzuQ219SmIO9SrT?input=https://doh.centraleu.pi-dns.com/dns-query?dns=q80BAAABAAAAAAAAA3d3dwdleGFtcGxlA2NvbQAAAQAB
-			// for an example measurement with this behavior.
-			//
-			// See also https://github.com/ooni/probe/issues/2299.
-			if obs.TCPConnectFailure.Unwrap() != "" {
-				wa.TCPConnectExpectedFailure.Add(obs.EndpointTransactionID.Unwrap())
-			}
```
bassosimone, line 508: Bug `#3` is that this code should have been in the `if` above.

PR #1471, opened 2024-01-24
chore(minipipeline): add data to understand ooni/probe#1511

bassosimone:
This commit adds data collected from https://github.com/ooni/probe/issues/1511#issuecomment-1908208226 that helps us to understand why `minipipeline` flags the measurement as `accessible = null` and `blocking = null`.


internal/minipipeline/testdata/webconnectivity/manual/8844/analysis_classic.json
```diff
@@ -0,0 +1,42 @@
...
+  "HTTPRoundTripUnexplainedFailure": [],
+  "HTTPFinalResponseSuccessTLSWithoutControl": null,
+  "HTTPFinalResponseSuccessTLSWithControl": null,
+  "HTTPFinalResponseSuccessTCPWithoutControl": null,
+  "HTTPFinalResponseSuccessTCPWithControl": null,
+  "HTTPFinalResponseDiffBodyProportionFactor": null,
+  "HTTPFinalResponseDiffStatusCodeMatch": null,
+  "HTTPFinalResponseDiffTitleDifferentLongWords": null,
+  "HTTPFinalResponseDiffUncommonHeadersIntersection": null,
+  "Linear": null
```
bassosimone, line 41: That's the issue: `Linear` should not be `null` here. The problem is that we only consider IP addresses resolved by the probe when performing the classic linear analysis. When the IP address is part of the URL, this design choice means that we don't have any linear analysis when in classic mode.

internal/minipipeline/testdata/webconnectivity/manual/8844/observations_classic.json
```diff
@@ -0,0 +1,9 @@
+{
+  "DNSLookupFailures": [],
+  "DNSLookupSuccesses": [],
+  "KnownTCPEndpoints": {},
```
bassosimone, line 4: As pointed out above, here we should have endpoints. However, the classic filter removes all the IP addresses that are not resolved by the probe, including the one provided as part of the input URL.

PR #1472, opened 2024-01-24
fix(minipipeline): handle IP-addr URLs in classic linear analysis

bassosimone:
This diff fixes https://github.com/ooni/probe/issues/1511#issuecomment-1908208226.

This is the rationale of the diff: we need to track the origin of IP addresses:

* "dns" if discovered using DNS;
* "th" if discovered using the test helper;
* null otherwise.

When filtering for classic analysis, we include "dns" entries if resolved using getaddrinfo, drop "th" entries because they're not relevant, and include null entries under the assumption that the probe discovered them either directly from the input URL or because a redirect redirected to an URL containing an IP address.

We also update the minipipeline test suite and show that the only changes are related to the test added by https://github.com/ooni/probe-cli/pull/1471.


internal/minipipeline/testdata/webconnectivity/manual/8844/analysis_classic.json
```diff
@@ -31,12 +31,84 @@
...
-  "Linear": null
+  "HTTPFinalResponseDiffBodyProportionFactor": 1,
+  "HTTPFinalResponseDiffStatusCodeMatch": true,
+  "HTTPFinalResponseDiffTitleDifferentLongWords": {},
+  "HTTPFinalResponseDiffUncommonHeadersIntersection": {
+    "access-control-allow-origin": true,
+    "alt-svc": true,
+    "content-length": true
+  },
+  "Linear": [
```
bassosimone, line 45: We clearly fixed the bug!

internal/minipipeline/testdata/webconnectivity/manual/8844/observations_classic.json
```diff
@@ -1,7 +1,75 @@
 {
   "DNSLookupFailures": [],
   "DNSLookupSuccesses": [],
-  "KnownTCPEndpoints": {},
+  "KnownTCPEndpoints": {
```
bassosimone, line 4: We clearly fixed the bug!

PR #1473, opened 2024-01-24
fix(webconnectivitylte): handle domains w/o A/AAAA records

bassosimone:
This diff improves webconnectivitylte and minipipeline to correctly handle the case where a website has no configured IP addresses but the domain for the website exists. When this happens, the TH returns no error but an empty list of resolved addresses for historical reasons, while the probe fails with the `"dns_no_answer"` error. We need to intercept and correctly handle this scenario both for the ordinary analysis and the classic analysis.

Let's also add QA test cases to make sure we don't regress.

Part of https://github.com/ooni/probe/issues/2652.


internal/experiment/webconnectivitylte/analysisclassic.go
```diff
@@ -373,6 +373,15 @@ func analysisClassicComputeBlockingAccessible(woa *minipipeline.WebAnalysis, tk
...
+			// 5.3. special case for when the probe failed with dns_no_data and the
+			// test helper returned no addresses, which is its way to tell us the same
+			// error occurred but we need to interpret it as dns_no_data.
+			if !entry.ControlDNSResolvedAddrs.IsNone() &&
+				entry.ControlDNSResolvedAddrs.Unwrap().Len() <= 0 {
+				tk.setWebsiteDown()
+				return
+			}
+
 			// 5.3. Handle the case where just the probe failed.
```
bassosimone, line 385:
```suggestion
			// 5.4. Handle the case where just the probe failed.
```

internal/experiment/webconnectivitylte/analysisclassic.go
```diff
@@ -373,7 +373,16 @@ func analysisClassicComputeBlockingAccessible(woa *minipipeline.WebAnalysis, tk
 				return
 			}
 
-			// 5.3. Handle the case where just the probe failed.
+			// 5.3. special case for when the probe failed with dns_no_data and the
```
bassosimone, line 376:
```suggestion
			// 5.3. Special case for when the probe failed with dns_no_data and the
```

internal/experiment/webconnectivityqa/websitedown.go
```diff
@@ -60,3 +62,29 @@ func websiteDownTCPConnect() *TestCase {
...
 	}
 }
+
+// websiteDownNoAddrs describes the test case where the website domain
+// does not return any address according to the TH and the probe.
+func websiteDownNoAddrs() *TestCase {
+	return &TestCase{
+		Name:  "websiteDownNoAddrs",
+		Flags: TestCaseFlagNoV04,
+		Input: "http://www.example.com/", // domain not defined in the simulation
```
bassosimone, line 72:
```suggestion
		Input: "http://www.example.com/",
```

PR #1475, opened 2024-01-25
chore(webconnectivitylte): add large file test cases

bassosimone:
Close https://github.com/ooni/probe/issues/1336


bassosimone:
So, I instrumented the place where we read bodies for the `largeFileHTTPS` case and I see:

```
2024/01/31 18:22:59 BONSOIR, ELLIOT: maxbody=524288, len(body)=524288, err=<nil>
```

This means that the maxbody and the body length are equal. Why do we fail to set `truncated`?

PR #1490, opened 2024-02-05
BREAKING CHANGE: feat: move StreamAllContext to netxlite

bassosimone:
This diff moves the `StreamAllContext` implementation from `webconnectivitylte` to `netxlite`.

We flagged this diff as BREAKING CHANGE because we also fixed a bug where, in case of context timeout, `StreamAllContext` was suppressing the error rather than reporting it. Re-reading the code, this feels incorrect because it does not allow us to clearly know that we timed out reading the body, which could be caused by throttling.

The new behavior seems more accurate.

Because of this change, I also felt it was time to add explicit tests for cases where we download fails because it takes too much time for us to fetch the whole response body. I am not 100% sure we are correctly flagging this case, but an `http-failure` probably seems fine at the moment and we can always revisit this as we learn more about throttling.

Part of https://github.com/ooni/probe/issues/2654.


go.mod
```diff
@@ -24,7 +24,7 @@ require (
 	github.com/miekg/dns v1.1.57
 	github.com/mitchellh/go-wordwrap v1.0.1
 	github.com/montanaflynn/stats v0.7.1
-	github.com/ooni/netem v0.0.0-20231213064826-3cc1ea5e942c
+	github.com/ooni/netem v0.0.0-20240205182847-14e4ce92d41e
```
bassosimone, line 27: Updated to add additional throttling functions

internal/experiment/webconnectivitylte/cleartextflow.go
```diff
@@ -267,7 +267,7 @@ func (t *CleartextFlow) httpTransaction(ctx context.Context, network, address, a
 			t.CookieJar.SetCookies(req.URL, cookies)
 		}
 		reader := io.LimitReader(resp.Body, maxbody)
-		body, err = StreamAllContext(ctx, reader)
+		body, err = netxlite.StreamAllContext(ctx, reader)
```
bassosimone, line 270: Moved implementation

bassosimone:
Okay, I was expecting tests to fail but maybe I was overly pessimistic. The thing is that I assumed each run of the new added QA tests could product a different body because throttling is not deterministic. But, actually, the thing is that we do not regenerate the measurements at each run, we just use them along with the minipipeline. Because of this, we're fine!

internal/minipipeline/testdata/webconnectivity/generated/throttlingWithHTTP/measurement.json
```diff
@@ -0,0 +1,1599 @@
...
...3kon8w9QEv0qwn/iluXd0kMfhF9eZ6Mxjh2tL8V3IVtWlVMSNB8F4VCN+P07l1OC0ZTtQG0CxZ2p0LlYxtEuhImw1tBzv90c+0HKC+4A9cAJ4ACaASkOg9ce2uDakcUo9AqXsSC5cg6KMeIGREkdjq5rxUBHnMJIzliMSOkaFlHEdt3JgY4GnCxa5iOSoEB0bSd4PeWRtteTDBEczFLuA61J9v6jZ893D5T69+vm/KvkFy9fMKM5U+Ele4i1leS8gkd57vSo/eu7E08c6WwhZwvq9J0MOU7fB+qyAi8fdBkcurRM5+dxyMxGtbwRQrW03+2xQIqJ9Whm0uxm1IZcmijkNXVmv5Yn4LkWeTGZK98vFcCv5kyaRSu5BcMamWQ8k2O0a+gjauch+axxhR1gIDIkge+mYfMhkMDvfIwaZqo4O66UJf7P4LiE7zhhsdNqfixD1IGOnygazQ/vL4qMQk5tE5k7EHcwA1yYL8coVX+Yvhc6QzVWgyjrakXSCYsiK2Cm72FCzHs99q4oCBEM1wFjJKdDjVNCwyEXB6Fir1UDfqI6eDy03OVEN+pjOaEI7//FsgrRjNLwx7KnFeS0hA3cdXw2F1Y4JBej874hnPr1+GqnVvyOFkPJf8fMOhcPZp4VsZnfZfSrx+o2buKbifwAM0SndT5pMkw1UYYmOGGfkvi9OoU1EEK8J/LXIEMpSH4BGAaDxdDztsnxNU8jOXsEHHSvdboC9Zaqrrecwjfxpi6WpxbtJaiaYyRZ8TeTDdP9LZKCYXQldtkoapOkdYKMIK66YnBsWQmNElvxfEX1SFUjgS953bviU0JiDvVD/M7kkuNcs5lmvBV7KZlg8gJdwx2Rwl2ZQ2C9sR4BXmcmzZaVVmIRSWWfMI/XslSHpbTTKKq3Buf4HEwDKjBkL5DLjCnV6rqMCPFcWJsjM5OgKab3HeQltLjimfu8NDfe8p/YySrFNwwGL+yrAQKFz37y0tzgM+3rn8kYb8iEtPJ+bgbvQBmOZBaVe1Mgorc4LHs5sYDtqOTqwY46MBVeCRfCJXI0VWuPhHNUO8hHOG3DhiLnT+Q1XXe/nmJpvIHVOLcwYk4hWWQt1HsbJQusS2k3US1Eo0pLKfU/0oxyQRfdOI50L8c88vrCJsnoCGL/Ku0w5gmxJKUnNXy6yMESObRGk95LmiAQRzyZ72skiDpQ==",
```
bassosimone, line 1330: This shows we download _some_ body before timing out!

PR #1491, opened 2024-02-06
refactor(all): improve SummaryKeys management

bassosimone:
While approaching https://github.com/ooni/probe/issues/2667, I recognized that the approach we're using to generate `SummaryKeys` is redundant, verbose, and fragile. This diff attempts to improve our implementation.

We define new types for generating summary keys and `engine.MeasurementSummaryKeys` to either return the proper summary keys, if the experiment `TestKeys` support that or return a default value.

While there, disable the flaky `throttlingWithHTTP` QA check (see https://github.com/ooni/probe/issues/2668).


internal/experiment/dash/measurer.go
```diff
@@ -125,26 +124,26 @@ func NewExperimentMeasurer(config Config) model.ExperimentMeasurer {
 	return &Measurer{config: config}
 }
 
+var _ model.MeasurementSummaryKeysProvider = &TestKeys{}
+
 // SummaryKeys contains summary keys for this experiment.
-//
-// Note that this structure is part of the ABI contract with ooniprobe
-// therefore we should be careful when changing it.
```
bassosimone, line 131: We can safely remove this comment because we have moved the implicit requirement that every structure that implements summary MUST contain a field called `IsAnomaly` with boolean type.

internal/experiment/dnscheck/dnscheck.go
```diff
@@ -317,16 +317,3 @@ func NewExperimentMeasurer(config Config) model.ExperimentMeasurer {
...
-// Note that this structure is part of the ABI contract with ooniprobe
-// therefore we should be careful when changing it.
-type SummaryKeys struct {
-	IsAnomaly bool `json:"-"`
-}
-
-// GetSummaryKeys implements model.ExperimentMeasurer.GetSummaryKeys.
-func (m *Measurer) GetSummaryKeys(measurement *model.Measurement) (interface{}, error) {
-	return SummaryKeys{IsAnomaly: false}, nil
-}
```
bassosimone, line 332: Now we implicitly provide the correct always-returning-false implementation when an experiment does not implement the `MeasurementSummaryKeysProvider` interface. So, we can safely zap the boilerplate code concerning the summary keys for each experiment that always returned false.

internal/database/actions.go
```diff
@@ -350,25 +349,16 @@ func (d *Database) CreateOrUpdateURL(urlStr string, categoryCode string, country
...
 	// Maybe generics are not so bad after-all, heh golang?
-	isAnomalyValue := reflect.ValueOf(tk).FieldByName("IsAnomaly")
-	if isAnomalyValue.IsValid() && isAnomalyValue.Kind() == reflect.Bool {
-		isAnomaly = isAnomalyValue.Bool()
-		isAnomalyValid = true
-	}
-	msmt.TestKeys = string(tkBytes)
-	msmt.IsAnomaly = sql.NullBool{Bool: isAnomaly, Valid: isAnomalyValid}
+	msmt.TestKeys = string(skBytes)
+	msmt.IsAnomaly = sql.NullBool{Bool: sk.Anomaly(), Valid: true}
```
bassosimone, line 361: Question: do we want to avoid generating an anomaly if there is no summary? Maybe that would be closer in spirit to what the original code was doing.

PR #1505, opened 2024-02-12
fix(qatool): reduce generated test cases churn

bassosimone:
We add dns.nextdns.io to the QA suite, such that we don't get a wrong NXDOMAIN when trying to use it as the resolver.

We optionally sort the test keys to ensure there's less churn in diffs produced when regenerating minipipeline test data.

We fix ./script/updateminipipeline.bash to use ./script/go.bash for building as opposed to using go.

Also, include endpoint information in `tls_handshake_{start,stop}` and `http_transaction_{start,stop}` events such that they are sorted correctly when we sort events to reduce churn. This is necessary because we're using the endpoint information to sort the network events.

Part of https://github.com/ooni/probe/issues/2677;  the churn is still too high, and I need more changes.


internal/experiment/webconnectivitylte/testkeys.go
```diff
@@ -291,13 +291,6 @@ func (tk *TestKeys) WithDNSWhoami(fun func(*DNSWhoamiInfo)) {
 	tk.mu.Unlock()
 }
 
-// SetClientResolver sets the ClientResolver field.
-func (tk *TestKeys) SetClientResolver(value string) {
-	tk.mu.Lock()
-	tk.ClientResolver = value
-	tk.mu.Unlock()
-}
-
```
bassosimone, line 300: We don't need this function because we set the field after all measurements have completed.

internal/minipipeline/analysis.go
```diff
@@ -59,13 +59,15 @@ func NewLinearWebAnalysis(input *WebObservationsContainer) (output []*WebObserva
 		// TagDepth sort at the end of the generated list.
 		if left.TagDepth.UnwrapOr(-1) > right.TagDepth.UnwrapOr(-1) {
 			return true
-		} else if left.TagDepth.UnwrapOr(-1) < right.TagDepth.UnwrapOr(-1) {
+		}
+		if left.TagDepth.UnwrapOr(-1) < right.TagDepth.UnwrapOr(-1) {
```
bassosimone, line 63: We prefer a coding style without `else if` when there's a previous `return`.

internal/minipipeline/sorting.go
```diff
@@ -0,0 +1,149 @@
...
+func SortTLSHandshakeResults(
+	inputs []*model.ArchivalTLSOrQUICHandshakeResult) (outputs []*model.ArchivalTLSOrQUICHandshakeResult) {
+	// copy the original slice
+	outputs = append([]*model.ArchivalTLSOrQUICHandshakeResult{}, inputs...)
+
+	// sort using complex sorting rule
+	sort.SliceStable(outputs, func(i, j int) bool {
+		left, right := outputs[i], outputs[j]
+
+		if left.Address < right.Address {
```
bassosimone, line 129:
```suggestion
		// we sort by endpoint address to significantly reduce the churn
		if left.Address < right.Address {
```

internal/minipipeline/testdata/webconnectivity/generated/badSSLWithExpiredCertificate/measurement.json
```diff
@@ -28,17 +28,95 @@
...
+      {
+        "failure": null,
+        "operation": "tls_handshake_done",
+        "t0": 0.02195,
+        "t": 0.02195,
+        "transaction_id": 3,
+        "tags": [
+          "depth=0",
+          "fetch_body=true"
+        ]
```
bassosimone, line 55: There's another bug here: we're not adding the endpoint to `tls_handshake_xxx` and `http_transaction_xxx` events, which means that they do not sort correctly within the generated JSON.

bassosimone: I will fix this bug as a separate pull request because it seems this one is already huge 😞

bassosimone: Oh, and maybe instead I can just fix it here

bassosimone: Okay, so, compromise: doing here for LTE, doing in the future for the DSLs.

internal/experiment/webconnectivitylte/cleartextflow.go
```diff
@@ -253,8 +253,16 @@ func (t *CleartextFlow) httpTransaction(ctx context.Context, network, address, a
...
+		trace.Index(),
+		started,
+		"http_transaction_start",
+		network,
+		address,
+		0,
+		nil,
+		started,
+		trace.Tags()...,
 	))
```
bassosimone, line 266: This is such that we have the endpoints in this event (because... why not? It helps with sorting and I don't see why omitting a piece of information we know would ever be wrong).

internal/experiment/webconnectivitylte/cleartextflow.go
```diff
@@ -272,8 +280,16 @@ func (t *CleartextFlow) httpTransaction(ctx context.Context, network, address, a
...
+	t.TestKeys.AppendNetworkEvents(measurexlite.NewArchivalNetworkEvent(
+		trace.Index(),
+		finished,
+		"http_transaction_done",
+		network,
+		address,
+		0,
+		nil,
+		finished,
+		trace.Tags()...,
```
bassosimone, line 292: This is such that we have the endpoints in this event (because... why not? It helps with sorting and I don't see why omitting a piece of information we know would ever be wrong).

internal/experiment/webconnectivitylte/secureflow.go
```diff
@@ -308,8 +308,16 @@ func (t *SecureFlow) httpTransaction(ctx context.Context, network, address, alpn
...
+	t.TestKeys.AppendNetworkEvents(measurexlite.NewArchivalNetworkEvent(
+		trace.Index(),
+		started,
+		"http_transaction_start",
+		network,
+		address,
+		0,
+		nil,
+		started,
+		trace.Tags()...,
```
bassosimone, line 320: This is such that we have the endpoints in this event (because... why not? It helps with sorting and I don't see why omitting a piece of information we know would ever be wrong).

internal/experiment/webconnectivitylte/secureflow.go
```diff
@@ -327,8 +335,16 @@ func (t *SecureFlow) httpTransaction(ctx context.Context, network, address, alpn
...
+	t.TestKeys.AppendNetworkEvents(measurexlite.NewArchivalNetworkEvent(
+		trace.Index(),
+		finished,
+		"http_transaction_done",
+		network,
+		address,
+		0,
+		nil,
+		finished,
+		trace.Tags()...,
```
bassosimone, line 347: This is such that we have the endpoints in this event (because... why not? It helps with sorting and I don't see why omitting a piece of information we know would ever be wrong).

internal/measurexlite/tls.go
```diff
@@ -70,8 +79,17 @@ func (tx *Trace) OnTLSHandshakeDone(started time.Time, remoteAddr string, config
...
+		tx.Index(),
+		t,
+		"tls_handshake_done",
+		"tcp",
+		remoteAddr,
+		0,
+		nil,
+		t,
+		tx.tags...,
+	):
```
bassosimone, line 92: This is such that we have the endpoints in this event (because... why not? It helps with sorting and I don't see why omitting a piece of information we know would ever be wrong).

PR #1520, opened 2024-03-19
oohelperd: move prometheus metrics behind HTTP basic auth + docker image build + codepipeline buildspec

hellais:
We implement the following changes:
* Move prometheus metric behind HTTP basic auth
* Read the prometheus password from the `PROMETHEUS_METRICS_PASSWORD` environment variable like we do for ooniapi-services
* Add Dockerfile to build oohelpderd
* Add codepipeline buildspec for oonihelpderd


hellais:
This implements what we discussed in https://github.com/ooni/devops/pull/31#discussion_r1530226836

PR #1521, opened 2024-03-20
fix(pkg/gobash/version.go): validate zip files content

bassosimone:
See https://github.com/ooni/probe-cli/security/code-scanning/12

While there, make sure we run tests using go1.22, which has been released in February while I was on leave.

### DIFF
--- a/pkg/gobash/version.go
+++ b/pkg/gobash/version.go
@@ -255,6 +255,18 @@ func unpackZip(targetDir, archiveFile string) error {
 	defer zr.Close()
 
 	for _, f := range zr.File {
+		// See https://github.com/ooni/probe-cli/security/code-scanning/12
+		//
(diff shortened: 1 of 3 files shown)


pkg/gobash/version.go
```diff
@@ -255,6 +255,18 @@ func unpackZip(targetDir, archiveFile string) error {
 	defer zr.Close()
 
 	for _, f := range zr.File {
+		// See https://github.com/ooni/probe-cli/security/code-scanning/12
+		//
+		// The validRelPath function rejects empty paths, paths containing backslash, paths
+		// starting with / and paths containing ../. Additionally, according to
+		// src/archive/zip/reader.go, the zip specification only allows files containing
+		// forward slashes and considers files containing backslashes to be inscure.
```
bassosimone, line 263:
```suggestion
		// forward slashes and considers files containing backslashes to be insecure.
```

PR #1522, opened 2024-03-21
chore: update psiphon

bassosimone:
It seems there's an issue by which some `replace` directives in the `go.mod` used by Psiphon aren't picked up by us. The thing is that `replace` only works for the main module. We thus asked for clarification but we can merge this PR anyway, since it helps us to find dependency intersection between snowflake and psiphon (I am thinking mainly in terms of refraction-networking/utls).

See https://github.com/ooni/probe/issues/2690


bassosimone:
I am going to merge, because I need this to understand whether we can intersect psiphon and snowflake as dependencies.

I took note in the release check list of the implications of the `replace` directive.

PR #1524, opened 2024-03-21
chore: upgrade C dependencies

bassosimone:
See https://github.com/ooni/probe/issues/2690


bassosimone:
I needed to cancel the linux workflow because it was stuck due to https://github.com/ooni/probe/issues/2692.

PR #1546, opened 2024-04-09
doc: explain the throttling methodology

bassosimone:
- outline the current methodology;

- explain its theoretical assumptions;

- describe limitations;

- mention possible future directions.

Part of https://github.com/ooni/ooni.org/issues/1406.

Closes https://github.com/ooni/probe-cli/pull/684 because we've now gone full circle and we have infrastructure to setup a comparable experiment via richer input without the need of *implementing* a new experiment.


DecFox (approved):
As we had discussed with @bassosimone, we should also document the following:

- how we can use this methodology while looking at data from webconnectivity v0.5: analyzing speeds between domains across different temporal data points.

docs/design/dd-007-throttling.md
```diff
@@ -0,0 +1,254 @@
...
+instead, need to collect several samples over time and cross compare them using
+the [ooni/data](https://github.com/ooni/data) measurement pipeline.
+
+More specifically, we would need to compare current measurements with past
+measurements collected for the same target website by users living in the same
+country and using the same autonomous system. Alternatively, we could compare
+measurements collected during the same time frame towards different websites, even
+though this signal is weaker because it can just be caused by interconnection
+issues. In any case, these considerations imply that our methodology rests
+on the assumption that we will have several measurements for the targer websites,
```
DecFox, line 184:
```suggestion
on the assumption that we will have several measurements for the target websites,
```

DecFox (approved):
The changes look good. I left a small comment. I think this should be good to go

PR #1547, opened 2024-04-10
hack: allow using new collector with SNI overrides

bassosimone:
This PR is not meant to be merged. Rather it documents testing the new TH deployment.

See https://github.com/ooni/probe/issues/2703


bassosimone:
The reference issue https://github.com/ooni/probe/issues/2703 is now complete. We don't need this PR/PoC anymore.

PR #1552, opened 2024-04-15
Issue/2704

bassosimone:
## Checklist

## Description


internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+	- [Invalid bridge without cached data](#invalid-bridge-without-cached-data)
+	- [Invalid bridge with cached data](#invalid-bridge-with-cached-data)
+	- [Valid bridge with invalid cached data](#valid-bridge-with-invalid-cached-data)
+- [Limitations and Future Work](#limitations-and-future-work)
+
+## Goals & Assumptions
+
+We define "bridge" an IP address with the following properties:
+
+1. the IP address is not expected to change;
```
hellais, line 30:
```suggestion
1. the IP address is not expected to change frequently;
```

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+
+2. the IP address listens on port 443 and accepts _any_ incoming SNI;
+
+3. the webserver on port 443 proxies to the OONI APIs.
```
hellais, line 34:
```suggestion
3. the webserver on port 443 provides unified access to [OONI API services](https://docs.ooni.org/backend/ooniapi/services/).
```

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+5. adopt a censored-users-first approach where the strategy we use by default
+should allow for smooth operations _for them_ rather than prioritizing the
+non-censored case and using additional tactics as the fallback;
+
+6. try to defer sending the true `SNI` on the wire, therefore trying to
+avoid triggering potential residual censorship blocking a given TCP endpoint
+for some time regardless of what `SNI` is being used next;
+
+7. allow users to force specific bridges and SNIs by editing `$OONI_HOME/engine/bridges.conf`.
+
```
hellais, line 63:
```suggestion
8. In the event of no censorship affecting OONI API Services, the fact no tactic is working should not be a hindrance to communicating with the API. 
```

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+3. then use the first two tactics coming from bridges, if any;
+
+4. then use the first two tactics coming from the DNS, if successful;
+
+5. finally, randomly remix the remaining tactics.
+
+Excluding the case where we have a valid entry in `bridges.conf`, the following
+diagram illustrates how we're mixing tactics:
+
```
hellais, line 483: Do I understand it correctly, that in the default thing we ship in probe we effectively have a `bridges.conf` file with the hardcoded IP?

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+
+1. if there is a `$OONI_HOME/engine/bridges.conf` with a valid entry,
+use it without trying more tactics; otherwise,
+
+2. use the first two tactics coming from stats, if any;
+
+3. then use the first two tactics coming from bridges, if any;
+
+4. then use the first two tactics coming from the DNS, if successful;
```
hellais, line 477: Can we flip the order of these around in the algorithm so that if DNS is working we use that instead of relying on the bridges?

Basically it would be optimal if we first tried everywhere the simple case (i.e. direct DNS resolution without the hardcoded bridge) and then only if that failed we moved onto trying the hardcoded bridge IP.

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+[      2.093148] <info> httpsDialer: [#4] TLSHandshake with 162.55.247.208:443 SNI=api.ooni.io ALPN=[h2 http/1.1]... ok
+[      2.093181] <info> httpsDialer: [#4] TLSVerifyCertificateChain api.ooni.io... started
+[      2.095923] <info> httpsDialer: [#4] TLSVerifyCertificateChain api.ooni.io... ok
+[      2.096054] <info> httpsDialer: [#1] TCPConnect 10.0.0.1:443... interrupted
+[      2.096077] <info> httpsDialer: [#2] TCPConnect 10.0.0.1:443... interrupted
+```
+
+**Listing 8.** Run with no previous cached state and unreachable hardcoded bridge address.
+
+After 2s, we start dialing with the IP addresses obtained through the DNS.
```
hellais, line 641: My understanding is that this is taking this amount of time, because we are trying the bridges first (see: https://github.com/ooni/probe-cli/pull/1552/files#r1576554009), can we invert the order so that we always try first via DNS in the case of an empty cache?

hellais (changes requested):
Thanks for putting this together. Having the design doc is really helpful in understanding how this system is meant to work and why it's built in a certain way.

I left some comments of improvements to make the design doc and some areas that I would like to understand how complex it is to change the logic to try DNS as the first thing in the case of an empty cache.

It would also be good to add some docs somewhere on what needs to be done to zap the cache to start afresh, so that we are able to do that as part of app releases.

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
+2. overcome SNI-based censorship for "api.ooni.io" and test helpers by choosing
+from a pre-defined list of SNIs;
+
+3. remember and use tactics for creating TLS connections that worked previously;
+
+4. recover ~quickly if the conditions change (e.g., if a bridge is discontinued);
+
+5. adopt a censored-users-first approach where the strategy we use by default
+should allow for smooth operations _for them_ rather than prioritizing the
+non-censored case and using additional tactics as the fallback;
```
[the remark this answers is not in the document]
bassosimone: Can you provide a suggestion of how you would rephrase this requirement? 🙏

internal/enginenetx/DESIGN.md
bassosimone: I find this new requirement to be partially aligned with this requirement:

> 4. recover ~quickly if the conditions change (e.g., if a bridge is discontinued);

I say this because they both morally say that the system should work without big performance penalties in case there's no censorship and/or conditions change. However, your proposal for 8 is more strict and in fact looks to me like a stricter version of 4.

As such, could you propose an edit that amends 4 rather than introducing 8? Thank you!

internal/enginenetx/DESIGN.md
bassosimone: We have discussed this in person. Here's a recap of the conversation, for the records. The `bridges.conf` is user provided and, as such, completely overrides any other tactic. What we ship with the probe is:

1. similar to `bridges.conf` in that it hardcodes a configuration;
2. different from `bridges.conf` in that it is just one of the pieces of information used to dial.

I remember that, when we chatted, you had some suggestions regarding wording that could have helped clarify the difference. Please, provide them such that I can integrate them! 🙏

internal/enginenetx/DESIGN.md
bassosimone: Yes, it is possible to do that by changing the order with which we construct and compose policies.

I will modify the code to this effect.

internal/enginenetx/DESIGN.md
bassosimone: Yes, it is possible to do that by changing the order with which we construct and compose policies.

I will modify the code to this effect.

bassosimone:
@hellais wrote:

> It would also be good to add some docs somewhere on what needs to be done to zap the cache to start afresh, so that we are able to do that as part of app releases.

Sure, I will document that. I suppose the main use case you are thinking about are mobile apps. In such a case, I would recommend creating an API that tells the probe to clear the cache, because that would reduce coupling.

internal/enginenetx/DESIGN.md
hellais: done

internal/enginenetx/DESIGN.md
hellais: Re-reading it I think it's fine. I'm not sure how the text can be improved if not trying to make it more concise, but I suggest we do that as future iterations.

bassosimone: ack!

internal/enginenetx/DESIGN.md
```diff
@@ -0,0 +1,815 @@
...
```
bassosimone, line 56:
```suggestion
5. for users in censored regions it should be possible to use tactics to overcome the restrictions;
```

Adding suggestion from @hellais

internal/enginenetx/DESIGN.md
bassosimone: Thanks!

internal/enginenetx/DESIGN.md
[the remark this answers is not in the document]
bassosimone: Thanks, added as a suggestion!

bassosimone:
To implement the required changes, I think the best approach is that of decoupling a policy for generating tactics and a policy for mixing, rather than invoking mixing algorithms inside policies. This feels like a small refactoring and I started experimenting with it in this branch. But, this change will pay off a lot, because, at the moment, there are some implicit assumptions in the ordering of policies and how mixing works. Whereas, after the change, it will be much easier to compose several algorithms together. So, I am going to implement these changes first and commit them into `master` before circling back and (a) adapting the policy we're using here and (b) updating the design document.

bassosimone:
I guess this PR has served its purpose as being a platform for discussing the previously existing design, for experimenting with fast recover in case of issues, and as an integration branch for testing out changes to be merged. When we'll merge https://github.com/ooni/probe-cli/pull/1595, we should then close this PR and keep it as a documentation of the design discussion that happened, and the various change that were made to address the requested design changes.

PR #1555, opened 2024-04-17
refactor(enginenetx): assign InitialDelay when dialing

bassosimone:
This diff refactors enginenetx to assign InitialDelay only when dialing. It is pointless to do that before. Also, take advantage of algorithms introduced by https://github.com/ooni/probe-cli/pull/1556 to make the code more compact.

Part of https://github.com/ooni/probe/issues/2704.


internal/enginenetx/bridgespolicy.go
```diff
@@ -27,33 +27,18 @@ var _ httpsDialerPolicy = &bridgesPolicy{}
...
-		for tx := range p.maybeRewriteTestHelpersTactics(p.Fallback.LookupTactics(ctx, domain, port)) {
-			tx.InitialDelay = happyEyeballsDelay(index)
-			index += 1
-			out <- tx
-		}
-	}()
-
-	return out
+		p.maybeRewriteTestHelpersTactics(p.Fallback.LookupTactics(ctx, domain, port)),
+	)
```
bassosimone, line 41: Here we're replacing an inline expansion of `mixSequentially` with a call to `mixSequentially`.

bassosimone: Additionally, the `tx.InitialDelay` is gone because we'll compute that once when dialing.

internal/enginenetx/dnspolicy.go
```diff
@@ -56,10 +56,10 @@ func (p *dnsPolicy) LookupTactics(
 		}
 
 		// The tactics we generate here have SNI == VerifyHostname == domain
-		for idx, addr := range addrs {
+		for _, addr := range addrs {
 			tactic := &httpsDialerTactic{
 				Address:        addr,
-				InitialDelay:   happyEyeballsDelay(idx),
+				InitialDelay:   0, // set when dialing
```
bassosimone, line 62: Here we removed computing the happy-eyeballs delay because we're computing it when dialing.

internal/enginenetx/httpsdialer.go
```diff
@@ -209,7 +209,7 @@ func (hd *httpsDialer) DialTLSContext(ctx context.Context, network string, endpo
 
 	// The emitter will emit tactics and then close the channel when done. We spawn 16 workers
 	// that handle tactics in parallel and post results on the collector channel.
-	emitter := hd.policy.LookupTactics(ctx, hostname, port)
+	emitter := httpsFilterTactics(hd.policy.LookupTactics(ctx, hostname, port))
```
bassosimone, line 212: This is where we're computing the happy eyeballs delay when dialing.

internal/enginenetx/statspolicy.go
```diff
@@ -30,48 +30,14 @@ var _ httpsDialerPolicy = &statsPolicy{}
...
-			uniq[key]++
-
-			// 🚀!!!
-			t.InitialDelay = happyEyeballsDelay(index)
-			index += 1
-			out <- t
-		}
-
+	// avoid emitting nil tactics and duplicate tactics
+	return filterOnlyKeepUniqueTactics(filterOutNilTactics(mixSequentially(
```
bassosimone, line 34: Rather than using `maybeEmitTactic` here we're using filtering to achieve 2/3 of the original goals, i.e., filtering out nils and only keeping unique tactics. The final original goal, i.e., adding happy-eyeballs delays, is now implemented just once when we're dialing, so there's no point in doing that also here.

PR #1572, opened 2024-04-29
chore(probeservices): improve tests for /api/v1/login

bassosimone:
Part of https://github.com/ooni/probe/issues/2718


internal/probeservices/login_test.go
```diff
@@ -2,75 +2,266 @@ package probeservices
...
+		state := &testingx.OONIBackendWithLoginFlow{}
+
+		// expose the state via HTTP
+		srv := testingx.MustNewHTTPServer(state.NewMux())
+		defer srv.Close()
+
+		// create a probeservices client
+		client := newclient()
+
+		// override the HTTP client so we speak with out local server rather than the true backend
```
bassosimone, line 64:
```suggestion
		// override the HTTP client so we speak with our local server rather than the true backend
```

PR #1577, opened 2024-04-30
fix(httpclientx): add support for cloudfronting

bassosimone:
I originally thought about dropping support for cloudfronting given that we are not using it anymore. I wonder why.

It's a few lines of code (including testing) and it seems a good idea to keep supporting it.

Part of https://github.com/ooni/probe/issues/2723

### DIFF
--- a/internal/httpclientx/getjson_test.go
+++ b/internal/httpclientx/getjson_test.go
@@ -163,7 +163,7 @@ func TestGetJSONHeadersOkay(t *testing.T) {
 		t.Fatal("unexpected Accept-Encoding value", value)
 	}
 
-	// now make sure we could our cloudfronting
+	// now make sure we could use cloudfronting
 	if gothost != "www.cloudfront.com" {
 		t.Fatal("unexpected Host value", gothost)
 	}
--- a/internal/httpclientx/getraw_test.go
+++ b/internal/httpclientx/getraw_test.go
@@ -125,7 +125,7 @@ func TestGetRawHeadersOkay(t *testing.T) {
 		t.Fatal("unexpected Accept-Encoding value", value)
 	}
 
-	// now make sure we could our cloudfronting
+	// now make sure we could use cloudfronting
 	if gothost != "www.cloudfront.com" {
 		t.Fatal("unexpected Host value", gothost)
 	}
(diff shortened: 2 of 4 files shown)


internal/httpclientx/getjson_test.go
```diff
@@ -159,6 +162,11 @@ func TestGetJSONHeadersOkay(t *testing.T) {
 	if value := gotheaders.Get("Accept-Encoding"); value != "gzip" {
 		t.Fatal("unexpected Accept-Encoding value", value)
 	}
+
+	// now make sure we could our cloudfronting
```
bassosimone, line 166:
```suggestion
	// now make sure we could use cloudfronting
```

internal/httpclientx/getraw_test.go
```diff
@@ -121,6 +124,11 @@ func TestGetRawHeadersOkay(t *testing.T) {
 	if value := gotheaders.Get("Accept-Encoding"); value != "gzip" {
 		t.Fatal("unexpected Accept-Encoding value", value)
 	}
+
+	// now make sure we could our cloudfronting
```
bassosimone, line 128:
```suggestion
	// now make sure we could use cloudfronting
```

internal/httpclientx/getxml_test.go
```diff
@@ -155,6 +158,11 @@ func TestGetXMLHeadersOkay(t *testing.T) {
 	if value := gotheaders.Get("Accept-Encoding"); value != "gzip" {
 		t.Fatal("unexpected Accept-Encoding value", value)
 	}
+
+	// now make sure we could our cloudfronting
```
bassosimone, line 162:
```suggestion
	// now make sure we could use cloudfronting
```

internal/httpclientx/postjson_test.go
```diff
@@ -236,6 +239,11 @@ func TestPostJSONCommunicationOkay(t *testing.T) {
 	if value := gotheaders.Get("Accept-Encoding"); value != "gzip" {
 		t.Fatal("unexpected Accept-Encoding value", value)
 	}
+
+	// now make sure we could our cloudfronting
```
bassosimone, line 243:
```suggestion
	// now make sure we could use cloudfronting
```

PR #1586, opened 2024-05-03
httpclientx: implement fast recovery and return endpoint index

bassosimone:
Closes https://github.com/ooni/probe/issues/2728


internal/httpclientx/overlapped.go
```diff
@@ -83,81 +95,149 @@ func NewOverlappedPostJSON[Input, Output any](input Input, config *Config) *Over
...
+	//
+	// this happens if the user provided no endpoints to measure
 	if len(errorv) <= 0 {
 		errorv = append(errorv, ErrGenericOverlappedFailure)
 	}
 
 	// return zero value and errors list
-	return *new(Output), errors.Join(errorv...)
+	//
+	// note thay errors.Join returns nil if all the errors are nil or the
```
bassosimone, line 218:
```suggestion
	// note that errors.Join returns nil if all the errors are nil or the
```

PR #1609, opened 2024-06-04
refactor: MeasurementTarget -> MeasurementInput

bassosimone:
While starting to work on richer input, I realised that it would be very handy to reserve the name "target" to describe a richer input target as the tuple containing options and some input.

However, using such a naming would ~clash with the definition of MeasurementTarget, which is a string definition that is only there to say semantically that the Input field of a measurement is the "thing" that we measure, as well as to apply a specific JSON serialization policy that is consistent with the historical behavior of OONI Probe.

Therefore, for making semantic space for a richer input target, I am going to refactor MeasurementTarget to be MeasurementInput.

No functional change.


DecFox:
LGTM!

bassosimone:
I have a LGTM from @DecFox so I am going to merge as soon as tests are all green.

PR #1610, opened 2024-06-04
cleanup(oonirun): remove unused testing-hook argument

bassosimone:
This diff removes an unused argument from a testing hook. Removing this argument simplifies the implementation of the richer input patches and does not change the behavior in any way since the argument wasn't used.


DecFox:
LGTM!

PR #1612, opened 2024-06-05
refactor(experiment): make report goroutine safe

bassosimone:
This diff refactors *engine.experiment to make the report field goroutine safe. It also moves at the bottom of experiment.go code that was intermixed with *engine.experiment methods.


DecFox:
LGTM!

DecFox (approved):

PR #1613, opened 2024-06-05
cleanup: remove asynchronous experiments

bassosimone:
We originally developed asynchronous experiments as a mean to support websteps, a nettest that returned more than one measurement per invocation of its Measure method.

Since then, we removed websteps. Therefore, this code is currently technically unused. Additionally, this code further complicates implementing richer input, because it is another way of performing measurements.

As such, in the interest of switfly moving forward with richer input _and_ of simplifying the engine, we are now removing this unused functionality from the tree.

While there, better the documentation of OnProgress and undeprecate functions to open/close reports since it's clear that, for now, using a submitter in cmd/ooniprobe is a bit of a stretch given our current goals. So, it feels best to avoid deprecating until we have nice plan for replacement.

To have more confidence that the job we did is ~correct, we use the following table to cross compare the operations that we previously performed for running sync experiments (i.e., all experiments) in an async way to the code we're using after this diff to run experiments. (Note that the diff itself is such that you can easily see the deleted and the added code inside of the `experiment.go` file.) In both cases, we're looking at the operations performed starting from `MeasureWithContext`.

| Operation                   | Before | After |          
| --------------------------- | ------ | ----- |
| MaybeLookupLocationContext  | yes    | yes   |
| use e.session.byteCounter   | yes    | yes   |
| use e.byteCounter           | yes    | yes   |
| newMeasurement              | yes    | yes   |
| save start time             | yes    | yes   |
| initialize experiment args  | yes    | yes   |
| call measurer.Run           | yes    | yes   |
| save end time               | yes    | yes   |
| compute measurement runtime | yes    | yes   |
| scrub measurement           | yes    | yes   | 
| return measurement          | yes    | yes   |

Depends on https://github.com/ooni/probe-cli/pull/1612

### DIFF
--- a/internal/engine/experiment.go
+++ b/internal/engine/experiment.go
@@ -96,104 +96,6 @@ func (e *experiment) ReportID() string {
 	return report.ReportID()
 }
 
-// experimentAsyncWrapper makes a sync experiment behave like it was async
-type experimentAsyncWrapper struct {
-	*experiment
-}
-
-var _ model.ExperimentMeasurerAsync = &experimentAsyncWrapper{}
-
-// RunAsync implements ExperimentMeasurerAsync.RunAsync.
-func (eaw *experimentAsyncWrapper) RunAsync(
-	ctx context.Context, sess model.ExperimentSession, input string,
-	callbacks model.ExperimentCallbacks) (<-chan *model.ExperimentAsyncTestKeys, error) {
-	out := make(chan *model.ExperimentAsyncTestKeys)
-	measurement := eaw.experiment.newMeasurement(input)
-	start := time.Now()
-	args := &model.ExperimentArgs{
-		Callbacks:   eaw.callbacks,
-		Measurement: measurement,
-		Session:     eaw.session,
-	}
-	err := eaw.experiment.measurer.Run(ctx, args)
-	stop := time.Now()
-	if err != nil {
-		return nil, err
-	}
-	go func() {
-		defer close(out) // signal the reader we're done!
-		out <- &model.ExperimentAsyncTestKeys{
-			Extensions:         measurement.Extensions,
-			Input:              measurement.Input,
-			MeasurementRuntime: stop.Sub(start).Seconds(),
-			TestKeys:           measurement.TestKeys,
-			TestHelpers:        measurement.TestHelpers,
-		}
-	}()
-	return out, nil
-}
-
-// MeasureAsync implements [model.Experiment].
-func (e *experiment) MeasureAsync(
-	ctx context.Context, input string) (<-chan *model.Measurement, error) {
-	err := e.session.MaybeLookupLocationContext(ctx) // this already tracks session bytes
-	if err != nil {
-		return nil, err
-	}
-	ctx = bytecounter.WithSessionByteCounter(ctx, e.session.byteCounter)
-	ctx = bytecounter.WithExperimentByteCounter(ctx, e.byteCounter)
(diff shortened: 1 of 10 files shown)


DecFox (approved):
LGTM!

internal/engine/experiment.go
```diff
@@ -277,6 +179,79 @@ func (e *experiment) OpenReportContext(ctx context.Context) error {
...
+
+	// Handle the case where there was a fundamental error.
+	if err != nil {
+		return nil, err
+	}
+
+	// Make sure we record the measurement runtime.
+	measurement.MeasurementRuntime = stop.Sub(start).Seconds()
+
+	// Scub the measurement removing the probe IP addr from it. We are 100% sure we know
```
ainghazal, line 237: s/Scub/Scrub/

bassosimone:
```suggestion
	// Scrub the measurement removing the probe IP addr from it. We are 100% sure we know
```

internal/engine/experiment.go
```diff
@@ -277,6 +179,79 @@ func (e *experiment) OpenReportContext(ctx context.Context) error {
...
+	// information such that we correctly populate the measurement and also
+	// VERY IMPORTANTLY to scrub the IP address from the measurement.
+	//
+	// Also, this SHOULD happen before wrapping the context for byte counting
+	// since MaybeLookupLocationContext already accounts for bytes I/O.
+	//
+	// TODO(bassosimone,DecFox): historically we did this only for measuring
+	// and not for opening a report, which probably is not correct. Because the
+	// function call is idempotent, call it also when opening a report?
+	if err := e.session.MaybeLookupLocationContext(ctx); err != nil {
```
ainghazal, line 194: I was wondering if `MaybeLookupLocationWithContext` would be more descriptive?

Re. moment to call it, a subtle concern here might be the low likelihood that there's a network change between measurement and submission. 

Thinking mostly in asynchronous submission, I think it'd be healthy to a) call it when submitting; b) raise an error (or warning) if submitting a msmt for a different (CC,ASN) that the report we've opened. I can work on that as part of my future deliverables, that include deferred submission (which is an special case).

bassosimone: @ainghazal wrote:

> I was wondering if `MaybeLookupLocationWithContext` would be more descriptive?

Yeah, so, first of all we can gain some more characters by removing `WithContext`, which is an artifact from when we also had a version of the call without a context. How would you suggest to change the name?

> Re. moment to call it, a subtle concern here might be the low likelihood that there's a network change between measurement and submission.
>
> Thinking mostly in asynchronous submission, I think it'd be healthy to a) call it when submitting; b) raise an error (or warning) if submitting a msmt for a different (CC,ASN) that the report we've opened. I can work on that as part of my future deliverables, that include deferred submission (which is an special case).

As far as I know, when we're resubmitting we're using a `probeservices.Submitter`: https://github.com/ooni/probe-cli/blob/master/internal/probeservices/collector.go#L185. The code should open reports only based on the data stored inside the measurement, so it should support the use case of measuring, say, w/o VPN, and then submitting with VPN. It would be great to have another pair of eyes looking at the code as well as having some manual testing to ensure whether this is still true. It might also be good to have a test guaranteeing that this is the case.

ainghazal: >  How would you suggest to change the name?
if so, perhaps just `MaybeLookupLocation`?

bassosimone: Sure!

PR #1614, opened 2024-06-05
fix(registry): instantiate factory for each invocation

bassosimone:
The current structure of the `*registry.Factory` registry is that we register an experiment like this:

```Go
func init() {
	AllExperiments["dnscheck"] = &Factory{
		build: func(config interface{}) model.ExperimentMeasurer {
			return dnscheck.NewExperimentMeasurer(
				*config.(*dnscheck.Config),
			)
		},
		config:           &dnscheck.Config{},
		enabledByDefault: true,
		inputPolicy:      model.InputOrStaticDefault,
	}
}
```

Then, when we're setting options, we're modifying the `config` directly with code like this:

```Go
// SetOptionAny sets an option given any value.
func (b *Factory) SetOptionAny(key string, value any) error {
	field, err := b.fieldbyname(b.config, key)
	if err != nil {
		return err
	}
	switch field.Kind() {
	case reflect.Int64:
		return b.setOptionInt(field, value)
	case reflect.Bool:
		return b.setOptionBool(field, value)
	case reflect.String:
		return b.setOptionString(field, value)
	default:
		return fmt.Errorf("%w: %T", ErrUnsupportedOptionType, value)
	}
}
```

 Finally, we pass the modified config to a new experiment when we're creating it:

```Go
func (b *Factory) NewExperimentMeasurer() model.ExperimentMeasurer {
	return b.build(b.config)
}
```

This means that, if we run two back experiments that both require options, we're going to always reuse the same option structure. This feels wrong. We should instead construct a new factory each time, so we start from empty options:

```Go
func init() {
	AllExperiments["dnscheck"] = func() *Factory {
		return &Factory{
			build: func(config interface{}) model.ExperimentMeasurer {
				return dnscheck.NewExperimentMeasurer(
					*config.(*dnscheck.Config),
				)
			},
			config:           &dnscheck.Config{},
			enabledByDefault: true,
			inputPolicy:      model.InputOrStaticDefault,
		}
	}
}
```

This diff applies this very simple mechanical change.

Note that so far we were ~good with the current codebase because we don't use options much and generally we invoke each experiment just once per run. This is going to change with OONI Run v2 and richer input. Therefore, it makes sense to tackle this issue now in the context of https://github.com/ooni/probe/issues/2607.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @bassosimone @hellais
+* @bassosimone @hellais @DecFox
--- a/internal/cmd/miniooni/main.go
+++ b/internal/cmd/miniooni/main.go
@@ -232,7 +232,7 @@ func registerOONIRun(rootCmd *cobra.Command, globalOptions *Options) {
 
 // registerAllExperiments registers a subcommand for each experiment
 func registerAllExperiments(rootCmd *cobra.Command, globalOptions *Options) {
-	for name, factory := range registry.AllExperiments {
+	for name, ff := range registry.AllExperiments {
 		subCmd := &cobra.Command{
 			Use:   name,
 			Short: fmt.Sprintf("Runs the %s experiment", name),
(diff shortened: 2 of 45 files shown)


DecFox (approved):
LGTM!

PR #1616, opened 2024-06-06
refactor: move engine.InputLoader to targetloading

bassosimone:
This commit moves the engine.InputLoader type to a new package called inputloading and adapts the naming to avoid stuttering. We therefore have engine.InputLoaderSession => targetloading.Session and other similar renames.

The reason for moving InputLoader is that the engine package depends on registry, and, per the plan described by the first richer input PR, https://github.com/ooni/probe-cli/pull/1615, we want to move target loading directly inside the registry. To this end, we need to move the target loading feature outside of engine to avoid creating import loops, which prevent the code from compiling because Go does not support them.

While there, name the package targetloading rather than inputloading since richer input is all about targets, where a target is defined by the (input, options) tuple. Also, try to consistently rename types to mention targets.

We keep an integration test inside the engine package because it seems such an integration test was checking both engine and the Loader together. We may further refactor this test in the future.

Part of https://github.com/ooni/probe-cli/pull/1612


internal/inputloading/inputloading.go
```diff
@@ -1,4 +1,5 @@
-package engine
+// Package inputloading contains common code to load input.
```
bassosimone, line 1:
```suggestion
// Package inputloading contains common code to load experiments inputs and options.
```

pkg/oonimkall/taskrunner.go
```diff
@@ -169,16 +170,16 @@ func (r *runnerForTask) Run(rootCtx context.Context) {
...
 
 	// TODO(bassosimone): replace the following code with an
-	// invocation of the InputLoader. Since I am making these
+	// invocation of the targetloading.Loader. Since I am making these
 	// changes before a release and I've already changed the
 	// code a lot, I'd rather avoid changing it even more,
 	// for the following reason:
 	//
-	// If we add an call InputLoader here, this code will
+	// If we add an call targetloading.Loader here, this code will
```
DecFox, line 178:
```suggestion
	// If we add and call targetloading.Loader here, this code will
```

DecFox (approved):
LGTM! I left a small remark.

PR #1617, opened 2024-06-06
refactor: construct TargetLoader using ExperimentBuilder

bassosimone:
This diff completes the set of preliminary richer input diffs. We build the TargetLoader using the ExperimentBuilder, which in turn uses a registry.Factory under the hood. This means that we can load targets for each experiment, because the actual implementation of the TargetLoader can be experiment dependent.


DecFox (approved):
LGTM!

PR #1618, opened 2024-06-06
feat(dnscheck): implement richer input

bassosimone:
This diff:

1. modifies `./internal/registry` and its `dnscheck.go` file such that `dnscheck` has its own private target loader;
2. modifies `./internal/targetloading` to allow reusing the code to read static input from CLI and from files;
3. modifies `./internal/model` to pass optional richer input to experiments;
4. modifies `./internal/engine` to pass richer input to experiments;
5. modifies `./internal/experiment/dnscheck` to read and use the richer input it is passed;
6. implements a custom target loader for `./internal/experiment/dnscheck` returning some static entries;
7. modifies `./internal/oonirun` to make sure richer-input experiments honor their options by moving option loading before the loading of targets.

Once this diff is merged, `miniooni` and `ooniprobe` will run `dnscheck` with richer input. Obviously, we are still lacking an API to fetch richer input, so for now this is static richer input, suitable as a starting point to land more diffs.

### DIFF
--- a/internal/experiment/dnscheck/dnscheck.go
+++ b/internal/experiment/dnscheck/dnscheck.go
@@ -130,7 +130,7 @@ func (m *Measurer) Run(ctx context.Context, args *model.ExperimentArgs) error {
 		return ErrInputRequired
 	}
 	target := args.Target.(*Target)
-	config, input := target.options, target.input
+	config, input := target.Options, target.URL
 
 	// 1. fill the measurement with test keys
 	tk := new(TestKeys)
--- a/internal/experiment/dnscheck/dnscheck.go
+++ b/internal/experiment/dnscheck/dnscheck.go
@@ -96,7 +96,6 @@ type TestKeys struct {
 
 // Measurer performs the measurement.
 type Measurer struct {
-	Config
 	Endpoints *Endpoints
 }
 
@@ -125,6 +124,10 @@ func (m *Measurer) Run(ctx context.Context, args *model.ExperimentArgs) error {
 	measurement := args.Measurement
 	sess := args.Session
 
+	// 0. obtain the richer input target, config, and input or panic
+	target := args.Target.(*Target)
+	config, input := target.options, target.input
+
 	// 1. fill the measurement with test keys
 	tk := new(TestKeys)
 	tk.Lookups = make(map[string]urlgetter.TestKeys)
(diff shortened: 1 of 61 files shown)


internal/experiment/dnscheck/dnscheck.go
```diff
@@ -114,7 +114,7 @@ func (m *Measurer) ExperimentVersion() string {
 // errors are in addition to any other errors returned by the low level packages
 // that are used by this experiment to implement its functionality.
 var (
-	ErrInputRequired        = errors.New("this experiment needs input")
+	ErrInputRequired        = targetloading.ErrInputRequired
```
bassosimone, line 117: Note: this change breaks the ABI in terms of the exact error string that we emit in such a case. Because we do not have an API/ABI strict stability guarantee and because it makes sense to avoid having so many duplicate errors, I think it makes sense to apply this change in the interested of reducing unnecessary duplication a bit.

internal/oonirun/experiment_test.go
```diff
@@ -70,10 +67,12 @@ func TestExperimentRunWithFailureToSubmitAndShuffle(t *testing.T) {
...
-								// Implementation note: the convention for input-less experiments is that
-								// they require a single entry containing an empty input.
-								entry := model.NewOOAPIURLInfoWithDefaultCategoryAndCountry("")
-								return []model.ExperimentTarget{entry}, nil
+								results := []model.ExperimentTarget{
+									model.NewOOAPIURLInfoWithDefaultCategoryAndCountry("a"),
+									model.NewOOAPIURLInfoWithDefaultCategoryAndCountry("b"),
+									model.NewOOAPIURLInfoWithDefaultCategoryAndCountry("c"),
+								}
+								return results, nil
```
bassosimone, line 75: This fixes a test that I broke previously. Now that we have a loader, we need to return "a", "b", "c" here rather than using the Inputs fields above. With this change we're back at 100% coverage.

internal/registry/factory_test.go
```diff
@@ -860,3 +861,49 @@ func TestFactoryNewTargetLoaderWebConnectivity(t *testing.T) {
...
+func TestFactoryCustomTargetLoaderForRicherInput(t *testing.T) {
+	// create factory creating a custom target loader
+	factory := &Factory{
+		build:            nil,
+		canonicalName:    "",
+		config:           &customConfig{},
+		enabledByDefault: false,
+		inputPolicy:      "",
+		interruptible:    false,
+		newLoader: func(config *targetloading.Loader, options any) model.ExperimentTargetLoader {
```
DecFox, line 885: I think we should have another test case similar to this one where we set `newLoader` as `nil` and assert that we get a valid loader from the passed in `ExperimentTargetLoaderConfig`

internal/reflectx/reflectx_test.go
```diff
@@ -0,0 +1,70 @@
...
+	Age  int64
+	Ch   chan int64
+	F    bool
+	Fmp  map[string]*bool
+	Fp   *bool
+	Fv   []bool
+	Fvp  []*bool
+	Name string
+	Ptr  *int64
+	V    []int64
```
DecFox, line 20:
```suggestion
	V    []int64
	p    string          
```
Although `_` is a private field, I think we should another private field explicitly to clarify that we only check for public fields using the `IsExported` check.

DecFox (changes requested):
The changes look good to me. I left some comments which I think should make the diff good to go.

internal/registry/factory_test.go
bassosimone: ✔️

internal/reflectx/reflectx_test.go
bassosimone: ✔️

DecFox (approved):
LGTM!

PR #1619, opened 2024-06-07
cleanup(pkg/oonimkall): simplify test code

bassosimone:
This diff simplifies test code in `pkg/oonimkall` in preparation for further richer-input related changes.

While there (a) realize we don't need anymore to init the example experiment name in its factory constructor, so normalize its construction to be equal to all the other experiments; (b) realize that the `TestTaskRunnerRun` does not require the network and is ~short, so there's no point to skip it when in short mode; (c) avoid using the deprecated `legacy/mockable` package and instead use the `mocks` package.


pkg/oonimkall/taskmocks_test.go
DecFox, line 1: As discussed with @bassosimone, we no longer have custom mocks and we should rename this file to `taskutils_test.go`

DecFox (approved):
The changes look good. As we discussed, I left a remark for changing a file name.

PR #1620, opened 2024-06-07
fix(oonimkall): use richer-input-aware target loader

bassosimone:
This pull request fixes technical debt where `./pkg/oonimkall` ws using custom code to load experiments targets rather than using a (richer-input-aware) loader. We make sure to use a context with timeout when loading.

Web Connectivity should continue to WAI on mobile after this change, given that the app is calling check-in itself and it is still a supported use case to execute an experiment with explicit input.

We also start preparing for simplifying both `./pkg/oonimkall` and the mobile apps, by providing the category code and the country code inside the `status.measurement_start` event, which will allow mobile apps to update its URLs table in response to such event, rather than manually calling the check-in API and providing inputs to the engine.

Closes https://github.com/ooni/probe/issues/2764

Closes https://github.com/ooni/probe/issues/2765


pkg/oonimkall/taskrunner.go
```diff
@@ -180,47 +179,23 @@ func (r *runnerForTask) Run(rootCtx context.Context) {
...
-			r.emitter.EmitFailureStartup("experiment does not accept input")
-			return
-		}
-		r.settings.Inputs = append(r.settings.Inputs, "")
+	// Load targets. Note that, for Web Connectivity, the mobile app has
+	// already loaded inputs and provides them as r.settings.Inputs.
+	loader := builder.NewTargetLoader(&model.ExperimentTargetLoaderConfig{
+		CheckInConfig: &model.OOAPICheckInConfig{
+			// Not needed since the app already provides the
+			// inputs to use for Web Connectivity.
```
bassosimone, line 187: This is not necessarily true anymore. We should probably forward the relevant fields.

pkg/oonimkall/taskrunner.go
```diff
@@ -180,47 +179,23 @@ func (r *runnerForTask) Run(rootCtx context.Context) {
...
+		},
+		Session:      sess,
+		StaticInputs: r.settings.Inputs,
+		SourceFiles:  []string{},
+	})
+	targets, err := loader.Load(rootCtx)
```
bassosimone, line 193: I think we should have a context to bound the maximum runtime of this step here.

pkg/oonimkall/taskrunner_test.go
```diff
@@ -354,77 +370,15 @@ func TestTaskRunnerRun(t *testing.T) {
...
-				{Key: eventTypeStatusProgress, Count: 3},
-				{Key: eventTypeStatusGeoIPLookup, Count: 1},
-				{Key: eventTypeStatusResolverLookup, Count: 1},
-				{Key: eventTypeFailureStartup, Count: 1},
-				{Key: eventTypeStatusEnd, Count: 1},
+		fake.Builder.MockNewTargetLoader = func(config *model.ExperimentTargetLoaderConfig) model.ExperimentTargetLoader {
+			return &mocks.ExperimentTargetLoader{
+				MockLoad: func(ctx context.Context) ([]model.ExperimentTarget, error) {
+					return nil, errors.New("mocked error")
+				},
```
bassosimone, line 380: Because now we don't switch explicitly on the InputPolicy, we need just one test case for failing target loading.

pkg/oonimkall/taskrunner_test.go
```diff
@@ -462,12 +416,9 @@ func TestTaskRunnerRun(t *testing.T) {
 		assertReducedEventsLike(t, expect, reduced)
 	})
 
-	t.Run("with success and InputNone policy", func(t *testing.T) {
+	t.Run("with success and just a single entry", func(t *testing.T) {
 		runner, emitter := newRunnerForTesting()
 		fake := fakeSuccessfulDeps()
-		fake.Builder.MockInputPolicy = func() model.InputPolicy {
-			return model.InputNone
-		}
```
bassosimone, line 470: We don't need the input policy anymore. We just rely on mocking the target loader in `fakeSuccessfulDeps`.

pkg/oonimkall/taskrunner_test.go
```diff
@@ -546,62 +491,25 @@ func TestTaskRunnerRun(t *testing.T) {
...
+		fake.Builder.MockNewTargetLoader = func(config *model.ExperimentTargetLoaderConfig) model.ExperimentTargetLoader {
+			return &mocks.ExperimentTargetLoader{
+				MockLoad: func(ctx context.Context) (targets []model.ExperimentTarget, err error) {
+					// We need to mimic wht would happen when settings.Inputs is explicitly provided
+					for _, input := range inputs {
+						targets = append(targets, model.NewOOAPIURLInfoWithDefaultCategoryAndCountry(input))
+					}
+					return
+				},
+			}
```
bassosimone, line 512: Like before, we can remove test cases because we treat experiments uniformly since we use a target loader.

internal/targetloading/targetloading.go
```diff
@@ -215,10 +215,10 @@ var dnsCheckDefaultInput = []string{
 
 var stunReachabilityDefaultInput = stuninput.AsnStunReachabilityInput()
 
-// StaticBareInputForExperiment returns the list of strings an
+// staticBareInputForExperiment returns the list of strings an
 // experiment should use as static input. In case there is no
 // static input for this experiment, we return an error.
-func StaticBareInputForExperiment(name string) ([]string, error) {
+func staticBareInputForExperiment(name string) ([]string, error) {
```
bassosimone, line 221: This function can become private because `./pkg/oonimkall` does not need it anymore.

internal/targetloading/targetloading.go
```diff
@@ -364,15 +364,6 @@ func (il *Loader) checkIn(
...
 
-// fetchOpenVPNConfig fetches vpn information for the configured providers
-func (il *Loader) fetchOpenVPNConfig(ctx context.Context, provider string) (*model.OOAPIVPNProviderConfig, error) {
-	reply, err := il.Session.FetchOpenVPNConfig(ctx, provider, "XX")
-	if err != nil {
-		return nil, err
-	}
-	return reply, nil
-}
-
```
bassosimone, line 375: We can remove this function, since no-one else is using it.

pkg/oonimkall/taskrunner.go
```diff
@@ -291,62 +304,68 @@ func (r *runnerForTask) Run(rootCtx context.Context) {
...
 		}
 
-		// Richer input implementation note: in mobile, we only observe richer input
-		// for Web Connectivity and only store this kind of input into the database and
-		// otherwise we ignore richer input for other experiments, which are just
-		// treated as experimental. As such, the thinking here is that we do not care
-		// about *passing* richer input from desktop to mobile for some time. When
-		// we will care, it would most likely suffice to require the Inputs field to
-		// implement in Java the [model.ExperimentTarget] interface, which is something
-		// we can always do, since it only has string accessors.
```
bassosimone, line 305: This comment was not accurate.

pkg/oonimkall/taskrunner_test.go
```diff
@@ -270,6 +277,15 @@ func TestTaskRunnerRun(t *testing.T) {
...
 			},
+
+			Loader: &mocks.ExperimentTargetLoader{
+				MockLoad: func(ctx context.Context) ([]model.ExperimentTarget, error) {
+					targets := []model.ExperimentTarget{
+						model.NewOOAPIURLInfoWithDefaultCategoryAndCountry(""),
+					}
+					return targets, nil
+				},
+			},
```
bassosimone, line 288: By default we mock an experiment taking an empty entry (like dash, ndt, telegram, etc.)

DecFox (approved):
LGTM!

PR #1621, opened 2024-06-07
chore(richer input): add living design document

bassosimone:
This commit adds a living design document for richer input. The intent for the final document is to explain the problem we wanted to solve, the alternatives we considered, and how we specifically implemented it.

For now, the intent is to collect rough notes about what we have done so far and what to do next, such that @DecFox and I can synchronize on it and continue working on it, possibly also with input from @ainghazal.


DecFox (approved):
LGTM!

PR #1624, opened 2024-06-19
chore(oonimkall): log the events when testing

bassosimone:
This diff, which is part of https://github.com/ooni/probe/issues/2752, prints the event occurring when testing, which is useful for me to better understand what is going on while changing the package.

Extracted from https://github.com/ooni/probe-cli/pull/1620.


DecFox (approved):
LGTM!

PR #1628, opened 2024-06-25
refactor(probeservices): use better naming

bassosimone:
Calling the endpoint base URL (e.g., `https://www.example.com/`) "endpoint" could cause confusion because usually an API endpoint is something like `https://www.example.com/api/v1/check-in`. To obviate this semantic issue, this diff attempts to avoid calling base URLs as endpoint throughout the tree.

I am going to account this work as part of https://github.com/ooni/backend/issues/754 because this is one of the possibly issues on which we can account this work. The original conversation with @ainghazal, which prodded me to make these changes, was related to his feedback after implementing https://github.com/ooni/probe-cli/pull/1625.


internal/probeservices/probeservices.go
```diff
@@ -32,8 +32,8 @@ import (
 )
 
 var (
-	// ErrUnsupportedEndpoint indicates that we don't support this endpoint type.
-	ErrUnsupportedEndpoint = errors.New("probe services: unsupported endpoint type")
+	// ErrUnsupportedServiceType indicates that we don't support this service type.
+	ErrUnsupportedServiceType = errors.New("probe services: unsupported service type")
```
bassosimone, line 36: We generally do not change error strings. However, this error string is for a very uncommon case, because the service type should always be supported and so it's fine to modify it for consistency.

ainghazal (approved):
lgtm

PR #1630, opened 2024-06-26
fix: correctly set options from richer input

bassosimone:
While there, update the living design document with features and fixes implemented in this sprint.

Closes https://github.com/ooni/probe/issues/2763.


DecFox (approved):
LGTM!

PR #1634, opened 2024-07-03
feat: add engine docs to docs.ooni.org

DecFox:
## Checklist

## Description

This diff introduces workflow to show docs in `ooni/probe-cli` on docs.ooni.org


.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Checkout ooni/docs
+        uses: actions/checkout@v2
+        with:
+          repository: "ooni/docs"
+          ssh-key: ${{ secrets.OONI_DOCS_DEPLOYKEY }}
+          path: "ooni-docs"
+
+      - name: Update docs
+        run: cp -R dist/docs/* ooni-docs/src/content/docs/mydocs/
```
hellais, line 26: Wherever it says `mydocs` it should instead say `probe`

scripts/build_docs.sh
```diff
@@ -0,0 +1,21 @@
...
+mkdir -p $DOCS_ROOT
+
+strip_title() {
+    # Since the title is already present in the frontmatter, we need to remove
+    # it to avoid duplicate titles
+    local infile="$1"
+    cat $infile | awk 'BEGIN{p=1} /^#/{if(p){p=0; next}} {print}'
+}
+
+cat <<EOF>$DOCS_ROOT/00-index.md
```
hellais, line 15: You should in here actually make an inventory of all the docs that exist in this repo and copy them over to the correct `probe` directory so we can render them.

Be sure to setup a reasonable slug for each of them so that the permalink is correct.

hellais (changes requested):
It's a good start, but it's missing critical pieces to make it mergeable.

Basically I would do a `find . -type f -name "*.md"` in the root of `probe-cli` and then figure out where they should live on docs.ooni.org and come up with reasonable slugs and titles for each of them.

You can ignore readmes that are very redundant, things like this one: https://github.com/ooni/probe-cli/blob/master/internal/cmd/README.md, which have basically not relevant content in them. However things like this: https://github.com/ooni/probe-cli/blob/master/internal/tutorial/dslx/chapter01/README.md should be included in docs.ooni.org and placed in the right place (eg. /probe-engine/dslx/tutorials/chapter-1)

Moreover the directory of the docs should not be `mydocs`, but rather should be `probe` or `probe-engine`

script/build_docs.sh
DecFox, line 1: We are only pushing the following docs to docs.ooni.org for now: `docs/design`, `docs/releasing.md`, `pkg/oonimkall` and `README.md`.

DecFox:
Thanks for the review. Pushed the desired changes. As proposed, we have renamed the content directory for `ooni/probe-cli` docs to be `probe-engine` here:  https://github.com/ooni/probe-cli/pull/1634/files#diff-4dfe935b4852393622c2b414783d1d11e1a868fd8485709124d90208fa61ec8bR26

hellais (approved):
This looks good to me. We should also add probe-engine to the sidebar, but that can be done directly in the ooni/docs repo.

PR #1635, opened 2024-07-05
Start working on fixing image paths

hellais:
This diff moves assets used in docs to `src/assets` in `ooni/docs` and also replaces the doc references in markdown files to the appropriate location.


hellais:
Yeah this looks good. Thanks for putting it together. I suggest you merge it when CI passes if you want to work on the docs side of things.

DecFox:
The CI seems to be failing with a go race error which is unrelated to the diff here. I am going to go ahead and merge.

PR #1639, opened 2024-08-04
fix(build): fix build on FreeBSD

rene0:
## Checklist

- [x] reference issue for this pull request: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270020

## Description

The build of ooni-probe currently fails on FreeBSD because the compiler cannot find some required symbols (C.SOCK_STREAM, C.AF_INET, C.AF_INET6 and C.AF_UNSPEC). This causes the package [net/ooni-probe-cli](https://www.freshports.org/net/ooni-probe-cli/) to be unavailable.

Fix this by including sys/socket.h when building on FreeBSD.

See also these FreeBSD bug reports (the second report also applies to Go 1.21 and Go 1.22):

- [270020: net/ooni-probe-cli: update 3.21.1 → 3.22.0, fix build](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270020)
- [272643: lang/go120: Build of one port fails with could not determine kind of name for C.AF_INET, etc](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272643)

I took a bit of a shortcut for the fix in the first URL by omitting the #ifdef as that patch would only exist in the FreeBSD port anyway. Both URLs also discuss if this is a problem in OONI or Go itself.


bassosimone:
Thank you for your diff!

I've taken a look at [IEEE Std 1003.1](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/socket.h.html), and it says that the `sys/socket.h` file should define `SOCK_STREAM, `AF_INET`, etc. About [netdb.h](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/basedefs/netdb.h.html), it says that "inclusion of the <netdb.h> header may also make visible all symbols from [...] sys/socket.h". Because it says *may*, for correctness, I think this repository should always include `sys/socket.h` when it is not not compiling for Windows.

Regarding this:

> Both URLs also discuss if this is a problem in OONI or Go itself.

This issue is definitely an issue with this codebase and not an issue with Go itself.

Cc: @DecFox

DecFox:
Thank you for the PR! LGTM. I am going to go ahead and merge.

PR #1654, opened 2024-10-08
feat: add fallback domain names for openvpn experiment

ainghazal:
while working on this, I also gave more priority to possible oonirun descriptors passed in the command line.

- Related: #2805

## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2805

## Description

Add fallback domains for openvpn default endpoints to be probed.

### DIFF
--- a/internal/experiment/openvpn/targets.go
+++ b/internal/experiment/openvpn/targets.go
@@ -4,9 +4,87 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"slices"
 )
 
-const defaultOpenVPNEndpoint = "openvpn-server1.ooni.io"
+// defaultOpenVPNEndpoints contain a list of all default endpoints
+// to be tried, in the order that we want the name resolution to happen.
+var defaultOpenVPNEndpoints = []endpoint{
+	// default domain. this should work fine for most places.
+	{
+		IPAddr:      "",
+		DomainName:  "openvpn-server1.ooni.io",
+		Obfuscation: "none",
+		Port:        "1194",
+		Protocol:    "openvpn",
+		Provider:    "oonivpn",
+		Transport:   "tcp",
+	},
+	{
+		IPAddr:      "",
+		DomainName:  "openvpn-server1.ooni.io",
+		Obfuscation: "none",
+		Port:        "1194",
+		Protocol:    "openvpn",
+		Provider:    "oonivpn",
+		Transport:   "udp",
+	},
+	// alt domain 1. still same endpoint ports, one udp and one tcp.
+	// TODO(ain): update to real domain names
+	{
+		IPAddr:      "",
+		DomainName:  "alt-domain1.example.org",
+		Obfuscation: "none",
+		Port:        "1194",
+		Protocol:    "openvpn",
+		Provider:    "oonivpn",
+		Transport:   "udp",
+	},
+	{
+		IPAddr:      "",
+		DomainName:  "alt-domain1.example.org",
+		Obfuscation: "none",
+		Port:        "1194",
+		Protocol:    "openvpn",
+		Provider:    "oonivpn",
+		Transport:   "tcp",
+	},
+	// alt domain 2. still same endpoint ports, one udp and one tcp.
+	// TODO(ain): update to real domain names
+	{
+		IPAddr:      "",
+		DomainName:  "alt-domain2.example.org",
+		Obfuscation: "none",
+		Port:        "53",
+		Protocol:    "openvpn",
+		Provider:    "oonivpn",
+		Transport:   "udp",
+	},
+	{
+		IPAddr:      "",
+		DomainName:  "alt-domain2.example.org",
+		Obfuscation: "none",
+		Port:        "443",
+		Protocol:    "openvpn",
+		Provider:    "oonivpn",
+		Transport:   "tcp",
+	},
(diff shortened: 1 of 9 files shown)


ainghazal:
changed the logic after 1:1 with @hellais, ready for review

internal/experiment/openvpn/targets.go
```diff
@@ -1,49 +1,170 @@
...
+	"a.composer-presenter.com",
+	"a.goodyear2dumpster.com",
+}
+
+// maxDefaultOONIAddresses is how many IPs to use from the
+// set of resolved IPs.
+var maxDefaultOONIAddresses = 3
+
+// sampleN takes max n elements sampled ramdonly from the array a.
+func sampleN(a []string, n int) []string {
```
hellais, line 30: This function can probably be optimized by shuffling the list and return the first N elements from it instead of having to keep track of which indexes you picked as this might have a not so predictable runtime.

internal/experiment/openvpn/targets.go
```diff
@@ -1,49 +1,170 @@
...
-func defaultOONIOpenVPNTargetUDP() (string, error) {
-	ip, err := resolveTarget(defaultOpenVPNEndpoint)
-	if err != nil {
-		return "", err
+// resolveOONIAddresses returns a max of maxDefaultOONIAddresses after
+// performing DNS resolution. The returned IP addreses exclude possible
+// bogons.
+func resolveOONIAddresses(logger model.Logger) ([]string, error) {
+	resolver := netx.NewResolver(netx.Config{
+		BogonIsError: false,
```
hellais, line 56: It's probably worth adding a note here that we set this to `BogonIsError` to `false` and after remove them from the list so that we are able to control the rate at which we run tests by adding bogon addresses to the domain records for the test.

internal/experiment/openvpn/targets.go
```diff
@@ -1,49 +1,170 @@
...
+
+	targets := make([]string, 0)
+	for _, e := range endpoints {
+		targets = append(targets, e.AsInputURI())
+	}
+	if len(targets) > 0 {
+		return targets, nil
 	}
-	return defaultOONITargetURL(ip) + "&transport=tcp", nil
+	return nil, fmt.Errorf("cannot find any usable endpoint")
```
hellais, line 167: Will this result in a hard failure of the test? Is there some way we can ensure that's not the case? For example we might want for certain runs in particular countries to not run the vpn tests and it would be ideal if that results in a noop as opposed to the test being considered a failure.

internal/experiment/openvpn/targets.go
ainghazal: fixed. 

```
goos: linux
goarch: amd64
pkg: github.com/ainghazal/test
cpu: AMD Ryzen 7 5800X 8-Core Processor
                     │ results.txt │
                     │   sec/op    │
SampleFunc1_Small-16   5.322µ ± 1% <- baseline
SampleFunc2_Small-16   5.286µ ± 1% <- new version
SampleFunc3_Small-16   5.482µ ± 1%
geomean                5.362µ
```

internal/experiment/openvpn/targets.go
ainghazal: added!

internal/experiment/openvpn/targets.go
ainghazal: if this function returns error, we simply return an empty targets list, see: 
https://github.com/ooni/probe-cli/blob/d9c4b3e4e8ea764829e7e0f5cf5f2a302bdf8b61/internal/experiment/openvpn/richerinput.go#L113

ainghazal:
To clarify, I've added a few more comments to the docstring of the Load() function. An error is ignored (logged now); returning the empty list (in case of all bogons) is equivalent of passing a null input to the experiment.

@hellais: here I've pointed both domains to localhost in /etc/hosts, see:

```
[      0.328261] <info> Looking up your location; please be patient...
[      0.328277] <info> iplookup: using cloudflare
[      0.328407] <info> sessionresolver: lookup www.cloudflare.com using https://wikimedia-dns.org/dns-query... started
[      0.376407] <info> sessionresolver: lookup www.cloudflare.com using https://wikimedia-dns.org/dns-query... ok
[      0.485202] <info> - country: NO
[      0.485209] <info> - network: FOOBAR (AS0)
[      0.485213] <info> - resolver's IP: 172.253.5.153
[      0.485216] <info> - resolver's network: Google LLC (AS15169)
[      0.485338] <warn> Picking from default OpenVPN endpoints
[      0.485343] <warn> No targets loaded from default endpoints
[      0.485351] <info> experiment: recv   0.00  byte, sent   0.00  byte
[      0.485519] <info> whole session: recv   3.86 kbyte, sent   1.46 kbyte
```

internal/experiment/openvpn/targets.go
ainghazal: see the [comment below](https://github.com/ooni/probe-cli/pull/1654#issuecomment-2436080116)

DecFox (approved):
LGTM! I pushed a change to the `go.sum` file here: https://github.com/ooni/probe-cli/pull/1654/commits/ffbd670baa28b0e1ab8f49d6b9f39877b375c2e8, to update the `minivpn` dependency to `v0.0.7`.

PR #1656, opened 2024-10-15
Simplify the dnscheck list

hellais:
Trim down the dnscheck list.

Keep only 2 addresses per providers (http3 enabled and not), since measuring the ads or malware blocking endpoints is less important (we only query example.com). The main difference for the malware blocking endpoints would be that the SNI is different, but we can probably generate similar telemetry without doing a full DoH measurement.


DecFox:
Thanks for the changes! I made some changes here: https://github.com/ooni/probe-cli/pull/1656/commits/ca8384933d7a316b63405b3baa3fa42cf2272758, to shorten the default target list while renaming the former list to `extendedInput`. We can possibly in the future have a configuration flag to decide which target list we should return.

PR #1667, opened 2024-11-25
feat(dnscheck): add resolver urls for trust anchor change testing

DecFox:
## Checklist

## Description

This diff adds support for measuring using the system resolver and also adds resolver urls allowing us to test the trust anchor change.


hellais:
I tested this branch both with how it's setup now and by replacing the address definitions with `system://`, however I don't think it's working as intended.

The fact that `dnscheck` has this two step bootstrap and lookup process, means that it's interpreting the input definitions as the configuration vectors for a DoH/DoT server.

I think that in order to get this work well we should spend a bit more time refining the test, so I suggest we defer merging this to a future probe-engine release and in the interim we just add the URLs to web_connectivity.

hellais (changes requested):
We should defer this to a future release.

PR #1673, opened 2024-12-07
Add support for returning measurement_uid as part of oonimkall API

hellais:
Implements: https://github.com/ooni/probe/issues/2521


DecFox (approved):
LGTM! I have tested this with the `probe-android` repo and we can successfully expose the `measurement_uid` in the `EventResult` class through `oonimkall`.

PR #1675, opened 2024-12-19
Add ddr experiment

frcroth:
## Checklist

- [x] reference issue for this pull request: https://github.com/ooni/probe/issues/2819
- [x] if you changed anything related to how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: https://github.com/ooni/spec/pull/300

## Description

As described here: https://github.com/ooni/probe/issues/2819, I would be interested to run experiments of the DDR protocol (RFC 9462). This experiment uses the system resolver, as I am interested in whether the local system resolvers support this protocol (we have already conducted tests using zmap from one vantage point). 

To get the system resolver I am checking /etc/resolve, it should also be possible by querying using the default DNS provider and checking the answer or config, however I did not get this to work.

~~I have no idea about what the spec should look like. Right now I just put the DDR responses as JSON objects into the testkeys. If that is not supported because it could violate the spec, it would also be possible to just insert the string / do some parsing here to get some more useful information (e.g. if DoT or DoH is supported, what the target is...)~~

## TODOs

- [x] Testing
- [x] PR for spec

### DIFF
--- a/internal/bytecounter/resolver.go
+++ b/internal/bytecounter/resolver.go
@@ -41,6 +41,11 @@ func (r *ContextAwareSystemResolver) LookupHTTPS(ctx context.Context, domain str
 	return r.wrap(ctx).LookupHTTPS(ctx, domain)
 }
 
+// LookupSVCB implements model.Resolver.
+func (r *ContextAwareSystemResolver) LookupSVCB(ctx context.Context, domain string) ([]*model.SVCB, error) {
+	return r.wrap(ctx).LookupSVCB(ctx, domain)
+}
+
 // LookupHost implements model.Resolver.
 func (r *ContextAwareSystemResolver) LookupHost(ctx context.Context, hostname string) (addrs []string, err error) {
 	return r.wrap(ctx).LookupHost(ctx, hostname)
@@ -104,6 +109,14 @@ func (r *resolver) CloseIdleConnections() {
 func (r *resolver) LookupHTTPS(ctx context.Context, domain string) (*model.HTTPSSvc, error) {
 	r.updateCounterBytesSent(domain, 1)
 	out, err := r.Resolver.LookupHTTPS(ctx, domain)
+	r.updateCounterBytesRecv(err)
+	return out, err
+}
+
+// LookupSVCB implements model.Resolver
+func (r *resolver) LookupSVCB(ctx context.Context, domain string) ([]*model.SVCB, error) {
+	r.updateCounterBytesSent(domain, 1)
+	out, err := r.Resolver.LookupSVCB(ctx, domain)
 	r.updateCounterBytesRecv(err)
 	return out, err
 }
--- a/internal/bytecounter/resolver_test.go
+++ b/internal/bytecounter/resolver_test.go
@@ -122,6 +122,84 @@ func TestMaybeWrapSystemResolver(t *testing.T) {
 		})
 	})
 
+	t.Run("LookupSVCB works as intended", func(t *testing.T) {
+		t.Run("on success", func(t *testing.T) {
+			expected := []*model.SVCB{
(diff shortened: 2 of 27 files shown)


hellais:
Hi, thanks for proposing this experiment and for working on an implementation.

In general I think it would be great to have a specification written for it in order to assist reviewing the implementation better. Some particularly sensitive areas which I would like to ensure are covered in the specification are if there are any privacy implications for users running this test. 

For example, is there a risk that when performing the lookup for the designated resolvers, the list will include information that uniquely identifies a user? Has this been observed or is it possible as per the spec?

Regarding the specific implementation I see in this PR, the biggest issue is that it doesn't extend the DNS test specification (see: https://github.com/ooni/spec/blob/master/data-formats/df-002-dnst.md#dns-data-format). Ideally we would extend that to support encoding `SVCB` records and we might even opt for offloading the job of parsing them to the data pipeline instead of doing it inline in the probe.

Finally, I wonder if instead of creating a new specialized test for DDR, we could instead work on adding support for this to our `dnscheck` test or if not have a more general purpose DNS lookup test that allows us to run any query we might need.

This would also help us address the issue we had with needing to ship DNS queries as web_connectivity tests where it would have instead been cleaner to do it as part of a DNS specific test (see: https://github.com/citizenlab/test-lists/pull/1883).

In summary the next steps to move this PR forward are:
* [ ] Create PR updating the dns base test specification to support `SVCB` records
* [ ] Document somewhere the eventual privacy considerations associated with performing DDR queries
* [ ] Evaluate if it's possible to extend DNSCheck to do SVCB queries or if we should instead create a new simplified general purpose DNS experiement (@DecFox can you help with this?)
* [ ] Write spec for such test or update dnscheck spec to mention svcb record support
* [ ] Implement code for it

frcroth:
> Hi, thanks for proposing this experiment and for working on an implementation.
> 
> In general I think it would be great to have a specification written for it in order to assist reviewing the implementation better. Some particularly sensitive areas which I would like to ensure are covered in the specification are if there are any privacy implications for users running this test.
> 
> For example, is there a risk that when performing the lookup for the designated resolvers, the list will include information that uniquely identifies a user? Has this been observed or is it possible as per the spec?

The result would include information about the user's recursive resolver, including IP address and the DDR configuration, which could be used to uniquely identify this recursive resolver.
Since DDR is pretty new, there is basically no literature, so there has been nothing on identifying users afaik.

> Regarding the specific implementation I see in this PR, the biggest issue is that it doesn't extend the DNS test specification (see: [ooni/spec@`master`/data-formats/df-002-dnst.md#dns-data-format](https://github.com/ooni/spec/blob/master/data-formats/df-002-dnst.md?rgh-link-date=2025-01-14T10%3A23%3A54Z#dns-data-format)). Ideally we would extend that to support encoding `SVCB` records and we might even opt for offloading the job of parsing them to the data pipeline instead of doing it inline in the probe.
> 
> Finally, I wonder if instead of creating a new specialized test for DDR, we could instead work on adding support for this to our `dnscheck` test or if not have a more general purpose DNS lookup test that allows us to run any query we might need.

Thank you for that suggestion, I will look into it.

> This would also help us address the issue we had with needing to ship DNS queries as web_connectivity tests where it would have instead been cleaner to do it as part of a DNS specific test (see: [citizenlab/test-lists#1883](https://github.com/citizenlab/test-lists/pull/1883)).

frcroth:
@hellais 

I have now implemented "LookupSVCB" in the resolvers, which brought a lot of changes in 47a97b.
The ddr experiment is simplified by that, since querying and parsing is done in libraries. I could now go ahead and also try to merge this with dnscheck if that makes more sense. What do you think?

PR #1777, opened 2026-02-11
fix: update desktop JAR publishing process

aanorbel:


DecFox (approved):
LGTM! We should remove the experimental branch from the gh workflow and this should be GTG🐳

Issue #1728, opened 2025-08-29
[Bug] Investigate hard failures while running echcheck on specific targets


DecFox:
This is a follow up on the issue in the comment here: https://github.com/ooni/probe-cli/issues/1725#issuecomment-3235910775, reported by ooniprobe users. It seems echcheck fails with a panic when the target is `min-ng.test.defo.ie`. We should investigate why this happens and also ensure that the test in general does not panic on targets.

DecFox:
This is now fixed. I am going to go ahead and close this issue 🐳

Issue #1725, opened 2025-08-28
[bug] Inconsistent measurements for echcheck targets


DecFox:
An inconsistency in the number of measurements for the `echcheck` experiment was reported, where the ditribution between measurements with the target `min-ng.test.defo.ie` seem to be less than the ones for `https://cloudflare-ech.com/cdn-cgi/trace`.

DecFox:
On further analysis, @hellais rightly pointed out that `min-ng.test.defo.ie` is hardcoded as a target in the ooniprobe cli: https://github.com/ooni/probe-cli/blob/b68cb8bd8b34fd548f325b014f1daff840fca739/cmd/ooniprobe/internal/nettests/echcheck.go, which leads to the target being measured when using the CLI but not the mobile apps. The general policy is to avoid setting up targets directly in the CLI, hence the fix is to remove them and default to only testing the cloudflare domain set up in the experiment itself

PR #1817, opened 2026-09-09
fix(echcheck): do not attempt a TLS handshake when the TCP connect failed

macintoshi-nakamoto:
## Checklist

- [x] reference issue for this pull request: #1728
- [x] if you changed anything related to how experiments work and you need to reflect these changes in the ooni/spec repository, please link to the related ooni/spec pull request: not needed, the test keys do not change shape
- [x] if you changed code inside an experiment, make sure you bump its version number: 0.3.1 -> 0.3.2

## Description

`handshake()` in `internal/experiment/echcheck` dials the target and then calls `tls.Client(conn, ...).HandshakeContext()` without checking whether the dial succeeded. When it did not, `conn` is nil and `crypto/tls` panics while writing the ClientHello. This is the trace reported in #1725 and tracked in #1728: it shows up whenever a target stops answering on its port, as `min-ng.test.defo.ie:15443` did.

The change returns the test keys right after the failed TCP connect has been recorded, since that is the measurement: a `tcp_connects` entry with the failure and no `tls_handshakes` entry, because no handshake was attempted. No shape of `TestKeys` changes.

The regression test binds a port on localhost, closes it, and runs `startHandshake` against it with and without GREASE. Without the fix the test panics with the same nil pointer dereference as the issue; with the fix it passes. Verified with `go test -short ./internal/experiment/echcheck/`.

Small side observation, not addressed here to keep the diff minimal: the TLS conn is never closed after the handshake, so a successful handshake leaks the socket until the process exits. Happy to send that as a separate PR if you want it.


DecFox (approved):
LGTM! Thanks for working on this.

